PR #15, opened 2025-04-07
Exploring bdk-tx for Wallet type

thunderbiscuit:
This is just me hacking away at leveraging the bdk-tx crate to build transactions.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -6,6 +6,7 @@ members = [
     "examples/example_wallet_esplora_blocking",
     "examples/example_wallet_esplora_async",
     "examples/example_wallet_rpc",
+    "examples/example_wallet_bdk_tx",
 ]
 
 [workspace.package]
--- a/examples/example_wallet_bdk_tx/Cargo.toml
+++ b/examples/example_wallet_bdk_tx/Cargo.toml
@@ -1,0 +1,11 @@
+[package]
+name = "example_wallet_bdk_tx"
+version = "0.2.0"
+edition = "2021"
+publish = false
+
+[dependencies]
+bdk_wallet = { path = "../../wallet", features = ["file_store"] }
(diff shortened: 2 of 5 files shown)


thunderbiscuit:
The API for building transactions using bdk-tx doesn't have to be any different than what we currently have with the TxBuilder, but it could also be. Here are some rough sketches.

## Builder pattern

```rust
let mut tx_builder = wallet.create_transaction_builder();
tx_builder
    .add_output(address.script_pubkey(), SEND_AMOUNT)
    .fee_rate(FeeRate::from_sat_per_vb(4).unwrap());
let mut psbt = tx_builder.finish()?;
```

## Single method

```rust
// pseudocode

let mut psbt = wallet.create_transaction(
    outputs,
    target_feerate,
    must_spend,
    // ...
    // 12 other arguments that basically mimic the current TxBuilder methods
);
```

Note that this method could also take a single argument `TransactionParams` which would just be a data structure of the requirements for the tx, again similar to what we have in the private fields of the TxBuilder.

```rust
let transaction_params = TransactionParams {
    outputs: vec![(address.script_pubkey(), SEND_AMOUNT)],
    target_feerate: FeeRate::from_sat_per_vb(4).unwrap(),
    must_spend: Vec::new(),
};
let mut psbt = wallet.create_transaction(transaction_params);
```

## 2 methods (one for simple common cases)

The first one for simple cases often used and one that takes the full TransactionParams argument.

```rust
let mut psbt = wallet.create_simple_transaction(
    outputs, // (Script, Amount)
    target_feerate, // FeeRate
);

let mut psbt = wallet.create_complex_transaction(transaction_params);
```

thunderbiscuit:
The API for this is being built for production in #297. Note that the method chosen in that PR is the "single method" mentioned above with a `Params` struct passed to the method rather than the builder.

Issue #31, opened 2025-01-30
Add spent trait to improve sent_and_received()


matthiasdebernardini:
**Describe the enhancement**
Add a `TxAmountSpent` trait to improve how transaction spending amounts are calculated. The trait will add a `.spent()` method to tuples returned by `sent_and_received()`, calculating total amount spent including fees.

**Use case**
```rust
// Current: Manual calculation needed
let (sent, received) = wallet.sent_and_received(&tx);
let fee = wallet.calculate_fee(&tx)?;
let total = sent - received + fee; // Error-prone 

// Proposed: Clean trait implementation
let (sent, received) = wallet.sent_and_received(&tx);
let total = (sent, received, fee).spent(); // Clear and consistent
```

**Additional context**
- Non-breaking change that extends existing functionality
- Makes transaction cost calculations more intuitive
- Helps prevent calculation errors
- Small, focused improvement suitable for a first-time contributor

YuriQueirozAndrade:
Now I understand what you're saying about the fee. You're right, I forgot about it; it can be deducted fee from the received value.
About the sign of the result, I'm not sure, but if it's a net outflow, the sign should be negative. Maybe it's better to change the function name to `total_outflow` ( or another thing like this ) for better semantic clarity with the result.
I say this why, if i want a `total_outflow`, i want the negative value, right ? #

> * I think the `fee` term is superfluous, see my comment [feat(wallet): Introduce `TxAmountSpend` trait for improved spend calculation bdk#1842 (comment)](https://github.com/bitcoindevkit/bdk/pull/1842#issuecomment-3079763911).
> 
>     * How are we to interpret the sign of the result? If it's a net outflow should the sign be positive or negative?

PR #201, opened 2025-04-08
Bring back TxDetails

### LINKED ISSUE #100: The wallet should have a utility method to return standard transaction data
Since the purpose of the `bdk_wallet` crate is to provide useful/opinionated ways to perform standard wallet tasks, it feels to me like a simple-to-use "get me all important—nicely bundled if you don't mind—info about this wallet's transactions" method is missing from our API.

The data is there, but currently requires multiple round trips to bundle it all together, yet that data is what all user interfaces will need for one of the most common use cases, a transactions history screen. Note that this type of data object is what we previously had with the [TransactionDetails](https://docs.rs/bdk/0.29.0/bdk/struct.TransactionDetails.html#structfield.txid) struct.

The exact data that might be useful here is up for discussion, but let us assume a structure like this one would suffice for most use cases:
```rust
pub struct TxDetails<'a> {
    pub received: u64,
    pub sent: u64,
    pub fee: u64,
    pub fee_rate: FeeRate,
    pub txid: Txid,
    pub chain_position: ChainPosition<&'a ConfirmationTimeHeightAnchor>,
}
```

The current way to bundle this up ready for UI consumption is something like this:
```rust
let transactions = wallet.transactions();

let tx_details: Vec<TxDetails> = transactions.map(|tx| {
        let (sent, received) = wallet.sent_and_received(tx.tx_node.tx);
        let fee = wallet.calculate_fee(tx.tx_node.tx).unwrap();
        let fee_rate = wallet.calculate_fee_rate(tx.tx_node.tx).unwrap();
        let txid = tx.tx_node.txid;
        let chain_position = tx.chain_position.clone();

        TxDetails {
            received,
            sent,
            fee,
            fee_rate,
            txid,
            chain_position,
        }
    })
    .collect();

for tx in &tx_details {
    let ui_details = format!(
        "Received: {}\nSent: {}\nTotal fee: {} sat\nFee rate: {} sat/vbyte\nTxid: {}\nChain position: {:?}\n",
        tx.received,
        tx.sent,
        tx.fee,
        tx.fee_rate.to_sat_per_vb_ceil(),
        tx.txid,
        tx.chain_position
    );
    println!("{}", ui_details);
}
```

Notice the individual calls to the wallet for for fee, fee_rate, and sent_and_received values, as well as the pulling out of the txid. It gets even more unruly if you want to build fields like `confirmed: Boolean`, `confirmation_block: Option<u64>`, `confirmation_timestamp: Option<u64>` instead of using the ChainPosition type, etc.

I propose we add a method (name tbd, but maybe something like `get_txs_details()`, which would take care of building such a data structure for the user, instead of making all users develop their utility method on top of `wallet.transactions()`.

thunderbiscuit:
Fixes #100.

The `TxDetails` feature was put aside to focus on the 1.0 release, but now that we're done with that I want to revive the idea.

This is very much WIP. I'm trying to figure out exactly what we want in this data structure, and I'm not even sure my way to get at the data is the most efficient way to do it.

### Changelog notice

```md
Added:
    - Wallet::get_tx_details method [#201]

[#201]: https://github.com/bitcoindevkit/bdk_wallet/pull/201
```

#### All Submissions:

#### New Features:

#### Bugfixes:

### DIFF
--- a/wallet/src/wallet/mod.rs
+++ b/wallet/src/wallet/mod.rs
@@ -41,7 +41,7 @@ use bitcoin::{
     secp256k1::Secp256k1,
     sighash::{EcdsaSighashType, TapSighashType},
     transaction, Address, Amount, Block, BlockHash, FeeRate, Network, OutPoint, Psbt, ScriptBuf,
-    Sequence, Transaction, TxOut, Txid, Weight, Witness,
+    Sequence, SignedAmount, Transaction, TxOut, Txid, Weight, Witness,
 };
 use miniscript::{
     descriptor::KeyMap,
@@ -81,6 +81,7 @@ pub use changeset::ChangeSet;
 pub use params::*;
 pub use persisted::*;
 pub use utils::IsDust;
+pub use utils::TxDetails;
 
 /// A Bitcoin wallet
 ///
@@ -859,6 +860,33 @@ impl Wallet {
                 self.indexed_graph.index.outpoints().iter().cloned(),
             )
             .map(|((k, i), full_txo)| new_local_utxo(k, i, full_txo))
+    }
+
+    /// Get the [`TxDetails`] of a wallet transaction.
+    ///
+    /// If the transaction with txid [`Txid`] cannot be found in the wallet's transactions, `None`
+    /// is returned.
+    pub fn tx_details(&self, txid: Txid) -> Option<TxDetails> {
+        let tx: WalletTx = self.transactions().find(|c| c.tx_node.txid == txid)?;
+
+        let (sent, received) = self.sent_and_received(&tx.tx_node.tx);
+        let fee: Option<Amount> = self.calculate_fee(&tx.tx_node.tx).ok();
+        let fee_rate: Option<FeeRate> = self.calculate_fee_rate(&tx.tx_node.tx).ok();
+        let balance_delta: SignedAmount = self.indexed_graph.index.net_value(&tx.tx_node.tx, ..);
+        let chain_position = tx.chain_position;
+
+        let tx_details: TxDetails = TxDetails {
+            txid,
+            received,
+            sent,
+            fee,
+            fee_rate,
+            balance_delta,
+            chain_position,
+            tx: tx.tx_node.tx,
+        };
+
+        Some(tx_details)
     }
 
     /// List all relevant outputs (includes both spent and unspent, confirmed and unconfirmed).
--- a/wallet/src/wallet/utils.rs
+++ b/wallet/src/wallet/utils.rs
@@ -9,9 +9,12 @@
 // You may not use this file except in accordance with one or both of these
 // licenses.
 
+use alloc::sync::Arc;
 use bitcoin::secp256k1::{All, Secp256k1};
-use bitcoin::{absolute, relative, Amount, Script, Sequence};
-
+use bitcoin::{
+    absolute, relative, Amount, FeeRate, Script, Sequence, SignedAmount, Transaction, Txid,
+};
+use chain::{ChainPosition, ConfirmationBlockTime};
 use miniscript::{MiniscriptKey, Satisfier, ToPublicKey};
 
 use rand_core::RngCore;
@@ -132,6 +135,33 @@ pub(crate) fn shuffle_slice<T>(list: &mut [T], rng: &mut impl RngCore) {
 }
 
 pub(crate) type SecpCtx = Secp256k1<All>;
+
+/// Details about a transaction affecting the wallet (relevant and canonical).
+#[derive(Debug)]
+pub struct TxDetails {
+    /// The transaction id.
+    pub txid: Txid,
+    /// The sum of the transaction input amounts that spend from previous outputs tracked by this
+    /// wallet.
+    pub sent: Amount,
+    /// The sum of the transaction outputs that send to script pubkeys tracked by this wallet.
+    pub received: Amount,
+    /// The fee paid for the transaction. Note that to calculate the fee for a transaction with
+    /// inputs not owned by this wallet you must manually insert the TxOut(s) into the tx graph
+    /// using the insert_txout function. If those are not available, the field will be `None`.
+    pub fee: Option<Amount>,
+    /// The fee rate paid for the transaction. Note that to calculate the fee rate for a
+    /// transaction with inputs not owned by this wallet you must manually insert the TxOut(s) into
+    /// the tx graph using the insert_txout function. If those are not available, the field will be
+    /// `None`.
+    pub fee_rate: Option<FeeRate>,
+    /// The net effect of the transaction on the balance of the wallet.
+    pub balance_delta: SignedAmount,
+    /// The position of the transaction in the chain.
+    pub chain_position: ChainPosition<ConfirmationBlockTime>,
+    /// The complete [`Transaction`].
+    pub tx: Arc<Transaction>,
+}
 
 #[cfg(test)]
 mod test {
--- a/wallet/tests/wallet.rs
+++ b/wallet/tests/wallet.rs
@@ -19,7 +19,6 @@ use bdk_wallet::{KeychainKind, LoadError, LoadMismatch, LoadWithPersistError};
 use bitcoin::constants::{ChainHash, COINBASE_MATURITY};
 use bitcoin::hashes::Hash;
 use bitcoin::key::Secp256k1;
-use bitcoin::psbt;
 use bitcoin::script::PushBytesBuf;
 use bitcoin::sighash::{EcdsaSighashType, TapSighashType};
 use bitcoin::taproot::TapNodeHash;
@@ -27,6 +26,7 @@ use bitcoin::{
     absolute, transaction, Address, Amount, BlockHash, FeeRate, Network, OutPoint, ScriptBuf,
     Sequence, Transaction, TxIn, TxOut, Txid, Weight,
 };
+use bitcoin::{psbt, SignedAmount};
 use miniscript::{descriptor::KeyMap, Descriptor, DescriptorPublicKey};
 use rand::rngs::StdRng;
 use rand::SeedableRng;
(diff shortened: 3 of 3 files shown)


thunderbiscuit:
Who TH is this coveralls bot?

@coveralls explain yourself and tell me why I shouldn't just cut you out of my codebase my man

wallet/src/wallet/mod.rs
```diff
@@ -821,6 +821,45 @@ impl Wallet {
...
+        // method on the collection instead of collecting into a vector.
+
+        // This relies on the impossibility of having 2 transactions with the same Txid in the list
+        // canonical transactions but let me know if this is wrong!
+        let optional_tx: Option<CanonicalTx<Arc<Transaction>, ConfirmationBlockTime>> = tx_graph
+            .list_canonical_txs(&self.chain, self.chain.tip().block_id())
+            .filter(|c_tx| tx_index.is_tx_relevant(&c_tx.tx_node.tx))
+            .find(|tx| tx.tx_node.compute_txid() == txid);
+
+        let tx = optional_tx?;
```
ValuedMammal, line 843: Note on L841 you can also get the txid from the `c_tx.tx_node` (instead of computing it) :)

Could also condense this to something like the below, but also doesn't hurt to be explicit.

```rust
    let tx = self.transactions().find(|c| c.tx_node.txid == txid)?;
```

wallet/tests/wallet.rs
```diff
@@ -4306,3 +4306,17 @@ fn test_wallet_transactions_relevant() {
...
+    let (test_wallet, txid_1) = get_funded_wallet_wpkh();
+    let tx_details_1 = test_wallet.get_tx_details(txid_1);
+    // dbg!(&tx_details);
+
+    // Transaction id not part of the TxGraph
+    let txid_2 = Txid::from_raw_hash(Hash::all_zeros());
+    let tx_details_2 = test_wallet.get_tx_details(txid_2);
+
+    assert!(tx_details_1.is_some());
+    assert!(tx_details_2.is_none());
```
ValuedMammal, line 4321: 📌 Can also assert the expected value of the TxDetails struct.

ValuedMammal (commented):
Concept ACK. I did a first pass looks great so far.

wallet/src/wallet/mod.rs
```diff
@@ -821,6 +821,45 @@ impl Wallet {
...
+
+        let (sent, received) = self.sent_and_received(&tx.tx_node.tx);
+        let fee = self.calculate_fee(&tx.tx_node.tx).unwrap();
```
notmandatory, line 846: These unwraps will panic when you are calculating the fee or fee rate when you receive a tx that is paying to your wallet. The reason being that the inputs come from the senders wallet so you don't have the full input TX info in your wallet. I recently reminded of this when I was building a little bdk_wallet based cli app.  

I think the right solution is to make these values options, and they will only be populated if your wallet has the required input info. Logically this makes sense too, if you receive a tx from someone you don't pay any of the fees so there's nothing to show in your UI about fees or fee rate paid. If on the other hand you're paying someone else, you'll obviously have the TX with the UTXOs you're spending from and will want to show the fee amount you paid.

One edge case is what to do for something like a payjoin or coinjoin. In this case you have ~some~ of the input(s) and pay ~some~ of the fee. I don't have any bright ideas right now how to handle this situation.

wallet/src/wallet/mod.rs
thunderbiscuit: Done! Thanks for the tip.

wallet/tests/wallet.rs
thunderbiscuit: How should I go about this? I assume I'd need to create a TxDetails struct, but then that means adding a constructor on it no?

On the other hand I could assert some of the simpler values on it maybe?

thunderbiscuit: Ok I added some of those in!

thunderbiscuit:
Tagging some folks who had expressed interest in the feature: @matthiasdebernardini and @tnull.

I have a nagging feeling that there is probably more we want to have in this struct, and if we're going to build one then we might as well put the good stuff in it. Just wondering if you (or anyone else) has anything else they'd like to see as part of a struct that represent transaction details.

matthiasdebernardini:
Hey thanks! I'd also want a net impact in terms of balance on a per transaction basis. We had to build this ourselves but I think it would be very useful.

thunderbiscuit:
Ah yes I remember! Do you remember the issue number for the brainstorm/PR on this? I'll need to refresh my memory on what exactly we landed on. But that's a good idea.

matthiasdebernardini:
I think it was on discord, this is what we have implemented

```rust
pub struct WalletHistoryRecord {
    pub height: Option<u32>,
    pub node_info: TxNodeInfo,
    pub received: u64,
    pub sent: u64,
    pub delta: i128,
    pub balance: i128,
    pub txid: Txid,
}

pub struct TxNodeInfo {
    pub block_timestamp: Option<u32>,
    pub mempool_time: Option<u64>,
}
```

We use this to check if a wallet is spending too much money in a given time period (app specific) but I think its useful for building wallet summaries anyways.

If you want to implement this, you will need to also figure out a way to canonically sort these transactions if theres multiple transactions in the same block, if its random then youll get a different history each time you call it.

So far, I like what you have, but I would add wallet balance and the net change to the wallet.

thunderbiscuit:
Just linking here so I don't have so search for it later on: #31.

thunderbiscuit:
Thanks for the feedback. @matthiasdebernardini I decided to go with `balance_delta`, as I thought it more explicit (otherwise it's a delta of what? not as clear).

@ValuedMammal I know you'd talked about using the name `net_value`, but again I wasn't sure if that conveyed the meaning clearly. The value of something (what?), and then "net" of what? In this case I mean I know exactly what of course, but figured `balance_delta` was clearer on first read. Happy to bikeshed on this if others have different opinions!

@matthiasdebernardini I left out for now the idea of balance as part of this struct, because I feel like that's a bit different. It's not about the transaction per se, but really about the balance of the wallet at the time (and here the struct assumes the given transaction is the latest tx I guess?). I'm not again adding it, but I'm not yet convinced of the true value.

Again thanks for the feedback, this is ready for some more bikeshedding.

thunderbiscuit:
You know I had not thought about this before but it does sound interesting. "Balance at block X" type of method...

I'll open an issue and we can look at how best to add this.

wallet/src/wallet/mod.rs
```diff
@@ -822,6 +822,33 @@ impl Wallet {
             .map(|((k, i), full_txo)| new_local_utxo(k, i, full_txo))
     }
 
+    /// Build a [`TxDetails`] struct for a given transaction.
```
ValuedMammal, line 825:
```rs
    /// Get the details of a wallet transaction.
```

Or

```rs
    /// Get the [`TxDetails`] of a wallet transaction.
```

wallet/src/wallet/mod.rs
```diff
@@ -822,6 +822,33 @@ impl Wallet {
...
+    ///
+    /// If the transaction with txid [`Txid`] cannot be found in the wallet's transaction graph,
+    /// `None` is returned.
+    pub fn get_tx_details(&self, txid: Txid) -> Option<TxDetails> {
```
ValuedMammal, line 829: Unless you're adamant on sticking with `get_tx_details`, maybe consider just calling it `Wallet::tx_details` in keeping with the Rust API guidelines wrt naming.

wallet/src/wallet/utils.rs
```diff
@@ -133,6 +136,19 @@ pub(crate) fn shuffle_slice<T>(list: &mut [T], rng: &mut impl RngCore) {
 
 pub(crate) type SecpCtx = Secp256k1<All>;
 
+/// Details about a transaction the wallet knows about.
+#[derive(Debug)]
+pub struct TxDetails {
+    pub txid: Txid,
```
ValuedMammal, line 142: Ideally the public fields could be documented.

ValuedMammal (approved):
ACK

ValuedMammal:
> @ValuedMammal I know you'd talked about using the name `net_value`, but again I wasn't sure if that conveyed the meaning clearly. The value of something (what?), and then "net" of what? In this case I mean I know exactly what of course, but figured `balance_delta` was clearer on first read. Happy to bikeshed on this if others have different opinions!

`net_change` would sound ok too.

notmandatory:
Overall LGTM and +1 on Mammal's comments.

thunderbiscuit:
All comments addressed. 

@ValuedMammal I considered `net_change` earlier on in my thinking about this PR, but decided using the word `change` here potentially clashes with our other uses of the word change, (change keychains, sats coming back as change, etc.). 

In that sense I thought the field unclear; "net" change? Is that different than other change? etc. Here the word change would be used as meaning "the change in the wallet's balance", but it's not obvious at first read of a struct that handles sent/received values and whatnot. You might think it's just a trio of "what was sent", "what was received", "what was change" (even thought that's weird and doesn't make a lot of sense). In any case I decided against using change here and took "delta" from @matthiasdebernardini (thanks for the idea!).

ValuedMammal (approved):
tACK 3cd8921c6285e7fe6a1479fc454cf7cdc359d144

notmandatory (approved):
utACK 3cd8921c6285e7fe6a1479fc454cf7cdc359d144

wallet/tests/wallet.rs
```diff
@@ -4315,3 +4315,27 @@ fn test_wallet_transactions_relevant() {
     assert!(full_tx_count_before < full_tx_count_after);
     assert!(canonical_tx_count_before < canonical_tx_count_after);
 }
+
+#[test]
+fn test_tx_details_method() {
+    let (test_wallet, txid_1) = get_funded_wallet_wpkh();
+    let tx_details_1_option = test_wallet.tx_details(txid_1);
+    dbg!(&tx_details_1_option);
```
oleonardolima, line 4323: nit: keeping this `dbg!()` was intentional ?

oleonardolima (commented):
utACK 3cd8921c6285e7fe6a1479fc454cf7cdc359d144

oleonardolima (approved):
utACK 5f47713954a5657ccf76abef29c8064546c2afe3

wallet/src/wallet/mod.rs
```diff
@@ -72,7 +72,7 @@ use crate::wallet::{
     error::{BuildFeeBumpError, CreateTxError, MiniscriptPsbtError},
     signer::{SignOptions, SignerError, SignerOrdering, SignersContainer, TransactionSigner},
     tx_builder::{FeePolicy, TxBuilder, TxParams},
-    utils::{check_nsequence_rbf, After, Older, SecpCtx},
+    utils::{check_nsequence_rbf, After, Older, SecpCtx, TxDetails},
```
ValuedMammal, line 75: @thunderbiscuit Can we `pub use utils::TxDetails` along with the other re-exports? The reason is that the `wallet::utils` module isn't fully public, so I got this warning when I tried to build the docs. That probably explains why we didn't see a warning earlier when the fields were undocumented.

```
warning: public documentation for `tx_details` links to private item `TxDetails`
   --> wallet/src/wallet/mod.rs:864:19
    |
864 |     /// Get the [`TxDetails`] of a wallet transaction.
    |                   ^^^^^^^^^ this item is private
    |
    = note: this link will resolve properly if you pass `--document-private-items`
    = note: `#[warn(rustdoc::private_intra_doc_links)]` on by default
```

ValuedMammal:
ACK 6aef191f7e70980ee5df5dc34c950abc3ba111d1

PR #208, opened 2025-04-12
deps!: update `bdk_chain` to 0.22.0

### LINKED ISSUE #202: `TxBuilder` continously tries to spend missing output, `Wallet` unable to spend any funds
**Describe the bug**  
We got a report of a user's user whose LDK Node instance is unable to spend any on-chain funds due to the `TxBuilder` always trying to spend an output that's absent from the chain. Hence, the wallet receives a `sendrawtransaction RPC error: {\"code\":-25,\"message\":\"bad-txns-inputs-missingorspent\"}` error when trying to broadcast the resulting transaction.

The wallet is synced against an Esplora backend, and indeed syncs via `EsploraExt::sync`/`Wallet::apply_update` are finishing successfully, but the wallet doesn't seem to be picking up on the missing output. They also tried forcing a `full_scan`, which unfortunately didn't result in any improvement.

**To Reproduce**  
I don't know what happened, but one theory would be that there was a reorg and the wallet somehow didn't handle it correctly.

**Expected behavior**  
The wallet picking up that the transaction is absent from the chain, and `TxBuilder` not trying to spend it.

**Build environment**  
 - LDK Node 0.4.2, which depends on:

```
bdk_chain = { version = "=0.19.0", default-features = false, features = ["std"] }
bdk_esplora = { version = "=0.18.0", default-features = false, features = ["async-https-rustls"]}
bdk_wallet = { version = "=1.0.0-beta.4", default-features = false, features = ["std", "keys-bip39"]}
```

### LINKED ISSUE #224: [deps] Update bdk_chain to 0.22.0

ValuedMammal:
This PR updates `bdk_chain` dependency to the latest version 0.22.0

### Notes to the reviewers

This PR doesn't make use of `CanonicalizationParams` for any advanced use cases - instead we rely on the default params for all tx-graph queries.

fixes #202
fixes #224

### Changelog notice

- deps: bump `bdk_chain` to 0.22.0
- deps: bump `bdk_file_store` to 0.20.0

#### Added

- Added `start_sync_with_revealed_spks_at`
- Added `start_full_scan_at`
- Added `apply_evicted_txs`

#### Changed

- `start_sync_with_revealed_spks` is feature gated by "std", as it uses the system time to provide the time of the sync and also the tx last-seen.
- `apply_update` no longer requires std, as the timestamp is provided in the sync / full scan request
- `start_full_scan` is also gated behind "std"
- `start_sync_*` now includes the expected SPK history (via `list_expected_spk_txids`)

#### Removed

- Removed `apply_update_at`

#### BREAKING

- `bdk_core::TxUpdate` is non-exhaustive bitcoindevkit/bdk#1838
- For the optional file_store feature, the Load variant of the `FileStoreError` is changed to hold a new inner type bitcoindevkit/bdk#1684

#### Changes to persisted data

The `tx_graph::ChangeSet` gained a new default-able field `last_evicted` to persist the fact that a tx is no longer part of the canonical chain.

#### All Submissions:

* [x] This pull request breaks the existng API

### DIFF
--- a/wallet/src/test_utils.rs
+++ b/wallet/src/test_utils.rs
@@ -4,7 +4,7 @@ use alloc::string::ToString;
 use alloc::sync::Arc;
 use core::str::FromStr;
 
-use bdk_chain::{tx_graph, BlockId, ConfirmationBlockTime};
+use bdk_chain::{BlockId, ConfirmationBlockTime, TxUpdate};
 use bitcoin::{
     absolute, hashes::Hash, transaction, Address, Amount, BlockHash, FeeRate, Network, OutPoint,
     Transaction, TxIn, TxOut, Txid,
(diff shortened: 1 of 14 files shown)


evanlinjin:
Why is this part of the 1.3 milestone and not 2.0?

wallet/src/wallet/mod.rs
```diff
@@ -2397,16 +2369,44 @@ impl Wallet {
 
 /// Methods to construct sync/full-scan requests for spk-based chain sources.
 impl Wallet {
+    /// Create a partial [`SyncRequest`] for all revealed spks at `start_time`.
+    ///
+    /// The `start_time` is used to record the time that a mempool transaction was last seen
+    /// (or evicted). See [`Wallet::start_sync_with_revealed_spks`] for more.
+    pub fn start_sync_at(&self, start_time: u64) -> SyncRequestBuilder<(KeychainKind, u32)> {
```
evanlinjin, line 2376:
```suggestion
    pub fn start_sync_with_revealed_spks_at(&self, start_time: u64) -> SyncRequestBuilder<(KeychainKind, u32)> {
```
For consistency.

wallet/src/test_utils.rs
```diff
@@ -312,43 +312,43 @@ pub fn insert_checkpoint(wallet: &mut Wallet, block: BlockId) {
         .unwrap();
 }
 
-/// Insert transaction
+/// Insert a transaction at the current system time.
```
evanlinjin, line 315:
```suggestion
/// Inserts a transaction into the local view, assuming it is currently present in the mempool.
///
/// This can be used, for example, to track a transaction immediately after it is broadcast.
```

notmandatory:
I renamed the old milestone so had to put this one back on the new 2.0.

ValuedMammal:
I added a commit to expose the `insert_evicted_at` logic for the Wallet type 3c5fa0bfb3ae05a02b42618483d2aa9b5fc4ef06.

> - [ ] Update `example_wallet_rpc` to use new `expected_mempool_txids` functionality.

I think this could be done in a follow up PR.

evanlinjin (commented):
Looks good, just a small nit.

Also, I think we could also include eviction logic in `example_wallet_rpc`. I've done so in 4a5207464d8db8f804cc8131c02e474cb8907bcb

ValuedMammal:
Thanks @evanlinjin

notmandatory (approved):
ACK e12501414eaa1fe74b3940d1feacae0bb1b637c4

PR #213, opened 2025-04-19
chore: create and apply rustfmt.toml

### LINKED ISSUE #3: Add project `rustfmt.toml` file
**Describe the enhancement**  

Add a project wide `rustfmt.toml` file with configs:

[format_code_in_doc_comments](https://rust-lang.github.io/rustfmt/?version=v1.6.0&search=#format_code_in_doc_comments)
[wrap_comments](https://rust-lang.github.io/rustfmt/?version=v1.6.0&search=#wrap_comments)

**Use case**  

A project level config file will make it easier for devs to create uniform comments with formatted code and text.

**Additional context**

See: bitcoindevkit/rust-esplora-client#95

https://rust-lang.github.io/rustfmt/?version=v1.6.0&search=
https://github.com/rust-lang/rustfmt?tab=readme-ov-file#configuring-rustfmt

luisschwab:
### Description

Closes #3.

This PR adds basic formatting configuration for uniform comments. Note that the nightly toolchain is necessary.

Changes:
- Add `rustfmt.toml` with this config:
```
comment_width = 100
format_code_in_doc_comments = true
wrap_comments = true
```
- Apply `rustfmt.toml`.
- Update the `fmt` CI job to use the nightly toolchain.
- Update PR template to use nightly toolchain (`cargo fmt` -> `cargo +nightly fmt`).

Note: I had to add `#[rustfmt::skip]` to `test_extract_satisfaction_timelock()` because the base64 PSBT would get wrapped due to the slashes.

#### All Submissions:

* [X] I've signed all my commits
* [X] I followed the [contribution guidelines](https://github.com/bitcoindevkit/bdk/blob/master/CONTRIBUTING.md)
* [X] I ran `cargo +nightly fmt` and `cargo clippy` before committing

#### Bugfixes:

* [X] I'm linking the issue being fixed by this PR

### DIFF
--- a/rustfmt.toml
+++ b/rustfmt.toml
@@ -1,0 +1,3 @@
+comment_width = 100
+format_code_in_doc_comments = true
+wrap_comments = true
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -20,7 +20,7 @@ of the PR were done in a specific way -->
 
 * [ ] I've signed all my commits
 * [ ] I followed the [contribution guidelines](https://github.com/bitcoindevkit/bdk/blob/master/CONTRIBUTING.md)
-* [ ] I ran `cargo fmt` and `cargo clippy` before committing
+* [ ] I ran `cargo +nightly fmt` and `cargo clippy` before committing
 
 #### New Features:
 
--- a/.github/workflows/cont_integration.yml
+++ b/.github/workflows/cont_integration.yml
@@ -111,7 +111,6 @@ jobs:
         run: cargo check --target wasm32-unknown-unknown --no-default-features --features miniscript/no-std,bdk_chain/hashbrown
 
   fmt:
-    needs: prepare
     name: Rust fmt
     runs-on: ubuntu-latest
     steps:
@@ -122,12 +121,12 @@ jobs:
       - name: Install Rust toolchain
         uses: actions-rs/toolchain@v1
         with:
-            toolchain: ${{ needs.prepare.outputs.rust_version }}
+            toolchain: nightly
             override: true
             profile: minimal
             components: rustfmt
       - name: Check fmt
-        run: cargo fmt --all -- --config format_code_in_doc_comments=true --check
+        run: cargo fmt --all --check
 
   clippy_check:
     needs: prepare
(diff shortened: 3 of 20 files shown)


notmandatory:
I didn't realize this would require +nightly. Any idea when the `fmt` features we need will be in a `stable` release?

notmandatory (approved):
ACK 0b5e3f6ab4feb2efec1cfbdddbb19bb2d61af3ae

It's a little ackward to have to use `cargo +nightly fmt` but since it only needs to be done before pushing the PR and will make all the docs and docs tests more consistent I'd still like to do it.

luisschwab:
> Any idea when the fmt features we need will be in a stable release?

Not sure, but once it's in stable we can revert it.

By the way, I made an identical PR on `bdk` as well.

ValuedMammal:
Concept ACK

If we do this I think we should remove the extra `--config` option from the fmt job in CI now that it's in the config file. I guess it technically doesn't need the "prepare" job either because we won't be getting the toolchain from the `rust-version` file.

For the future I would look at increasing some of the [granular width settings](https://github.com/rust-lang/rustfmt/blob/master/Configurations.md#use_small_heuristics) such as `chain_width` to prevent method chains from going vertical too quickly. Can also take some inspiration from rust-bitcoin [rustfmt.toml](https://github.com/rust-bitcoin/rust-bitcoin/blob/master/rustfmt.toml).

> I ran `cargo +nightly fmt` and `cargo clippy` before committing

You mean I ran `just fmt` 😄

luisschwab:
> If we do this I think we should remove the extra --config option from the fmt job in CI now that it's in the config file. 

True.

> I guess it technically doesn't need the "prepare" job either because we won't be getting the toolchain from the rust-version file.

Maybe just comment it out so it's easier to revert when these options become stable?

> You mean I ran just fmt 😄

We need one of those!

ValuedMammal:
I don't understand what `wrap_comments` is doing though, is it supposed to also respect the value of `comment_width`?

luisschwab:
I got this from the rustfmt docs:

>[comment_width](https://rust-lang.github.io/rustfmt/?version=v1.8.0&search=#comment_width)
Maximum length of comments. No effect unless wrap_comments = true.

notmandatory (approved):
ACK e736cf9a9742f82c8f34c627f6ca1814c2a9d325

rustfmt.toml
```diff
@@ -0,0 +1,3 @@
+comment_width = 140
```
notmandatory, line 1: This should be changed to 100 to match bitcoindevkit/bdk#1946 then this one should be good to go in.
```suggestion
comment_width = 100
```

notmandatory: And it also needs a rebase to pickup latest CI changes.

luisschwab: Done.

luisschwab:
Rebased to trigger CI.

Issue #166, opened 2023-01-31
Let `TxBuilder` avoid previously used UTXOs (UTXO locking)


tnull:
**Describe the enhancement**  
When sequentially creating multiple transactions, `TxBuilder` might currently reuse the same UTXOs, potentially exposing us to double-spend ourselves if we want to bring all of these generated transactions on-chain. It would be nice if the `TxBuilder` could track the used UTXOs internally, and allow to auto-unlock them after a set amount of time.

**Use case**  
This is important for sequentially opening multiple Lightning channels, e.g., in LDK Node.

notmandatory:
The basic infrastructure was adding in `bdk_chain` `0.22.0` to address the stuck tx due to a double-spent input issue bitcoindevkit/bdk_wallet#202 which will be fixed with `bdk_wallet` `2.0.0`. But support for avoiding or including unbroadcast TX hasn't been exposed yet in the `bdk_wallet` apis. I'll discuss with the team what needs to be done and if this can be done for a `bdk_wallet` `2.1.0.` release.

See draft `bdk_wallet` work  in bitcoindevkit/bdk_wallet#220.

DanGould:
Maybe redundant, but I want to emphasize that payjoin requires utxo locking for unicast transactions transmitted to a counterparty and awaiting their signature or broadcast, not just those broadcast to the p2p network.

notmandatory:
@evanlinjin thanks for that correction, I'm pretty sure I was over-thinking what @tnull is even requesting, which is simply to add convenience functions to lock and unlock UTXOs from being used in the `TxBuilder`. This feature exists in bitcoin core with the RPC commands:

- [lockunspent](https://bitcoincore.org/en/doc/29.0.0/rpc/wallet/lockunspent/): Updates list of temporarily unspendable outputs. Locks can be persisted.
- [listlockunspent](https://bitcoincore.org/en/doc/29.0.0/rpc/wallet/listlockunspent/): Returns list of temporarily unspendable outputs.

[...]

PR #220, opened 2025-04-25
Test canonicalization mods API

evanlinjin:
### Description

This PR builts on top of #208 and bitcoindevkit/bdk#1808.

The goal is to evaluate whether the changes introduced in bitcoindevkit/bdk#1808 provide a sufficient foundation to implement and maintain a broadcast queue within `bdk_wallet`.

Requirements:

* Creating a new tx should not replace unbroadcasted txs that were already in the wallet.
- The broadcast queue should support transaction replacements, with the most recently inserted version taking precedence.

### Notes to the reviewers

WIP

### Changelog notice

WIP

#### All Submissions:

#### New Features:

#### Bugfixes:


evanlinjin:
# To Do (as discussed)

- [ ] Unexpose `CanonicalizationParams` from the public API.
- [ ] By default, include the whole broadcast queue in `assume_canonical`.

notmandatory:
@evanlinjin is it be possible to create a version of this feature, possibly without persistence changes, in a non-breaking way for the `bdk_wallet` `2.1.0` release? Or is there some other way to fix #166 with the current `bdk_wallet` + `bdk_chain` ?

evanlinjin:
@notmandatory I don't think this can be done in a non-breaking way, as the only way something like this would be useful is if we persist the data (thus modifying the changeset). I could finish this PR soon though

evanlinjin:
Replaced by #257

Issue #168, opened 2023-01-18
BIP 329 "Wallet Labels Export Format" Support


gabridome:
The spec is new but it seems not invasive.
I find particularly interesting the outputs's labels.
https://github.com/bitcoin/bips/blob/master/bip-0329.mediawiki

Musab1258:
Hi everyone, I participated in a program organized by ChaincodeLabs, the BOSS Program. And when I got to the Proof of Concept (POC) stage, I picked this issue as my POC project since it was part of the listed projects.

I can see that a lot of discussion had happened on this issue and @storopoli has made a PR which he eventually closed. And although I have started working on this issue, I want to to ask if anyone is currently not working on this issue so I can keep on working on it and make a PR.

notmandatory:
This may only need an example using the library @praveenperera created here: https://github.com/bitcoinppl/bip329

And/or this would be a useful new feature to add to the `bdk-cli` example wallet which was recently updated to use bdk 1.0.

Musab1258:
@notmandatory, I made a PR for an example using the library here #222. Can you please help me review it.

Musab1258:
> This may only need an example using the library [@praveenperera](https://github.com/praveenperera) created here: https://github.com/bitcoinppl/bip329
> 
> And/or this would be a useful new feature to add to the `bdk-cli` example wallet which was recently updated to use bdk 1.0.

@notmandatory To seek clarification, if I am to add this feature to bdk-cli based on your suggestion, I should also make use of the bip329 library right?

Leeyah-123:
Hello @notmandatory, can I get assigned this?

notmandatory:
in #509 @ValuedMammal proposed:
> provide the Rust types with serde support; users manage persistence

I like this approach, I'd suggest maybe the new types we need are already in the `bip329` crate and all we need are docs and examples for how to use it with `bdk_wallet`. Could also use `bdk-cli` as an example. Or @ValuedMammal did you mean we only need to add serde support to some rust bdk types?

ValuedMammal:
I was suggesting to "in-house" the bip329 implementation and forego adding a dependency.

schjonhaug:
Sharing a production-app data point, in case it helps the persistence discussion.

[Canary](https://github.com/schjonhaug/canary) is a BDK-based watch-only wallet that recently shipped BIP-329 transaction labels (user-editable names, plus JSONL import/export). A few constraints from that work:

[...]

PR #222, opened 2025-04-29
example: add an example using the BIP329 crate with bdk_wallet

Musab1258:
### Description

This PR adds a new example file, `bip329_example.rs,` that demonstrates how to manage BIP-329 wallet labels when using `bdk_wallet`.

This fixes issue #168 by using @praveenperera's [bip329 crate](https://crates.io/crates/bip329) rather than integrating label storage directly into the core `bdk_wallet`. This follows the feedback I received from @notmandatory on the issue.

The example shows how to:

- Create an in-memory `bdk_wallet::Wallet`.
- Load/Save labels from/to a separate JSON Lines (`.jsonl`) file using the `bip329` crate.
- Build a lookup map to correlate wallet items (addresses, txids, outpoints) with their labels.
- Demonstrate basic addition/updating of labels in memory and saving them back to the file.
- Uses `tempfile` to avoid leaving artifacts after running.

### Notes to the reviewers

- This example requires the `keys-bip39` feature to generate the mnemonic and descriptors used.

#### All Submissions:

#### New Features:

#### Bugfixes:

### DIFF
--- a/wallet/Cargo.toml
+++ b/wallet/Cargo.toml
@@ -46,6 +46,7 @@ bdk_wallet = { path = ".", features = ["rusqlite", "file_store", "test-utils"] }
 bdk_file_store = { version = "0.18.1" }
 anyhow = "1"
 rand = "^0.8"
+bip329 = "0.4.0"
 
 [package.metadata.docs.rs]
 all-features = true
@@ -60,3 +61,8 @@ required-features = ["all-keys"]
 name = "miniscriptc"
 path = "examples/compiler.rs"
 required-features = ["compiler"]
+
+[[example]]
+name = "bip329_example"
+path = "examples/bip329_example.rs"
+required-features = ["keys-bip39"]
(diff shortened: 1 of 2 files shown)


wallet/Cargo.toml
```diff
@@ -46,6 +46,8 @@ bdk_wallet = { path = ".", features = ["rusqlite", "file_store", "test-utils"] }
 bdk_file_store = { version = "0.18.1" }
 anyhow = "1"
 rand = "^0.8"
+bip329 = "0.4.0"
+log = "0.4"
```
notmandatory, line 50: ⛏️ small Nit, since all our other examples use `println!` I'd do that for this example too and remove the `log` dependency. Even though it's just a dev dependency it's still a dependency that's not needed.

Musab1258: Hi @notmandatory, thanks for the review. I have substituted the logs with `println!`

Issue #192, opened 2020-12-30
Inconsistent checking of RBF rules


LLFourn:
[BIP125] rules 3 and 4 are not being checked properly according to my understanding:

https://github.com/bitcoindevkit/bdk/blob/c2b2da7601fb570952cfd858f3821b3a4fd87ec9/src/wallet/mod.rs#L730-L746

It should be checking rules 3 and 4 regardless of the fee policy chosen in txbuilder. 

[BIP125]: https://github.com/bitcoin/bips/blob/master/bip-0125.mediawiki#Implementation_Details

PR #225, opened 2025-05-02
Fix Inconsistent checking of RBF rules.

Anunayj:
### Description
This PR adds checks based on RBF policy ([BIP125](https://github.com/bitcoin/bips/blob/master/bip-0125.mediawiki) and [Bitcoin Core's Replacement Policy](https://github.com/bitcoin/bitcoin/blob/master/doc/policy/mempool-replacements.md)). 
More specifically it checks transactions:
1. Have a higher absolute fee than the original transaction
2. Pay enough for their bandwidth.
3. Have a higher feerate than conflicting transactions.

Fixes: #192 

### Notes to the reviewers

The addition of `satisfaction_weight` to `CoinSelectionResult` would be a breaking change, it isn't entirely necessary, as I could duplicate that functionality in wallet itself by making an `estimated_weight(psbt: Psbt)` function, but this was one of the cleaner ways to go about this, and I'm not entirely sure how acceptable breaking changes are in the project.

#### Changed
- RBF Policies (BIP125 and Bitcoin Core's) are now checked correctly when creating a transaction with `bumping_fee`.

#### BREAKING

- `CoinSelectionResult` now has a `satisfaction_weight` field, giving the max weight required to satisfy the selected inputs

#### All Submissions:

#### New Features:

#### Bugfixes:

### DIFF
--- a/wallet/src/wallet/coin_selection.rs
+++ b/wallet/src/wallet/coin_selection.rs
@@ -47,17 +47,19 @@
 //!     ) -> Result<CoinSelectionResult, coin_selection::InsufficientFunds> {
 //!         let mut selected_amount = Amount::ZERO;
 //!         let mut additional_weight = Weight::ZERO;
+//!         let mut satisfaction_weight = Weight::ZERO;
 //!         let all_utxos_selected = required_utxos
 //!             .into_iter()
 //!             .chain(optional_utxos)
 //!             .scan(
-//!                 (&mut selected_amount, &mut additional_weight),
-//!                 |(selected_amount, additional_weight), weighted_utxo| {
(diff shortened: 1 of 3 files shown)


rustaceanrob:
Nice job, although I am not sure how to tinker with this to make it non-breaking, but otherwise looks good. Would it be possible to migrate this to `bdk_tx`? Because it hasn't become an issue in production, it might be best to put this in the new module instead

evanlinjin:
There has already been work on this in https://github.com/bitcoindevkit/bdk-tx/pull/1. Please take a look there. Reviews and tests will be helpful.

Note that we are trying to put a feature-freeze on `TxBuilder` in `bdk_wallet` and want to focus effort in `bdk_tx`.

PR #226, opened 2025-05-03
Explorations for support of N keychains

thunderbiscuit:
Just pushing here to show/discuss as I explore the potential for support of any number of descriptors for the Wallet type for maybe a 3.0.

# Overview

The gist of my idea is a `KeyRing` (KeychainRing? KeychainSet?), which is basically a structure that holds a vector of keychains. You create a key ring with any number of descriptors (_must have_ 1 default, _can have_ 1 change, plus any number of others). The wallet constructors take this KeyRing as an argument, and from there the wallet methods need to be adjusted to work with the new paradigm.

The `KeyRing` has methods to add/check/query descriptors it holds, and is where the domain logic of ensuring only 1 descriptor is the default, that all descriptors use the same network, and any other requirements we have for basically this group of keychains lives. The wallet type then gets a new field key_ring, which it can use to query things like `KeyRing::get_default_descriptor()` or `KeyRing::list_keychains()`.

```rust
pub enum KeychainKind {
    Default,
    Change,
    Other(DescriptorId),
}

pub type WalletKeychain = (KeychainKind, Descriptor<DescriptorPublicKey>);

pub struct KeyRing {
    keychains: Vec<WalletKeychain>,
}
```

# Resulting API

The API is showcased in the `wallet/examples/n_keychains.rs` file.

```rust
// Create a keyring with a single, default descriptor (aka the KeychainKind::External from the 1.2.0 API)
let mut keyring = KeyRing::new(EXTERNAL_DESC, Network::Testnet4);

// Add 3 new custom descriptors
keyring.add_other_descriptor(OTHER_DESC_21);
keyring.add_other_descriptor(OTHER_DESC_31);
keyring.add_other_descriptor(OTHER_DESC_41);

let keychain_ids: Vec<DescriptorId> = keyring.list_keychain_ids();
println!("{:?}", keychain_ids);

// Create the wallet and peek addresses on each of the descriptors
let wallet: Wallet = Wallet::new(keyring).create_wallet_no_persist()?;
let address_1 = wallet.peek_address(KeychainKind::Default, 0).unwrap();
let address_2 = wallet.peek_address(KeychainKind::Other(keychain_ids[1]), 0).unwrap();
let address_3 = wallet.peek_address(KeychainKind::Other(keychain_ids[2]), 0).unwrap();
let address_4 = wallet.peek_address(KeychainKind::Other(keychain_ids[3]), 0).unwrap();

println!("Address 1 {:?} at index {:?} on keychain {:?}", address_1.address, address_1.index, address_1.keychain);
println!("Address 2 {:?} at index {:?} on keychain {:?}", address_2.address, address_2.index, address_2.keychain);
println!("Address 3 {:?} at index {:?} on keychain {:?}", address_3.address, address_3.index, address_3.keychain);
println!("Address 4 {:?} at index {:?} on keychain {:?}", address_4.address, address_4.index, address_4.keychain);
```

### DIFF
--- a/wallet/src/wallet/tx_builder.rs
+++ b/wallet/src/wallet/tx_builder.rs
@@ -253,19 +253,19 @@ impl<'a, Cs> TxBuilder<'a, Cs> {
     ///
     /// # Ok::<(), anyhow::Error>(())
     /// ```
-    pub fn policy_path(
-        &mut self,
-        policy_path: BTreeMap<String, Vec<usize>>,
-        keychain: KeychainKind,
-    ) -> &mut Self {
-        let to_update = match keychain {
-            KeychainKind::Internal => &mut self.params.internal_policy_path,
-            KeychainKind::External => &mut self.params.external_policy_path,
-        };
-
-        *to_update = Some(policy_path);
-        self
-    }
+    // pub fn policy_path(
+    //     &mut self,
+    //     policy_path: BTreeMap<String, Vec<usize>>,
+    //     keychain: KeychainKind,
+    // ) -> &mut Self {
+    //     let to_update = match keychain {
+    //         KeychainKind::Internal => &mut self.params.internal_policy_path,
+    //         KeychainKind::External => &mut self.params.external_policy_path,
+    //     };
+    //
+    //     *to_update = Some(policy_path);
+    //     self
+    // }
 
     /// Add the list of outpoints to the internal list of UTXOs that **must** be spent.
     ///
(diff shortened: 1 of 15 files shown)


thunderbiscuit:
Some development notes as I go along:

1. The `CreateParams` struct had 2 constructors: `CreateParams::new` and `CreateParams::new_single`. This simplifies the constructors because the `KeyRing` type handles all that stuff now.

wallet/src/wallet/tx_builder.rs
```diff
@@ -836,8 +837,8 @@ impl ChangeSpendPolicy {
     pub(crate) fn is_satisfied_by(&self, utxo: &LocalOutput) -> bool {
         match self {
             ChangeSpendPolicy::ChangeAllowed => true,
-            ChangeSpendPolicy::OnlyChange => utxo.keychain == KeychainKind::Internal,
-            ChangeSpendPolicy::ChangeForbidden => utxo.keychain == KeychainKind::External,
+            ChangeSpendPolicy::OnlyChange => utxo.keychain == KeychainKind::Change,
+            ChangeSpendPolicy::ChangeForbidden => utxo.keychain == KeychainKind::Default,
```
nymius, line 841: I was expecting the behavior of `ChangeForbidden` logic to not allow change, rather than deriving it to `UTxOs` owned by the `External` keychain. I may be mislead by thinking in terms of coin selection.

PR #229, opened 2025-05-09
Allow serialization of Update

icota:
### Description

The rationale behind this is to allow a separate instance of the same wallet be updated on a device that's not necessarily connected. Like a hardware wallet.

A connected device gets the response, serializes it with serde and passes the `Update` to the air-gapped one.

Depends on https://github.com/bitcoindevkit/bdk/pull/1954

#### All Submissions:


thunderbiscuit:
We should make sure this is closed by a human rather than auto-closed by a bot. This feature would be useful for the bindings in one way or another, so I'm reopening.

Issue #188, opened 2021-11-30
BDK Descriptor Improvements: Add multi descriptor capability in wallet


rajarshimaitra:
**Description**  
  
In order for BDK to extend beyond just a Bitcoin wallet, it needs a more powerful descriptor capability. This project aims at improving the descriptors used in the library. Currently BDK uses only a single descriptor for a single wallet. Giving a new descriptor to BDK will create a brand new wallet.

The goal of this project is to create a wallet that can handle multiple descriptors within itself (a "super wallet"), and can send funds from any or all of them together.   

**Expected Outcomes**  
 

- Through understanding of [descriptors](https://github.com/bitcoin/bitcoin/blob/master/doc/descriptors.md).
- Basic understanding of the [rust-miniscript](https://github.com/rust-bitcoin/rust-miniscript) library.
- Multi descriptor capability in BDK. 

**Resources**  
  
  
  
  
  

**Skills Required**  
  
  
  
  

- Experience with git. [Guide](https://git-scm.com/book/en/v2)
- Basic familiarity with rust. [First seven chapters of the book](https://doc.rust-lang.org/book/)
- Familiarity with BDK’s descriptor usage.

**Mentor(s)**  

**Difficulty**: Hard  

**Competency Test (optional)**  

- Install rust, compile and run all bdk examples and tests. 
- Read through the BDK docs.
- Make a dummy wallet with BDK with different blockchain backends. 
- Familiarity with basic rust, should be able to write basic custom trait implementations on foreign structures.

notmandatory:
(dup'ing my comment from #230 here for reference)
Here are my high-level goals for N keychain support. 
1. support importing/exporting bitcoincore wallets with bdk_wallet ( see [importdescriptors], [listdescriptors])
2. enable improved wallet privacy per #28 
3. make it easier to migrate from old to new descriptors
 
[importdescriptors]: https://bitcoincore.org/en/doc/29.0.0/rpc/wallet/importdescriptors/
[listdescriptors]: https://bitcoincore.org/en/doc/29.0.0/rpc/wallet/listdescriptors/

Issue #28, opened 2025-02-05
Wallet Privacy Improvements


notmandatory:
This is a tracking issue for wallet privacy features mentioned but not completed in bitcoindevkit/bdk#918. Some of these will require bitcoindevkit/bdk#486. 

> I'm doing some bitcoin core repo archeology to identify privacy related wallet features. I started with the excellent [core onboarding wallet write up](https://github.com/chaincodelabs/onboarding-to-bitcoin-core/blob/master/04_wallet.adoc) from @willcl-ark, which mentioned choosing an "appropriate" change script type. I have identified the below features BDK's wallet API should support. Some of these we already support, the rest I'll make new BDK issues for.  Please comment if you know of any I'm missing.

- [x] bitcoin/bitcoin#2340, fixed in bitcoindevkit/bdk#611
- [ ] bitcoin/bitcoin#12119
- [ ] bitcoin/bitcoin#12257, WIP bitcoindevkit/bdk#1851
- [ ] bitcoin/bitcoin#13756
- [ ] bitcoin/bitcoin#13808
- [ ] bitcoin/bitcoin#16512
- [ ] bitcoin/bitcoin#16528, the part about 6 descriptors is useful
- [ ] bitcoin/bitcoin#17621
- [ ] bitcoin/bitcoin#18418, WIP bitcoindevkit/bdk#1851
- [ ] bitcoin/bitcoin#23789
- [ ] bitcoin/bitcoin#24494
- [ ] bitcoin/bitcoin#24584 

 _Originally posted by @notmandatory in [#918](https://github.com/bitcoindevkit/bdk/issues/918#issuecomment-1491221046)_

tvpeter:
I will be working on https://github.com/bitcoin/bitcoin/pull/13756

PR #230, opened 2025-05-14
wip: Support N keychains

ValuedMammal:
Draft PR for supporting N keychains, details to follow.

### DIFF
--- a/wallet/src/wallet/mod.rs
+++ b/wallet/src/wallet/mod.rs
@@ -30,8 +30,8 @@ use bdk_chain::{
         SyncResponse,
     },
     tx_graph::{CalculateFeeError, CanonicalTx, TxGraph, TxUpdate},
-    BlockId, ChainPosition, ConfirmationBlockTime, DescriptorExt, FullTxOut, Indexed,
-    IndexedTxGraph, Indexer, Merge,
+    BlockId, ChainPosition, ConfirmationBlockTime, DescriptorExt, DescriptorId, FullTxOut, Indexed,
+    IndexedTxGraph, Indexer as _, Merge,
 };
 use bitcoin::{
     absolute,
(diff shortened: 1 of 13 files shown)


wallet/src/wallet/mod.rs
```diff
@@ -104,20 +104,179 @@ pub struct Wallet {
...
+    Default,
+    PartialEq,
+    Eq,
+    PartialOrd,
+    Ord,
+    Hash,
+    serde::Serialize,
+    serde::Deserialize,
+)]
+pub struct Keychain(pub(crate) usize);
```
evanlinjin, line 130: I'm wondering if `usize` being variably sized on different platforms will cause issues with serialization.

thunderbiscuit:
Wow this is taking shape! Quick comments/questions as I'm looking at the diff (sorry if it's jumping the gun and you're actually in the middle of adding new stuff as I review)!

1. I like the renaming of APIs like `Wallet::peek_address(KeychainKind)` to `Wallet::peek_default_address()` and `Wallet::next_unused_address(KeychainKind) to `Wallet::next_default_unused_address()`.
2. The idea of grouping the descriptors is potentially neat, but I would like to flesh out the use case in my mind a bit more because of all the complexity it potentially adds. Could that be used for APIs that would do something like "build me a transaction using keychain group 4"? If so, then the descriptors _within-groups_ might also need default/change flags. Is there something else you'd use them for?
3. APIs like `Wallet::mark_used(keychain: u32, index: u32)` feel less clear to me at the moment. Did you prefer the u32 over the DescriptorId for a reason? Why the `u32` instead of something like `Wallet::mark_used(keychain: DescriptorId, index: u32)`? Also, I see the group is a u32, so in this case would you not also need to specify the group + the keychain number (to uniquely identify a keychain)? Something like `Wallet::mark_used(keychain_group: u32, keychain: u32, index: u32)`? I might be misunderstanding something here.
4. I see the comment on the `KeychainGroup` type: `Descriptors in the same group share certain characteristics such as the descriptor type.` Is that something you'd like to enforce? I'm not sure if there would be usecases for them not being of the same type. I guess I'm not yet clear on the groups.

Some APIs I could see being useful on the `KeyRing`:
- `KeyRing::get_default_keychain`
- `KeyRing::get_change_keychain`
- `KeyRing::new_from_multipath`
- `KeyRing::from_v1_changeset`
- `KeyRing::from_v2_changeset`

nymius (commented):
Some thoughts I don't want to miss about this:

- We could enforce some expected format for descriptors to be accepted in a `KeyRing`. Mainly to avoid ambiguity and the aggregated complexity of many border cases. I would suggest to only consider descriptors that have wildcards. For cases like multipath, I would split the descriptor first into multiple descriptors.

- I would prefer a flat structure rather than nested. For the case of Internal descriptors, and thinking in terms of a table, I would add a field to link through the descriptor Id the one used for change. That field would be optional, if the descriptor isn't expected to create change, can be elided. Also, multiple keychains could refer to the same change keychain.

- Continuing with the idea of a flat structure, the external/internal consideration could be reduced to a boolean. If a keychain is linked to another keychain to use as change by the descriptor id, the second keychain should have the boolean field "external" enabled.

- I think the behavior desired by the use of group ids could also be achieved just by looking at the properties of the descriptor. Getting balance from all the descriptors of the same type should be easily achieved by filtering by their variant. In the case of multipath descriptors the balance can be obtained by filtering by the descriptor fingerprint. In more complex query cases that don't necessarily share any of the common traits a descriptor can have, I would leave the implementation for higher layers.

Thanks for working on this! Please, consider these comments as mere suggestions.

ValuedMammal:
Last time I talked to thunderbiscuit we agreed on possibly doing away with the extra "grouping" logic, and for simplicity go back to using `DescriptorId` as the keychain. I'm also exploring the "clean break" approach with some sort of upgrade path.

thunderbiscuit:
I've also been playing around with the "group" idea in my head and I can't see a super strong case for it yet. I think those kinds of grouping would be built at the application layer in a structure that would map descriptor ids to groups.

I suggest we build without, and bring them back if more review brings to light more information.

nymius:
I'm convinced by thunderbiscuit arguments, let's move this logic up to the application layer, and keep the interface simpler.

The only previous comment I still hold is the following.
>  We could enforce some expected format for descriptors to be accepted in a `KeyRing`.
> Mainly to avoid ambiguity and the aggregated complexity of many border cases.
> I would suggest to only consider descriptors that have wildcards
> For cases like multipath, I would split the descriptor first into multiple descriptors.

notmandatory:
Here are my high-level goals for N keychain support (originally called multi-descriptor support in #188). 
1. support importing/exporting bitcoincore wallets with bdk_wallet ( see [importdescriptors], [listdescriptors])
2. enable improved wallet privacy per #28 
3. make it easier to migrate from old to new descriptors
 
[importdescriptors]: https://bitcoincore.org/en/doc/29.0.0/rpc/wallet/importdescriptors/
[listdescriptors]: https://bitcoincore.org/en/doc/29.0.0/rpc/wallet/listdescriptors/

ValuedMammal:
Closing in favor of #318.

PR #206, opened 2025-04-10
feat: example using hwi signing with

sdmg15:
### Description

Integrates a new BDK signer using [async-hwi](https://github.com/wizardsardine/async-hwi/).
The included example uses Bitbox02 but can be easily extended to use any other supported device.

### Notes for reviewer
It's related to issue #20 

#### All Submissions:

* [x] I've signed all my commits
* [x] I followed the [contribution guidelines](https://github.com/bitcoindevkit/bdk/blob/master/CONTRIBUTING.md)
* [x] I ran `cargo fmt` and `cargo clippy` before committing

PR #235, opened 2025-05-19
refactor(wallet)!: remove signers and it's APIs, relying on rust-bitcoin's `Psbt::sign` instead.

### LINKED ISSUE #70: Remove signers from bdk_wallet
**Describe the enhancement**  

This is a tracking issue to remove signing capabilities from [`Wallet`](https://docs.rs/bdk_wallet/latest/bdk_wallet/struct.Wallet.html#) and provide a new way to sign and finalize PSBTs using `rust-bitcoin` [`Psbt::sign`](https://docs.rs/bitcoin/latest/bitcoin/psbt/struct.Psbt.html#method.sign) and implementing the [`GetKey`](https://docs.rs/bitcoin/latest/bitcoin/psbt/trait.GetKey.html#) trait for relevant types. Please comment below or mention this issue in related issues and PRs.

**Use case**  

Safer and more flexible approaches to signing PSBTs can be done outside of the `Wallet` type. 

**Additional context**

- [ ] related issues and PRs here...
- [x] bitcoindevkit/bdk-ffi#469
- [ ] rust-bitcoin/rust-miniscript#765
- [ ] bitcoindevkit/bdk#1484

oleonardolima:
fixes #70 

### Description

It's a work in progress to remove signers from `Wallet` and all related APIs. Users should use their own signer implementation, which needs to implement the `GetKey` trait from rust-bitcoin in order to rely on the `Psbt::sign` method.

Adds a reference implementation of a new `SignerWrapper` type as a test utility, which is currently being used on existing tests, instead of `Wallet::sign`.

#### All Submissions:

#### New Features:

#### Bugfixes:

### DIFF
--- a/examples/common/mod.rs
+++ b/examples/common/mod.rs
@@ -1,0 +1,133 @@
+//! Common signing utilities for bdk_wallet examples
+//!
+//! This module provides the `SignerWrapper` struct and related utilities
+//! that enable signing functionality for the wallet examples. These utilities
+//! wrap the KeyMap type to implement the GetKey trait, allowing examples
+//! to sign transactions and PSBTs.
+//!
+//! Note: This module is only required temporarily until miniscript 12.x is released,
+//! which will include signing capabilities for KeyMap natively.
+
+use miniscript::descriptor::{DescriptorSecretKey, KeyMap};
+use std::collections::BTreeMap;
+
+use bitcoin::{
+    key::Secp256k1,
+    psbt::{GetKey, GetKeyError, KeyRequest},
+};
+
+#[derive(Debug, Clone)]
+/// A wrapper over the [`KeyMap`] type that has the `GetKey` trait implementation for signing.
+pub struct SignerWrapper {
+    key_map: KeyMap,
+}
+
+impl SignerWrapper {
+    /// Creates a new [`SignerWrapper`] for the given [`KeyMap`].
+    pub fn new(key_map: KeyMap) -> Self {
+        Self { key_map }
+    }
+}
+
+impl GetKey for SignerWrapper {
+    type Error = GetKeyError;
+
+    fn get_key<C: bitcoin::secp256k1::Signing>(
+        &self,
+        key_request: KeyRequest,
+        secp: &bitcoin::key::Secp256k1<C>,
+    ) -> Result<Option<bitcoin::PrivateKey>, Self::Error> {
+        for key_map in self.key_map.iter() {
+            let (_, desc_sk) = key_map;
+            let wrapper = DescriptorSecretKeyWrapper::new(desc_sk.clone());
+            match wrapper.get_key(key_request.clone(), secp) {
+                Ok(Some(private_key)) => return Ok(Some(private_key)),
+                Ok(None) => continue,
+                // TODO: (@leonardo) how should we handle this ?
+                // we can't error-out on this, because the valid signing key can be in the next
+                // iterations.
+                Err(_) => continue,
+            }
+        }
+        Ok(None)
+    }
+}
+
+/// Wrapper for DescriptorSecretKey to implement GetKey trait
+pub struct DescriptorSecretKeyWrapper(DescriptorSecretKey);
+
+impl DescriptorSecretKeyWrapper {
+    /// Creates a new DescriptorSecretKeyWrapper
+    pub fn new(desc_sk: DescriptorSecretKey) -> Self {
+        Self(desc_sk)
+    }
+}
+
+impl GetKey for DescriptorSecretKeyWrapper {
+    type Error = GetKeyError;
+
+    fn get_key<C: bitcoin::secp256k1::Signing>(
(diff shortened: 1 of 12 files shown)


ValuedMammal:
These are some of my high-level impressions, granted the details may be subject to change.

- I hesitate about keeping the sign/finalize flow in the example crates while `SignerWrapper` is still in its infancy. If the idea of the issue is to simultaneously "provide a new way to sign", can that be achieved with the async-hwi example #206 ?
- Some tests that depend on the existence of signers can be changed to either sign using `SignerWrapper`, or else simulate signing by adding a fake sig / satisfaction weight. However my opinion is that any tests that were specifically written to test the old signing code (`signer.rs`) should probably just be removed.
- `finalize_psbt` is kind of a mess. `SignOptions::assume_height` is arguably not needed because `After` and `Older` are basically obsolete, but correct me if I'm wrong. The only possible error in `finalize_psbt` is `SignerError::IndexOutOfRange` - we can make a new error type for this. #267
- Make signer module `pub(crate)` and un-expose `SignOptions`, `SignerError`. For the moment `SignersContainer` is still being used.
- In terms of review it would be great if this PR was just a net-removal of code, and then have a potential followup with the experimental `SignerWrapper`, but I understand the motivation for having it here.

wallet/src/wallet/mod.rs
```diff
@@ -1319,13 +1220,22 @@ impl Wallet {
         let internal_descriptor = keychains.get(&KeychainKind::Internal);
 
         let external_policy = external_descriptor
-            .extract_policy(&self.signers, BuildSatisfaction::None, &self.secp)?
+            .extract_policy(
+                &SignersContainer::default(),
+                BuildSatisfaction::None,
+                &self.secp,
+            )?
```
ValuedMammal, line 1227: Following the idea of `SignersContainer::default`, in `make_generic_signature` we can set the contribution to None and remove the `signers` parameter from the function, and any function that calls it. The rationale is that a wallet with no signers won't be contributing a signature to the overall satisfaction.

This is a larger refactor but has the benefit of ultimately decoupling the signer module.

https://github.com/bitcoindevkit/bdk_wallet/blob/00dafe728c0e6666ee9c664afd547b00e4d3ed8c/wallet/src/descriptor/policy.rs#L763-L779

oleonardolima: Yes, I agree with the rationale and it's better than having to keep the `SignersContainers` only for it's default implementation.

I've started playing with this, you can take a look here: https://github.com/oleonardolima/bdk_wallet/commit/e9f6b0a38e3ebadeebd26c3560561dfccb55a873. I'm taking a look on the failing tests, to check if the breaking changes are no big deal.

ValuedMammal:
I thought about it and I don't mind having `SignerWrapper` as a test utility if it can be used in tests and examples.

oleonardolima:
Thanks for taking a look into this one, I still need to rebase and update some parts of it after backporting rust-miniscript#851.

> Is SignerWrapper going to remain a test utility? Or will it develop into some sort of default signer? One which we can always just create an instance of and pass it to the psbt sign method.

No, the idea is to use directly the implementation from rust-bitcoin/rust-miniscript. Yes, if you need to just wrap it for some reasone you'll need to create an isntance of it an pass for each test, instead of having a re-implementation of it in the test-utils.

> Before user just needed to pass the psbt to wallet.sign() without worrying about signer implementation and all. (sorry to drag you back if this had been talked about already)

It'd still be basically the same, the user won't need to care about the implementation of signing itself, just to have a type that implements a `GetKey`, if for some reason the user needs to use a different type that is not covered in rust-miniscript/rust-bitcoin, then it'll need to implement it's own GetKey, extracting the required key properly.

ItoroD:
> Thanks for taking a look into this one, I still need to rebase and update some parts of it after backporting rust-miniscript#851.
> 
> > Is SignerWrapper going to remain a test utility? Or will it develop into some sort of default signer? One which we can always just create an instance of and pass it to the psbt sign method.
> 
> No, the idea is to use directly the implementation from rust-bitcoin/rust-miniscript. Yes, if you need to just wrap it for some reasone you'll need to create an isntance of it an pass for each test, instead of having a re-implementation of it in the test-utils.
> 
> > Before user just needed to pass the psbt to wallet.sign() without worrying about signer implementation and all. (sorry to drag you back if this had been talked about already)
> 
> It'd still be basically the same, the user won't need to care about the implementation of signing itself, just to have a type that implements a `GetKey`, if for some reason the user needs to use a different type that is not covered in rust-miniscript/rust-bitcoin, then it'll need to implement it's own GetKey, extracting the required key properly.

Ah! Thanks for explanation!

So some of wrappers where sort of place holders for the updates now done in 851  rust-miniscript. This Pr was done before that was merged. Makes sense.

Look forward to seeing the update on here!

oleonardolima:
stale, superseded by #505

PR #247, opened 2025-05-27
deps: Update `bdk_chain` to 0.23.0

### LINKED ISSUE #237: Fix panic in `create_tx`
This looks like an unnecessarily eager use of `.expect` that will eventually cause a panic if `reveal_to_target` happens to return `None`.

https://github.com/bitcoindevkit/bdk_wallet/blob/7cdc9852faab7f193715bf49b73c29516b2db79f/wallet/src/wallet/mod.rs#L1542-L1551

### LINKED ISSUE #246: [deps] Update bdk_chain
This should be done for 2.0.0-alpha.

ValuedMammal:
### Description

The PR updates bdk_chain to 0.23.0. Additionally we introduce the ability to persist derived SPKs indexed by descriptor ID and derivation index.

The `CreateParams::use_spk_cache` method enables or disables a persistent cache for script pubkeys (SPKs) derived by the wallet. When enabled, the wallet will store and reuse previously derived SPKs, avoiding redundant derivation on subsequent loads.

When a wallet has many revealed addresses (i.e., many derived SPKs), loading the wallet can become slow if every address must be re-derived from scratch.

By enabling the SPK cache:

- **On wallet creation:** The wallet will persistently store each derived SPK as addresses are revealed.
- **On wallet load:** If `use_spk_cache` is also set in the corresponding `LoadParams`, the wallet will load the cached SPKs directly from storage, skipping the need to re-derive them from the descriptor. This can dramatically reduce load times for wallets with hundreds or thousands of revealed addresses, as the expensive crypto operations are avoided.

**Caveat:**  
Both creation and loading must have `use_spk_cache(true)` set for the cache to be used. If not, the wallet will fall back to deriving SPKs as usual.

fixes #246 
fixes #237 

#### Added

- `CreateParams::use_spk_cache`
- `LoadParams::use_spk_cache`

#### Changed

- bump bdk_chain to 0.23.0
- **Note:** This change extends the wallet `ChangeSet` type by adding `first_seen` to the tx_graph member, and adding `spk_cache` to the indexer.

#### All Submissions:

### DIFF
--- a/wallet/src/wallet/coin_selection.rs
+++ b/wallet/src/wallet/coin_selection.rs
@@ -750,6 +750,7 @@ mod test {
             value,
             index,
             ChainPosition::Unconfirmed {
+                first_seen: Some(last_seen),
                 last_seen: Some(last_seen),
             },
         )
@@ -850,7 +851,10 @@ mod test {
                             transitively: None,
                         }
                     } else {
-                        ChainPosition::Unconfirmed { last_seen: Some(0) }
+                        ChainPosition::Unconfirmed {
+                            first_seen: Some(1),
+                            last_seen: Some(1),
+                        }
                     },
                 }),
             });
@@ -875,7 +879,10 @@ mod test {
                     keychain: KeychainKind::External,
                     is_spent: false,
                     derivation_index: 42,
-                    chain_position: ChainPosition::Unconfirmed { last_seen: Some(0) },
+                    chain_position: ChainPosition::Unconfirmed {
+                        first_seen: Some(1),
+                        last_seen: Some(1),
+                    },
                 }),
             })
             .collect()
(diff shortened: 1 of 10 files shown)


wallet/src/wallet/coin_selection.rs
```diff
@@ -850,7 +851,10 @@ mod test {
                             transitively: None,
                         }
                     } else {
-                        ChainPosition::Unconfirmed { last_seen: Some(0) }
+                        ChainPosition::Unconfirmed {
+                            first_seen: Some(1),
+                            last_seen: Some(1),
```
notmandatory, line 856: I know this is just a test value but is there a reason you changed it from 0 to 1? With new code does it now have to always be greater than 0?

ValuedMammal: No reason, I think any value will do.

oleonardolima (commented):
utACK 8ad15dd345e87d9c43edb9f89bddd4df5ec61d5d

ValuedMammal:
I rebased and added a commit 6aba330e as a fix for #237.

PR #251, opened 2025-06-02
feat(signer): merge internal and external signer containers

### LINKED ISSUE #186: BDK should store signers in the same list
Right now BDK stores signers in two different `SignersContainer`s. One for the internal keychain and one for the external.
This structurally precludes signers that can sign for both (which is most!). Also BDK doesn't use this separation for anything (both internal and external are always called to sign).

I think these two separate lists should be merged into one. There might be some more design work around this where we store the signers in the list in such a way that given a derivation path we can quickly find the signer for that path (regardless of whether it is for external or internal addresses).

ndungudedan:
Currently BDK stores signers in two separate SignersContainer instances - one for internal keychain and one for external. This separation is unnecessary since:
1. Most signers can sign for both keychains
2. Both containers are always called during signing
3. The separation adds complexity without benefits

This change:
- Adds an `extend` method to SignersContainer to support merging containers
- Keeps a single signer container in Wallet

Closes #186

### Notes to the reviewers

- Looking for a concept ACK to validate that my approach is okay. 
- 3 tests fail with `Incompatible change descriptor`, from the export module. On further investigation, I see that the export node returns the change descriptor as a public key while the derived change descriptor is a private key. I am not sure why this is case and a second review would be great:
```rust
change_descriptor: Some("tr([73c5da0a/86'/0'/0']tprv8fMn4hSKPRC1oaCPqxDb1JWtgkpeiQvZhsr8W2xuy3GEMkzoArcAWTfJxYb6Wj8XNNDWEjfYKK4wGQXh3ZUXhDF2NcnsALpWTeSwarJt7Vc/1/*)")

export.change_descriptor(): Some("tr([73c5da0a/86'/0'/0']tpubDC3pD7UZXnsgh3EBjbtBQiB1FnLask7UHBSunZ1DPK4dCFFZoFRkgxHB8gt42FvLzx1DpxfHWxAsYaY6b643RVcGjDxXxns7wKKYnnfEcbB/1/*)")
```

#### All Submissions:

#### New Features:

#### Bugfixes:

### DIFF
--- a/wallet/src/wallet/export.rs
+++ b/wallet/src/wallet/export.rs
@@ -119,11 +119,7 @@ impl FullyNodedExport {
     ) -> Result<Self, &'static str> {
         let descriptor = wallet
             .public_descriptor(KeychainKind::External)
-            .to_string_with_secret(
-                &wallet
-                    .get_signers(KeychainKind::External)
-                    .as_key_map(wallet.secp_ctx()),
-            );
+            .to_string_with_secret(&wallet.get_signers().as_key_map(wallet.secp_ctx()));
         let descriptor = remove_checksum(descriptor);
         Self::is_compatible_with_core(&descriptor)?;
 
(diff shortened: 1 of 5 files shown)


oleonardolima:
Thanks for working on this one, though, for the next breaking change release (v3.0), the plan is to remove the signers altogether from the Wallet API, see #70 and #235. That said, it supersedes the changes from this PR and issue #186, as it's also a breaking change.

ndungudedan:
Thanks! Closing the PR.

PR #252, opened 2025-06-02
fix: make sure tx_build does not change the order of add_utxos when u…

### LINKED ISSUE #244: TxOrdering::Untouched no longer ensures the order of tx input
**Describe the bug**  
Since add_utxo of build_tx in 1.2.0 uses hashmap to contain input, TxOrdrering cannot ensure that the order is not changed.

**To Reproduce**  
```rust
    let utxos = wallet1
        .list_unspent()
        .map(|o| o.outpoint)
        .collect::<Vec<_>>();
    let builder = wallet1.build_tx();
    builder
        .ordering(bdk_wallet::TxOrdering::Untouched)
        .fee_rate(FeeRate::from_sat_per_vb(5).unwrap())
        .add_utxo(utxos[0])
        .unwrap()
        .add_utxo(utxos[1])
        .unwrap()
        .add_recipient(
            wallet1.peek_address(External, 0).address.script_pubkey(),
            bitcoin::Amount::from_sat(1000),
        );

    let inputs = builder.finish().unwrap().unsigned_tx.input;
    // can't assert
    assert!(inputs[0].previous_output == utxos[0]);
    assert!(inputs[1].previous_output == utxos[1]);
```

**Expected behavior**  
```rust
    assert!(inputs[0].previous_output == utxos[0]);
    assert!(inputs[1].previous_output == utxos[1]);
```

**Build environment**  
 - BDK tag/commit: 1.2.0
 - OS+version:   
 - Rust/Cargo version: *
 - Rust/Cargo target: *

**Additional context**  
https://github.com/bitcoindevkit/bdk_wallet/blob/5817ed07f220c787a26e234307f8308d3ab6165e/crates/wallet/src/wallet/tx_builder.rs#L128

eauxxs:
fixes #244

### DIFF
--- a/wallet/Cargo.toml
+++ b/wallet/Cargo.toml
@@ -26,7 +26,6 @@ bdk_chain = { version = "0.23.0", features = [ "miniscript", "serde" ], default-
 # Optional dependencies
 bip39 = { version = "2.0", optional = true }
 bdk_file_store = { version = "0.21.0", optional = true }
-indexmap = "2.9.0"
 
 [features]
 default = ["std"]
--- a/wallet/Cargo.toml
+++ b/wallet/Cargo.toml
@@ -26,6 +26,7 @@ bdk_chain = { version = "0.23.0", features = [ "miniscript", "serde" ], default-
 # Optional dependencies
 bip39 = { version = "2.0", optional = true }
 bdk_file_store = { version = "0.21.0", optional = true }
+indexmap = "2.9.0"
 
 [features]
 default = ["std"]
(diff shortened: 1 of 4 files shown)


nymius:
Thanks for working on this! I'd prefer to not add the `indexmap` dependency just to cover this case. I've proposed an alternative solution in #261

nymius:
Thanks for this! I still think is a workaround for the dependency issue. You are moving the problem of maintaining compatibility with the dependency to maintaining the `OrderUtxos` structure ourselves. Is a solution, but I still think we can do better.

oleonardolima:
This can be closed, superseded by #262.

PR #259, opened 2025-06-07
feat!: Persist utxo lock status

### LINKED ISSUE #166: Let `TxBuilder` avoid previously used UTXOs (UTXO locking)

ValuedMammal:
### Description

Added `Wallet::lock_unspent`, `unlock_unspent`, and updated the wallet `ChangeSet` to persist information about wallet UTXOs including the lock status of an outpoint.

may resolve #166.

#### All Submissions:

#### New Features:

#### Bugfixes:

### DIFF
--- a/wallet/src/wallet/changeset.rs
+++ b/wallet/src/wallet/changeset.rs
@@ -1,11 +1,16 @@
+use alloc::collections::BTreeMap;
+
 use bdk_chain::{
     indexed_tx_graph, keychain_txout, local_chain, tx_graph, ConfirmationBlockTime, Merge,
 };
+use bitcoin::{OutPoint, Txid};
 use miniscript::{Descriptor, DescriptorPublicKey};
 use serde::{Deserialize, Serialize};
 
 type IndexedTxGraphChangeSet =
     indexed_tx_graph::ChangeSet<ConfirmationBlockTime, keychain_txout::ChangeSet>;
+
+use crate::UtxoInfo;
 
 /// A change set for [`Wallet`]
 ///
@@ -114,6 +119,8 @@ pub struct ChangeSet {
     pub tx_graph: tx_graph::ChangeSet<ConfirmationBlockTime>,
     /// Changes to [`KeychainTxOutIndex`](keychain_txout::KeychainTxOutIndex).
     pub indexer: keychain_txout::ChangeSet,
+    /// Changes to utxo info.
+    pub utxo_info: BTreeMap<OutPoint, UtxoInfo>,
 }
 
 impl Merge for ChangeSet {
@@ -142,6 +149,11 @@ impl Merge for ChangeSet {
             self.network = other.network;
         }
 
+        // To merge utxo_info we just extend the existing collection. If there's
+        // an existing entry for a given outpoint, then it is overwritten by the
+        // new utxo info.
+        self.utxo_info.extend(other.utxo_info);
+
         Merge::merge(&mut self.local_chain, other.local_chain);
         Merge::merge(&mut self.tx_graph, other.tx_graph);
         Merge::merge(&mut self.indexer, other.indexer);
@@ -154,6 +166,7 @@ impl Merge for ChangeSet {
             && self.local_chain.is_empty()
             && self.tx_graph.is_empty()
             && self.indexer.is_empty()
+            && self.utxo_info.is_empty()
     }
 }
 
@@ -163,6 +176,8 @@ impl ChangeSet {
     pub const WALLET_SCHEMA_NAME: &'static str = "bdk_wallet";
     /// Name of table to store wallet descriptors and network.
     pub const WALLET_TABLE_NAME: &'static str = "bdk_wallet";
+    /// Name of table to store wallet utxo info.
+    pub const WALLET_UTXO_TABLE_NAME: &'static str = "bdk_wallet_utxo_info";
 
     /// Get v0 sqlite [ChangeSet] schema
     pub fn schema_v0() -> alloc::string::String {
@@ -177,12 +192,25 @@ impl ChangeSet {
         )
     }
 
+    /// Get v1 sqlite [ChangeSet] schema. Schema v1 adds a table for wallet UTXO info.
+    pub fn schema_v1() -> alloc::string::String {
+        format!(
+            "CREATE TABLE {} ( \
+                txid TEXT NOT NULL, \
+                vout INTEGER NOT NULL, \
+                is_locked INTEGER, \
+                PRIMARY KEY(txid, vout) \
+                ) STRICT;",
+            Self::WALLET_UTXO_TABLE_NAME,
+        )
+    }
+
     /// Initialize sqlite tables for wallet tables.
     pub fn init_sqlite_tables(db_tx: &chain::rusqlite::Transaction) -> chain::rusqlite::Result<()> {
         crate::rusqlite_impl::migrate_schema(
             db_tx,
             Self::WALLET_SCHEMA_NAME,
-            &[&Self::schema_v0()],
+            &[&Self::schema_v0(), &Self::schema_v1()],
         )?;
 
         bdk_chain::local_chain::ChangeSet::init_sqlite_tables(db_tx)?;
@@ -218,6 +246,27 @@ impl ChangeSet {
             changeset.descriptor = desc.map(Impl::into_inner);
             changeset.change_descriptor = change_desc.map(Impl::into_inner);
             changeset.network = network.map(Impl::into_inner);
+        }
+
+        // Load utxo info.
+        let mut stmt = db_tx.prepare(&format!(
+            "SELECT txid, vout, is_locked FROM {}",
+            Self::WALLET_UTXO_TABLE_NAME,
+        ))?;
+        let rows = stmt.query_map([], |row| {
+            Ok((
+                row.get::<_, Impl<Txid>>("txid")?,
+                row.get::<_, u32>("vout")?,
+                row.get::<_, bool>("is_locked")?,
+            ))
+        })?;
+        for row in rows {
+            let (Impl(txid), vout, is_locked) = row?;
+            let utxo_info = UtxoInfo {
+                outpoint: OutPoint::new(txid, vout),
+                is_locked,
+            };
+            changeset.utxo_info.insert(utxo_info.outpoint, utxo_info);
         }
 
         changeset.local_chain = local_chain::ChangeSet::from_sqlite(db_tx)?;
@@ -265,6 +314,20 @@ impl ChangeSet {
             network_statement.execute(named_params! {
                 ":id": 0,
                 ":network": Impl(network),
+            })?;
+        }
+
+        // Persist utxo info.
+        let mut stmt = db_tx.prepare_cached(&format!(
+            "INSERT INTO {}(txid, vout, is_locked) VALUES(:txid, :vout, :is_locked) ON CONFLICT DO UPDATE SET is_locked = :is_locked",
+            Self::WALLET_UTXO_TABLE_NAME,
+        ))?;
+        for (&outpoint, utxo_info) in &self.utxo_info {
+            let OutPoint { txid, vout } = outpoint;
+            stmt.execute(named_params! {
+                ":txid": Impl(txid),
+                ":vout": vout,
+                ":is_locked": utxo_info.is_locked,
             })?;
         }
 
--- a/wallet/src/wallet/mod.rs
+++ b/wallet/src/wallet/mod.rs
@@ -109,6 +109,7 @@ pub struct Wallet {
     stage: ChangeSet,
     network: Network,
     secp: SecpCtx,
+    utxo_info: BTreeMap<OutPoint, UtxoInfo>,
 }
 
 /// An update to [`Wallet`].
@@ -449,6 +450,7 @@ impl Wallet {
         let change_descriptor = index.get_descriptor(KeychainKind::Internal).cloned();
         let indexed_graph = IndexedTxGraph::new(index);
         let indexed_graph_changeset = indexed_graph.initial_changeset();
+        let utxo_info = BTreeMap::new();
 
         let stage = ChangeSet {
             descriptor,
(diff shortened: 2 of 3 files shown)


ChidiChuks:
How do you handle UTXO locking in concurrent wallet instances?

nymius (commented):
Could you expand a little bit more on how this mechanism could be used in broadcast scenarios?
From the test I guess all the logic of why a UTxO becomes available again is kept by the lib user? Like to be unlocked after a height in the future if there is not transaction spending it in the blockchain.

This is a competing solution for #257, right?

evanlinjin:
> This is a competing solution for #257, right?

@nymius My understanding is that this is NOT a competing solution with #257. UTXO-locking will not fully solve #166 as we want unbroadcasted outputs to be available for selection. In fact, UTXO-locking is already available by using `TxBuilder::add_unspendable`. The only difference here is that locked outpoints are persisted with BDK. So this is more of a convenience feature that users are requesting.

wallet/src/wallet/mod.rs
```diff
@@ -2580,6 +2685,16 @@ impl AsRef<bdk_chain::tx_graph::TxGraph<ConfirmationBlockTime>> for Wallet {
     }
 }
 
+/// Information about a wallet UTXO.
+#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, serde::Serialize, serde::Deserialize)]
+pub struct UtxoInfo {
```
ValuedMammal, line 2690: I originally called this `UtxoInfo` with the idea that it may be extended with more information, but since the structure itself doesn't qualify as a "sub-changeset", I think it should only contain the information related to the utxo lock.

wallet/src/wallet/mod.rs
```diff
@@ -653,18 +657,39 @@ impl Wallet {
...
+            .map(|output| {
+                (
+                    output.outpoint,
+                    UtxoInfo {
+                        outpoint: output.outpoint,
+                        is_locked: false,
+                    },
+                )
+            })
+            .collect();
```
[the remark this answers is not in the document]
ValuedMammal: The data in `utxo_info` / `locked_outpoints` can exist independently of what's returned by `list_unspent`, therefore I removed the extraneous logic involving `list_unspent`.

wallet/src/wallet/mod.rs
```diff
@@ -2377,6 +2404,84 @@ impl Wallet {
         &self.chain
     }
 
+    /// Is UTXO locked. `None` if the outpoint isn't found in `self.utxo_info`.
+    pub fn is_utxo_locked(&self, outpoint: OutPoint) -> Option<bool> {
```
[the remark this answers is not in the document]
ValuedMammal: Changed to return `bool`.

ValuedMammal:
@evanlinjin I took all of your suggestions other than I think the proposed `Wallet::list_locked_unspents` is not yet implemented.

wallet/src/wallet/changeset.rs
```diff
@@ -177,12 +192,26 @@ impl ChangeSet {
...
     }
 
+    /// Get v1 sqlite [`ChangeSet`] schema. Schema v1 adds a table for locked outpoints.
+    pub fn schema_v1() -> alloc::string::String {
+        format!(
+            "CREATE TABLE {} ( \
+                txid TEXT NOT NULL, \
+                vout INTEGER NOT NULL, \
+                is_locked INTEGER, \
+                expiration_height INTEGER, \
```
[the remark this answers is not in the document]
ValuedMammal: The use case I had in mind was to prevent a utxo from being selected for a specific duration of time measured in blocks. As mentioned, it frees the user from having to explicitly unlock it in the future.

wallet/src/wallet/changeset.rs
```diff
@@ -177,12 +188,26 @@ impl ChangeSet {
         )
     }
 
+    /// Get v1 sqlite [`ChangeSet`] schema. Schema v1 adds a table for locked outpoints.
+    pub fn schema_v1() -> alloc::string::String {
+        format!(
+            "CREATE TABLE {} ( \
+                txid TEXT NOT NULL, \
+                vout INTEGER NOT NULL, \
+                is_locked INTEGER, \
```
notmandatory, line 197: I understand why `is_locked` is needed in the ChangeSet but I don't think we need an `is_locked` column. I'd expect that if we persist a ChangeSet with a `false` here we'd remove the record from the table rather than updated it with a `is_locked=false`. Otherwise this table would grow unbounded for un-locked locks that don't have an `expiration_date`.

wallet/src/wallet/changeset.rs
```diff
@@ -220,6 +245,31 @@ impl ChangeSet {
...
+                row.get::<_, Option<u32>>("expiration_height")?,
+            ))
+        })?;
+        let locked_outpoints = &mut changeset.locked_outpoints;
+        for row in rows {
+            let (Impl(txid), vout, is_locked, expiration_height) = row?;
+            let outpoint = OutPoint::new(txid, vout);
+            locked_outpoints
+                .locked_outpoints
+                .insert(outpoint, is_locked);
```
notmandatory, line 267: If we remove `is_locked` you'd only have locked outpoints in the db table and this would always be "true".
```suggestion
                .insert(outpoint, true);
```

wallet/src/wallet/changeset.rs
```diff
@@ -268,6 +318,35 @@ impl ChangeSet {
...
 
+        // Insert locked outpoints.
+        let mut stmt = db_tx.prepare_cached(&format!(
+            "INSERT INTO {}(txid, vout, is_locked) VALUES(:txid, :vout, :is_locked) ON CONFLICT DO UPDATE SET is_locked=:is_locked",
+            Self::WALLET_OUTPOINT_LOCK_TABLE_NAME,
+        ))?;
+        let locked_outpoints = &self.locked_outpoints.locked_outpoints;
+        for (&outpoint, is_locked) in locked_outpoints.iter() {
+            let OutPoint { txid, vout } = outpoint;
+            stmt.execute(named_params! {
```
notmandatory, line 329: Also if we remove the `is_locked` column then you'd need the insert_stmt and a delete_stmt to remove outpoints where the changes has `is_locked=true`.

wallet/src/wallet/mod.rs
```diff
@@ -2377,6 +2402,104 @@ impl Wallet {
         &self.chain
     }
 
+    /// Get a reference to the locked outpoints.
+    pub fn locked_outpoints(&self) -> &BTreeMap<OutPoint, UtxoLock> {
+        &self.locked_outpoints
+    }
+
+    /// List unspent outpoints that are currently locked.
+    pub fn list_locked_unspent(&self) -> impl Iterator<Item = OutPoint> + '_ {
```
notmandatory, line 2411: ⛏️ a nit, but it's a little confusing to refer to the outpoints as "unspent" here and "outpoints" in the other new functions. Since we already have the "list_unspent" function would it make sense to use "unspent" for  all the related functions (ie. locked_unspent, is_unspent_locked, lock_unspent, unlock_unspent)? Even though we all know what "outpoints" mean for those unfamiliar with bitcoin internals it's not a very descriptive term.

wallet/src/wallet/mod.rs
```diff
@@ -2377,6 +2402,104 @@ impl Wallet {
...
+            }
+        };
+
+        self.stage.merge(ChangeSet {
+            locked_outpoints: changeset,
+            ..Default::default()
+        });
+    }
+
+    /// Unlock the wallet output of the specified `outpoint`.
```
notmandatory, line 2478: You should also warn the users as you did with `lock_outpoint` that these changes need to be persisted.

wallet/src/wallet/mod.rs
```diff
@@ -2377,6 +2385,102 @@ impl Wallet {
...
+                    };
+                    self.stage.merge(ChangeSet {
+                        locked_outpoints: [(outpoint, utxo_lock)].into(),
+                        ..Default::default()
+                    });
+                }
+            }
+        }
+    }
+
```
evanlinjin, line 2502: What do you think about an `unlock_all_outpoints` method? Just in case the caller needs to reset their state.

wallet/src/wallet/changeset.rs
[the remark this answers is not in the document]
evanlinjin: > Yes, I agree with everything you said. It's just that if you don't offer that API, the user will implement such a thing themselves, as at some point the user would need unlock previously-locked UTXOs for which the spend 'failed'.

The purpose of the `IntentTracker` API (as proposed in #257) is to address this issue.

> The use case I had in mind was to prevent a utxo from being selected for a specific duration of time measured in blocks. As mentioned, it frees the user from having to explicitly unlock it in the future.

I think we need a more thorough explanation on how this would be useful just because we have to maintain what we include.

One potential example I can think of is in payjoin. The sender keeps around the Original PSBT until a given timeout or they receive the Proposal PSBT. During this time, they lock the outpoints that fund these PSBTs. However, once the PJ transaction is broadcasted, there will be no need to keep these outpoints locked - and we can unlock them instantly. In other words, no need for auto-unlocking of outpoints.

wallet/src/wallet/mod.rs
```diff
@@ -2580,6 +2684,25 @@ impl AsRef<bdk_chain::tx_graph::TxGraph<ConfirmationBlockTime>> for Wallet {
...
 }
 
+/// Records the lock status of a wallet UTXO (outpoint).
+///
+/// Only 1 [`UtxoLock`] may be applied to a particular outpoint at a time. Refer to the docs
+/// for [`Wallet::lock_outpoint`]. Note that the lock status is user-defined and does not take
+/// into account any timelocks directly encoded by the descriptor.
+#[derive(
+    Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, serde::Serialize, serde::Deserialize,
+)]
```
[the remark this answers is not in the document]
evanlinjin: Okay I see what you mean. We don't want new fields to interfere with the logic in `lock_outpoint`.

evanlinjin (commented):
Implementation looks good. Just need a bit more clarity on `expiration_height` before ACK.

wallet/src/wallet/mod.rs
evanlinjin: I think it is okay since `unspent` here is short for `unspent_outpoint` whereas `outpoint` elsewhere is short for `outpoint`s which can be unspent/spent.

ValuedMammal:
- Changed Wallet `locked_outpoints` field to HashMap.
- Rebased

ValuedMammal:
In the latest push [compare](https://github.com/bitcoindevkit/bdk_wallet/compare/e3d3013044e59326f00b167bb2877e91a47e37c2..8edfcb75e377976e5570beb7493ecf94ffcd962e#diff-102dde0c4d3810d508aa61eed40caf17e10f79d4c30b3356a30bea258f8d1fb1)

- Using `INSERT OR INGORE` instead of `REPLACE INTO` to handle existing rows in `WALLET_OUTPOINT_LOCK_TABLE_NAME`
- Rebased

clippy.toml
```diff
@@ -1,4 +1 @@
-# TODO fix, see: https://rust-lang.github.io/rust-clippy/master/index.html#large_enum_variant
-enum-variant-size-threshold = 1032
-# TODO fix, see: https://rust-lang.github.io/rust-clippy/master/index.html#result_large_err
-large-error-threshold = 993
\ No newline at end of file
+msrv = "1.85.0"
```
nymius: Why this cannot go into a different PR?

src/persist_test_utils.rs
```diff
@@ -216,6 +217,7 @@ where
         local_chain: local_chain_changeset,
         tx_graph: tx_graph_changeset,
         indexer: keychain_txout_changeset,
+        locked_outpoints: Default::default(),
```
nymius: Same here.

nymius:
To not start diverging from the format used in practice so far, I would rewrite 834b6cdccc5c1fcb1929ba5833c93feebbad182f to [test(utils): Update persist_wallet_changeset to persist locked outpo](https://github.com/bitcoindevkit/bdk_wallet/pull/259/commits/834b6cdccc5c1fcb1929ba5833c93feebbad182f)

Also, as you moved the clippy msrv change to 12aed94449c714a49ce02d71a2ae6bf71babb86a, could you expand a little in the commit message about the relation between the clippy changes and the msrv flag there?

ValuedMammal:
> To not start diverging from the format used in practice so far, I would rewrite [834b6cd](https://github.com/bitcoindevkit/bdk_wallet/commit/834b6cdccc5c1fcb1929ba5833c93feebbad182f) to [test(utils): Update persist_wallet_changeset to persist locked outpo](https://github.com/bitcoindevkit/bdk_wallet/pull/259/commits/834b6cdccc5c1fcb1929ba5833c93feebbad182f)

I decided to go with `feat(test-utils): ...` since it adds functionality to the `persist_test_utils` module.

> Also, as you moved the clippy msrv change to [12aed94](https://github.com/bitcoindevkit/bdk_wallet/commit/12aed94449c714a49ce02d71a2ae6bf71babb86a), could you expand a little in the commit message about the relation between the clippy changes and the msrv flag there?

In 730fdb24 I restored `clippy.toml` back to the way it was (roughly) prior to #242. The `msrv` setting is likely redundant, as clippy can use the rust-version specified in Cargo.toml, however the config file would be empty otherwise. I don't have a strong preference about what to do with it, what do you think?

evanlinjin (approved):
ACK d4eaf4244f0a97349f5b66ef726b4c8d42531c28

ValuedMammal:
Dropped the msrv line from `clippy.toml`. @nymius

ValuedMammal:
Thank you to the reviewers @evanlinjin @notmandatory @nymius @tnull

PR #261, opened 2025-06-09
TxBuilder should not change order of insertion

nymius:
### Description

On my attempt to fix bitcoindevkit/bdk#1794 in bitcoindevkit/bdk#1798, I broke the assumption that insertion order is preserved when `TxBuilder::ordering` is `TxOrdering::Untouched`. Some users are relying in this assumption so here I'm trying to restore it back, without adding a new dependency for this single use case like #252 , and trying to not redesign the `TxBuilder::utxos` interface again.

However, I'm open to do that if there are strong reasons for it. Opening as Draft to gather opinions.

Fixes #244

cc: @eauxxs

### Changelog notice

No public APIs are changed by these commits.

### Checklists

> [!IMPORTANT]
> This pull request **DOES NOT** break the existing API
* [x] I've signed all my commits
* [x] I followed the [contribution guidelines](https://github.com/bitcoindevkit/bdk/blob/master/CONTRIBUTING.md)
* [x] I ran `cargo +nightly fmt` and `cargo clippy` before committing
* [x] I've added tests for the new code
* [x] I've expanded docs addressing new code
* [x] I've added tests to reproduce the issue which are now passing
* [x] I'm linking the issue being fixed by this PR

PR #262, opened 2025-06-10
Preserve insertion order of manually selected utxos if `TxOrdering::Untouched`

### LINKED ISSUE #244: TxOrdering::Untouched no longer ensures the order of tx input

nymius:
### Description

On my attempt to fix bitcoindevkit/bdk#1794 in bitcoindevkit/bdk#1798, I broke the assumption that insertion order is preserved when `TxBuilder::ordering` is `TxOrdering::Untouched`. Some users are relying in this assumption, so here I'm trying to restore it back, without adding a new dependency for this single use case like #252, or creating a new struct just to track this.

In this fourth alternative solution I'm going back to use `Vec` to store the manually selected UTxOs.

I was reluctant to do it in this way because `HashMap` seems a better solution giving its property of avoiding duplicates, but as we also want to keep the secuential nature of the insertion of UTxOs in `TxBuilder`, here is an alternative aligned with that principle. 

May replace #252 
May replace #261 .
Fixes #244

### Notes to the reviewers

Also, as I was working on this, I came back to the following tests:
- `test_prexisting_foreign_utxo_have_no_precedence_over_local_utxo_with_same_outpoint`
- `test_prexisting_local_utxo_have_precedence_over_foreign_utxo_with_same_outpoint`

Motivated during the implementation and review of bitcoindevkit/bdk#1798.

Which required the underlying structure to also hold the properties of no duplication for manually selected UTxOs, as the structures were accessed directly for these cases.

The test tries to cover the case when there are two wallets using the same descriptor, one tracks transactions and the other does not. The first passes UTxOs belonging to the second one and this one creates transactions using the `add_foreign_utxo` interface.
In this case it was expected for any `LocalUtxo` of the offline wallet to supersede any conflicting foreign UTxO. But, the simulation of this case went against the borrowing constraints of rust.
By how costly was to reproduce this behavior for me in the tests, I would like to have second opinions in the feasibility of the test case.

### Changelog notice

No public APIs are changed by these commits.

### Checklists

> [!IMPORTANT]
> This pull request **DOES NOT** break the existing API
* [x] I've added tests for the new code
* [x] I've expanded docs addressing new code

### DIFF
--- a/wallet/src/wallet/mod.rs
+++ b/wallet/src/wallet/mod.rs
@@ -1511,7 +1511,7 @@ impl Wallet {
 
         let (required_utxos, optional_utxos) = {
             // NOTE: manual selection overrides unspendable
-            let mut required: Vec<WeightedUtxo> = params.utxos.values().cloned().collect();
+            let mut required: Vec<WeightedUtxo> = params.utxos.clone();
             let optional = self.filter_utxos(&params, current_height.to_consensus_u32());
 
             // if drain_wallet is true, all UTxOs are required
@@ -1740,52 +1740,45 @@ impl Wallet {
                     })
                     .map(|(prev_tx, prev_txout, chain_position)| {
                         match txout_index.index_of_spk(prev_txout.script_pubkey.clone()) {
-                            Some(&(keychain, derivation_index)) => (
-                                txin.previous_output,
-                                WeightedUtxo {
-                                    satisfaction_weight: self
-                                        .public_descriptor(keychain)
-                                        .max_weight_to_satisfy()
-                                        .unwrap(),
-                                    utxo: Utxo::Local(LocalOutput {
-                                        outpoint: txin.previous_output,
-                                        txout: prev_txout.clone(),
-                                        keychain,
-                                        is_spent: true,
-                                        derivation_index,
-                                        chain_position,
-                                    }),
-                                },
-                            ),
+                            Some(&(keychain, derivation_index)) => WeightedUtxo {
+                                satisfaction_weight: self
+                                    .public_descriptor(keychain)
+                                    .max_weight_to_satisfy()
+                                    .unwrap(),
+                                utxo: Utxo::Local(LocalOutput {
+                                    outpoint: txin.previous_output,
+                                    txout: prev_txout.clone(),
+                                    keychain,
+                                    is_spent: true,
+                                    derivation_index,
+                                    chain_position,
+                                }),
+                            },
                             None => {
                                 let satisfaction_weight = Weight::from_wu_usize(
                                     serialize(&txin.script_sig).len() * 4
                                         + serialize(&txin.witness).len(),
                                 );
-
-                                (
-                                    txin.previous_output,
-                                    WeightedUtxo {
-                                        utxo: Utxo::Foreign {
-                                            outpoint: txin.previous_output,
-                                            sequence: txin.sequence,
-                                            psbt_input: Box::new(psbt::Input {
-                                                witness_utxo: prev_txout
-                                                    .script_pubkey
-                                                    .witness_version()
-                                                    .map(|_| prev_txout.clone()),
-                                                non_witness_utxo: Some(prev_tx.as_ref().clone()),
-                                                ..Default::default()
-                                            }),
-                                        },
-                                        satisfaction_weight,
+                                WeightedUtxo {
+                                    utxo: Utxo::Foreign {
+                                        outpoint: txin.previous_output,
+                                        sequence: txin.sequence,
+                                        psbt_input: Box::new(psbt::Input {
+                                            witness_utxo: prev_txout
+                                                .script_pubkey
+                                                .witness_version()
+                                                .map(|_| prev_txout.clone()),
+                                            non_witness_utxo: Some(prev_tx.as_ref().clone()),
+                                            ..Default::default()
+                                        }),
                                     },
-                                )
+                                    satisfaction_weight,
+                                }
                             }
                         }
                     })
             })
-            .collect::<Result<HashMap<OutPoint, WeightedUtxo>, BuildFeeBumpError>>()?;
+            .collect::<Result<Vec<WeightedUtxo>, BuildFeeBumpError>>()?;
 
         if tx.output.len() > 1 {
             let mut change_index = None;
@@ -2099,6 +2092,11 @@ impl Wallet {
             vec![]
         // only process optional UTxOs if manually_selected_only is false
         } else {
+            let manually_selected_outpoints = params
+                .utxos
+                .iter()
+                .map(|wutxo| wutxo.utxo.outpoint())
+                .collect::<HashSet<OutPoint>>();
             self.indexed_graph
                 .graph()
                 // get all unspent UTxOs from wallet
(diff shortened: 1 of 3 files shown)


wallet/src/wallet/tx_builder.rs
```diff
@@ -276,29 +276,31 @@ impl<'a, Cs> TxBuilder<'a, Cs> {
...
+            };
+
+            if let Some(idx) = self.params.utxos.iter().position(|wutxo| match wutxo.utxo {
+                Utxo::Local(LocalOutput { outpoint, .. }) => outpoint == *recv_outpoint,
+                Utxo::Foreign { outpoint, .. } => outpoint == *recv_outpoint,
+            }) {
+                self.params.utxos.remove(idx);
+            }
+            self.params.utxos.push(wutxo);
+        }
```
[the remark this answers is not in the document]
nymius: To bring back the original rationale about why favor precedence of new UTxOs over old ones, https://github.com/bitcoindevkit/bdk/pull/1798/files#r1974774295:
> Ideally, we would error here if an `OutPoint` conflicts with a foreign UTXO. However, we will need to add new error variants (which is a MAJOR change since the error type does not have `non-exhaustive`).
> 
> For now, how about we make a note in the docs that if you add the same UTXO as both foreign and local, the latest change has precedence (please word that better).
> 
> Same goes for the add-foreign-utxo methods.

In the light of the above comment, do you think we should still roll back to the old precedence @ValuedMammal? @evanlinjin do you have any opinion on this?

I think we still prefer `Utxo::Local` over `Utxo::Foreign`.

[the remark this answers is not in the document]
ValuedMammal: I guess that would work too.

wallet/src/wallet/tx_builder.rs
```diff
@@ -276,29 +276,45 @@ impl<'a, Cs> TxBuilder<'a, Cs> {
...
-                                    .wallet
-                                    .public_descriptor(output.keychain)
-                                    .max_weight_to_satisfy()
-                                    .unwrap(),
-                                utxo: Utxo::Local(output),
-                            },
-                        )
-                    })
+            .map(|&op| -> Result<_, AddUtxoError> {
+                visited.insert(op);
```
ValuedMammal, line 295: I still think we should avoid creating more of the same `WeightedUtxo` if the outpoint is already visited.

```suggestion
            .filter(|&&op| visited.insert(op))
            .map(|&op| -> Result<_, AddUtxoError> {
```

For example how should tx builder behave in this situation?

```rs
    let (mut wallet, _) = get_funded_wallet_wpkh();
    let utxo = wallet.list_unspent().next().unwrap();
    let op = utxo.outpoint;

    let mut tx_builder = wallet.build_tx();
    tx_builder.add_utxos(&[op, op, op]).unwrap();

    assert_eq!(tx_builder.params.utxos.len(), 1);
```

wallet/src/wallet/tx_builder.rs
```diff
@@ -407,25 +427,32 @@ impl<'a, Cs> TxBuilder<'a, Cs> {
...
 
-        if let Some(WeightedUtxo {
-            utxo: Utxo::Local { .. },
-            ..
-        }) = self.params.utxos.get(&outpoint)
-        {
-            None
-        } else {
-            self.params.utxos.insert(
+        let mut collisioning_foreign_utxo_idx: Option<usize> = None;
```
ValuedMammal, line 430: Naming nit
```suggestion
        let mut existing_index: Option<usize> = None;
```

wallet/tests/wallet.rs
```diff
@@ -4699,3 +4699,73 @@ fn test_tx_details_method() {
...
+        &mut wallet,
+        txid1,
+        ConfirmationBlockTime {
+            block_id: BlockId {
+                height: 2_100,
+                hash: BlockHash::all_zeros(),
+            },
+            confirmation_time: 300,
+        },
+    );
```
ValuedMammal, line 4737: For this to make sense you should also insert a checkpoint into the local chain, but for this test I think `insert_tx` on its own is sufficient.

ValuedMammal (changes requested):
For consistency with the rest of the codebase I want to stick with the "UTXO" style acronym.

ValuedMammal (approved):
ACK d6d204bfe15b0ac643f9abc0b8edb74cda391a04

wallet/src/wallet/mod.rs
```diff
@@ -2117,8 +2115,9 @@ impl Wallet {
                         .then(|| new_local_utxo(k, i, full_txo))
                 })
                 // only process UTxOs not selected manually, they will be considered later in the
-                // chain NOTE: this avoid UTxOs in both required and optional list
-                .filter(|may_spend| !params.utxos.contains_key(&may_spend.outpoint))
+                // chain
+                // NOTE: this avoid UTxOs in both required and optional list
```
oleonardolima, line 2119:
``` suggestion
                // chain
                // NOTE: this avoid UTXOs in both required and optional list
```

oleonardolima (commented):
utACK

I left a question and a single nit, but overall it looks good to me! Agree with the new version of test suggested by VM's, it looks much better.

wallet/src/wallet/tx_builder.rs
```diff
@@ -407,25 +421,32 @@ impl<'a, Cs> TxBuilder<'a, Cs> {
...
+                        existing_index = Some(idx);
+                        break;
+                    }
+                }
+            }
+        }
+
+        if let Some(idx) = existing_index {
+            self.params.utxos.remove(idx);
+        }
```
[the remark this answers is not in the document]
nymius: The idea is to only keep the latest item on conflict foreign or local). Locals can override foreign and local in case of conflict, but foreign can only override foreign. That's what this line is doing, removing the old conflicting foreign utxo.

The documentation is addressing this behavior now:
> /// Foreign UTXOs are not prioritized over local UTXOs. If a local UTXO is added to the
> /// manually selected list, it will replace any conflicting foreign UTXOs. However, a foreign
> /// UTXO cannot replace a conflicting local UTXO.

oleonardolima (approved):
ACK 0522114aa17c788d952ac60db8b26cb08ad1d871

PR #265, opened 2025-06-11
Prefer Utxo::Local over Utxo::Foreign in OldestFirstCoinSelection

### LINKED ISSUE #264: `OldestFirstCoinSelection` selects `Utxo::Foreign`s first instead of last
The `OldestFirstCoinSelection` algorithm actually selects foreign UTXOs *first*, contra the comment, as `Option`'s `Ord` impl defines `None` as "less than" `Some`. This bug is on the master branch: https://github.com/bitcoindevkit/bdk_wallet/blob/master/wallet/src/wallet/coin_selection.rs

```rust
/// OldestFirstCoinSelection always picks the utxo with the smallest blockheight to add to the
/// selected coins next
///
/// This coin selection algorithm sorts the available UTXOs by blockheight and then picks them
/// starting from the oldest ones until the required amount is reached.
#[derive(Debug, Default, Clone, Copy)]
pub struct OldestFirstCoinSelection;

impl CoinSelectionAlgorithm for OldestFirstCoinSelection {
    fn coin_select<R: RngCore>(
        &self,
        required_utxos: Vec<WeightedUtxo>,
        mut optional_utxos: Vec<WeightedUtxo>,
        fee_rate: FeeRate,
        target_amount: Amount,
        drain_script: &Script,
        _: &mut R,
    ) -> Result<CoinSelectionResult, InsufficientFunds> {
        // We put the "required UTXOs" first and make sure the optional UTXOs are sorted from
        // oldest to newest according to blocktime
        // For utxo that doesn't exist in DB, they will have lowest priority to be selected
        let utxos = {
            optional_utxos.sort_unstable_by_key(|wu| match &wu.utxo {
                Utxo::Local(local) => Some(local.chain_position),
                Utxo::Foreign { .. } => None,
            });

            required_utxos
                .into_iter()
                .map(|utxo| (true, utxo))
                .chain(optional_utxos.into_iter().map(|utxo| (false, utxo)))
        };

        select_sorted_utxos(utxos, fee_rate, target_amount, drain_script)
    }
}

...

fn select_sorted_utxos(
    utxos: impl Iterator<Item = (bool, WeightedUtxo)>,
    fee_rate: FeeRate,
    target_amount: Amount,
    drain_script: &Script,
) -> Result<CoinSelectionResult, InsufficientFunds> {
    let mut selected_amount = Amount::ZERO;
    let mut fee_amount = Amount::ZERO;
    let selected = utxos
        .scan(
            (&mut selected_amount, &mut fee_amount),
            |(selected_amount, fee_amount), (must_use, weighted_utxo)| {
                if must_use || **selected_amount < target_amount + **fee_amount {
                    **fee_amount += fee_rate
                        * TxIn::default()
                            .segwit_weight()
                            .checked_add(weighted_utxo.satisfaction_weight)
                            .expect("`Weight` addition should not cause an integer overflow");
                    **selected_amount += weighted_utxo.utxo.txout().value;
                    Some(weighted_utxo.utxo)
                } else {
                    None
                }
            },
        )
        .collect::<Vec<_>>();

    let amount_needed_with_fees = target_amount + fee_amount;
    if selected_amount < amount_needed_with_fees {
        return Err(InsufficientFunds {
            needed: amount_needed_with_fees,
            available: selected_amount,
        });
    }

    let remaining_amount = selected_amount - amount_needed_with_fees;

    let excess = decide_change(remaining_amount, fee_rate, drain_script);

    Ok(CoinSelectionResult {
        selected,
        fee_amount,
        excess,
    })
}
```

nymius:
### Description

The comments in the `OldestFirstCoinSelection` implementation stated the following:
> // For utxo that doesn't exist in DB, they will have lowest priority to be selected

But this was not honored in the code.

This PR enforces this and ensures the expected behaviour through the two following new tests:
- `test_oldest_first_coin_selection_uses_all_optional_with_foreign_utxo_locals_sorted_first`
- `test_oldest_first_coin_selection_uses_only_all_optional_local_utxos_not_a_single_foreign`

Fixes #264 

### Changelog notice

No public APIs are changed by these commits.

### Checklists

> [!IMPORTANT]
> This pull request **DOES NOT** break the existing API
* [x] I've added tests for the new code
* [x] I've expanded docs addressing new code

### DIFF
--- a/wallet/src/wallet/coin_selection.rs
+++ b/wallet/src/wallet/coin_selection.rs
@@ -275,11 +275,12 @@ impl CoinSelectionAlgorithm for OldestFirstCoinSelection {
     ) -> Result<CoinSelectionResult, InsufficientFunds> {
         // We put the "required UTXOs" first and make sure the optional UTXOs are sorted from
         // oldest to newest according to blocktime
-        // For utxo that doesn't exist in DB, they will have lowest priority to be selected
+        // For UTXOs that doesn't exist in DB (Utxo::Foreign), they will have lowest priority to be
+        // selected
         let utxos = {
             optional_utxos.sort_unstable_by_key(|wu| match &wu.utxo {
-                Utxo::Local(local) => Some(local.chain_position),
-                Utxo::Foreign { .. } => None,
+                Utxo::Local(local) => (false, Some(local.chain_position)),
+                Utxo::Foreign { .. } => (true, None),
             });
 
             required_utxos
@@ -726,10 +727,11 @@ fn calculate_cs_result(
 mod test {
     use assert_matches::assert_matches;
     use bitcoin::hashes::Hash;
-    use bitcoin::OutPoint;
+    use bitcoin::{psbt, OutPoint, Sequence};
     use chain::{ChainPosition, ConfirmationBlockTime};
     use core::str::FromStr;
     use rand::rngs::StdRng;
+    use std::boxed::Box;
 
     use bitcoin::{Amount, BlockHash, ScriptBuf, TxIn, TxOut};
 
@@ -776,6 +778,30 @@ mod test {
                 transitively: None,
             },
         )
+    }
+
+    fn foreign_utxo(value: Amount, index: u32) -> WeightedUtxo {
+        assert!(index < 10);
+        let outpoint = OutPoint::from_str(&format!(
+            "000000000000000000000000000000000000000000000000000000000000000{}:0",
+            index
+        ))
+        .unwrap();
+        WeightedUtxo {
+            utxo: Utxo::Foreign {
+                outpoint,
+                sequence: Sequence(0xFFFFFFFD),
+                psbt_input: Box::new(psbt::Input {
+                    witness_utxo: Some(TxOut {
+                        value,
+                        script_pubkey: ScriptBuf::new(),
+                    }),
+                    non_witness_utxo: None,
+                    ..Default::default()
+                }),
+            },
+            satisfaction_weight: Weight::from_wu_usize(P2WPKH_SATISFACTION_SIZE),
+        }
     }
 
     fn utxo(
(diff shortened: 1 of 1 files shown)


MaxFangX:
Thanks for the quick fix!

wallet/src/wallet/coin_selection.rs
```diff
@@ -273,19 +273,27 @@ impl CoinSelectionAlgorithm for OldestFirstCoinSelection {
...
             optional_utxos.sort_unstable_by_key(|wu| match &wu.utxo {
                 Utxo::Local(local) => Some(local.chain_position),
                 Utxo::Foreign { .. } => None,
             });
 
+            let (local_utxos, foreign_utxos): (Vec<WeightedUtxo>, Vec<WeightedUtxo>) =
+                optional_utxos.into_iter().partition(|wu| match &wu.utxo {
+                    Utxo::Local(..) => true,
+                    Utxo::Foreign { .. } => false,
+                });
```
evanlinjin, line 290: Why do you think about using `.sort_unstable_by` or just using `.sort_unstable_by_key` with a different key?

I think it's better to sort everything in one loop than in two goes.

```rust
            optional_utxos.sort_unstable_by_key(|wu| {
                Reverse(match &wu.utxo {
                    Utxo::Local(local) => Some(Reverse(local.chain_position)),
                    Utxo::Foreign { .. } => None,
                })
            });
```

nymius: I like it, however, by reading the `Reverse` docs I arrived to the following solution:
```rust
optional_utxos.sort_unstable_by_key(|wu| {
    match &wu.utxo {
        Utxo::Local(local) => (false, Some(local.chain_position)),
        Utxo::Foreign { .. } => (true, None),
    }
});
```

`false` < `true`, so all `LocalUtxos` are ordered at the beginning of the `Vec`. WDYT?

ValuedMammal (approved):
ACK 345cc6ea553a2db74742fa9e673985004d852fb9

oleonardolima (approved):
ACK 345cc6ea553a2db74742fa9e673985004d852fb9

PR #268, opened 2025-06-14
feat: add `justfile`

### LINKED ISSUE #240: Create justfile
`just` is the de facto standard for command runner utilities. I think it would be useful to have one here.

luisschwab:
### Description

Closes #240.

This PR adds a `justfile`, updates the PR template to use it, and adds a section on the `README.md` to show available recipes.

These are the implemented recipes:
```justfile
alias b := build
alias c := check
alias f := fmt
alias t := test
alias p := pre-push

build:
    cargo build

check:
    cargo +nightly fmt --all -- --check
    cargo check --workspace --exclude 'example_*' --all-features
    cargo clippy --all-features --all-targets -- -D warnings
    @[ "$(git log --pretty='format:%G?' -1 HEAD)" = "N" ] && \
        echo "\n⚠️  Unsigned commit: BDK requires that commits be signed." || \
        true

fmt:
    cargo +nightly fmt

test:
    cargo test --workspace --exclude 'example_*' --all-features

pre-push: fmt check test
```

`check` will verify if `HEAD` was signed and echo that warning if not. It does not check all commits, but I think that checking only the last is a pretty good heuristic (who only signs the last commit?).

Before pushing, one only needs to run `just p`.

#### All Submissions:

* [X] I've signed all my commits
* [X] I followed the [contribution guidelines](https://github.com/bitcoindevkit/bdk/blob/master/CONTRIBUTING.md)
* [X] I ran `just p` before pushing

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,7 +36,7 @@ It is built upon the excellent [`rust-bitcoin`] and [`rust-miniscript`] crates.
 There is currently only one published crate in this repository:
 
 - [`wallet`](./wallet): Contains the central high level `Wallet` type that is built from the low-level mechanisms provided by the other components.
-  
+
 Crates that `bdk_wallet` depends on are found in the [`bdk`] repository.
 
 Fully working examples of how to use these components are in `/examples`:
@@ -54,6 +54,12 @@ Fully working examples of how to use these components are in `/examples`:
 The libraries in this repository maintain a MSRV of 1.63.0.
 
 To build with the MSRV of 1.63.0 you will need to pin dependencies by running the [`pin-msrv.sh`](./ci/pin-msrv.sh) script.
+
+## Just
+
+This project has a [`justfile`](/justfile) for easy command running. You must have [`just`](https://github.com/casey/just) installed.
+
+To see a list of available recipes: `just -l`
 
 ## License
 
--- a/justfile
+++ b/justfile
@@ -1,0 +1,29 @@
+alias b := build
+alias c := check
+alias f := fmt
+alias t := test
+alias p := pre-push
+
+# Build the project
+build:
+   cargo build
+
+# Check code: formatting, compilation, linting, and commit signature
+check:
+   cargo +nightly fmt --all -- --check
+   cargo check --workspace --exclude 'example_*' --all-features
+   cargo clippy --all-features --all-targets -- -D warnings
+   @[ "$(git log --pretty='format:%G?' -1 HEAD)" = "N" ] && \
+       echo "\n⚠️  Unsigned commit: BDK requires that commits be signed." || \
+       true
+
+# Format all code
+fmt:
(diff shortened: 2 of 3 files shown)


ValuedMammal:
`fmt` should come before `check`.

README.md
```diff
@@ -55,6 +55,20 @@ The libraries in this repository maintain a MSRV of 1.63.0.
...
+To build: `just b`
+
+To check: `just c`
+
+To format: `just f`
+
+To test: `just t`
+
+To run the pre-push suite (check, format and test): `just p`
+
```
notmandatory, line 71: Instead of these docs how about:

```suggestion
To see a list of available recipes: `just -l`
```
And then add a docs line to your `Justfile` above each recipe with a little about what it does.  This would be easier to maintain since we won't need to update this README file when adding to or changing the `Justfile`.

justfile
```diff
@@ -0,0 +1,24 @@
+alias b := build
+alias c := check
+alias f := fmt
+alias t := test
+alias p := pre-push
+
+build:
+    cargo build
+
+check:
```
notmandatory, line 10:
```suggestion
# check for format, build, lint, and commit errors
check:
```

justfile
```diff
@@ -0,0 +1,24 @@
+alias b := build
+alias c := check
+alias f := fmt
+alias t := test
+alias p := pre-push
+
+build:
```
notmandatory, line 7:
```suggestion
# build the project
build:
```

justfile
```diff
@@ -0,0 +1,24 @@
...
+
+check:
+    cargo +nightly fmt --all -- --check
+    cargo check --workspace --exclude 'example_*' --all-features
+    cargo clippy --all-features --all-targets -- -D warnings
+    @[ "$(git log --pretty='format:%G?' -1 HEAD)" = "N" ] && \
+        echo "\n⚠️  Unsigned commit: BDK requires that commits be signed." || \
+        true
+
+fmt:
```
notmandatory, line 18:
```suggestion
# format the code
fmt:
```

justfile
```diff
@@ -0,0 +1,24 @@
...
+    cargo +nightly fmt
+
+test:
```
notmandatory, line 21:
```suggestion
# run all tests
test:
```

justfile
```diff
@@ -0,0 +1,24 @@
...
+    cargo test --workspace --exclude 'example_*' --all-features
+
+pre-push: fmt check test
```
notmandatory, line 24:
```suggestion
# run the pre-push suite (check, format and test)
pre-push: fmt check test
```

notmandatory (commented):
Great start. I have some suggestions to make the docs easier to maintain.

README.md
luisschwab: Agreed

PR #274, opened 2025-06-30
docs(wallet): add `sync` operation to `bdk_wallet` examples 

### LINKED ISSUE #253: Missing `bdk_wallet` examples with `sync`
We only have `full_scan` examples for Electrum/Esplora. We should also showcase how to `sync` in those examples.

We also need to ensure that we can detect evicted txs (I forgot if we are doing this).

tvpeter:
### Description

This PR adds `sync` operation to `bdk_wallet` examples for `electrum` and `esplora` clients.

Fixes #253 

#### All Submissions:

### DIFF
--- a/examples/example_wallet_electrum/src/main.rs
+++ b/examples/example_wallet_electrum/src/main.rs
@@ -1,31 +1,29 @@
-use bdk_wallet::bitcoin::FeeRate;
-use bdk_wallet::file_store::Store;
-use bdk_wallet::psbt::PsbtUtils;
-use bdk_wallet::Wallet;
-use std::io::Write;
-
 use bdk_electrum::electrum_client;
 use bdk_electrum::BdkElectrumClient;
 use bdk_wallet::bitcoin::Amount;
+use bdk_wallet::bitcoin::FeeRate;
 use bdk_wallet::bitcoin::Network;
 use bdk_wallet::chain::collections::HashSet;
+use bdk_wallet::psbt::PsbtUtils;
+use bdk_wallet::rusqlite::Connection;
+use bdk_wallet::Wallet;
 use bdk_wallet::{KeychainKind, SignOptions};
+use std::io::Write;
+use std::thread::sleep;
+use std::time::Duration;
 
-const DB_MAGIC: &str = "bdk_wallet_electrum_example";
 const SEND_AMOUNT: Amount = Amount::from_sat(5000);
 const STOP_GAP: usize = 50;
 const BATCH_SIZE: usize = 5;
 
-const NETWORK: Network = Network::Testnet;
+const DB_PATH: &str = "bdk-example-electrum.sqlite";
+const NETWORK: Network = Network::Testnet4;
 const EXTERNAL_DESC: &str = "wpkh(tprv8ZgxMBicQKsPdy6LMhUtFHAgpocR8GC6QmwMSFpZs7h6Eziw3SpThFfczTDh5rW2krkqffa11UpX3XkeTTB2FvzZKWXqPY54Y6Rq4AQ5R8L/84'/1'/0'/0/*)";
 const INTERNAL_DESC: &str = "wpkh(tprv8ZgxMBicQKsPdy6LMhUtFHAgpocR8GC6QmwMSFpZs7h6Eziw3SpThFfczTDh5rW2krkqffa11UpX3XkeTTB2FvzZKWXqPY54Y6Rq4AQ5R8L/84'/1'/0'/1/*)";
-const ELECTRUM_URL: &str = "ssl://electrum.blockstream.info:60002";
+const ELECTRUM_URL: &str = "ssl://mempool.space:40002";
 
 fn main() -> Result<(), anyhow::Error> {
-    let db_path = "bdk-electrum-example.db";
-
-    let (mut db, _) = Store::<bdk_wallet::ChangeSet>::load_or_create(DB_MAGIC.as_bytes(), db_path)?;
-
+    let mut db = Connection::open(DB_PATH)?;
     let wallet_opt = Wallet::load()
         .descriptor(KeychainKind::External, Some(EXTERNAL_DESC))
         .descriptor(KeychainKind::Internal, Some(INTERNAL_DESC))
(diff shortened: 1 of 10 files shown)


examples/example_wallet_electrum/src/main.rs
```diff
@@ -92,5 +97,29 @@ fn main() -> Result<(), anyhow::Error> {
...
+            .inspect(|sync_item, sync_progress| {
+                let progress_percent =
+                    (100 * sync_progress.consumed()) as f32 / sync_progress.total() as f32;
+                eprintln!("[ SCANNING {:03.0}%] {}", progress_percent, sync_item)
+            });
+
+    client.populate_tx_cache(wallet.tx_graph().full_txs().map(|tx_node| tx_node.tx));
+
+    let sync_update = client.sync(sync_request, BATCH_SIZE, false)?;
+    wallet.apply_update(sync_update)?;
```
oleonardolima, line 114: I think it's a good idea to also persist here. wdyt ?

examples/example_wallet_electrum/src/main.rs
```diff
@@ -92,5 +97,29 @@ fn main() -> Result<(), anyhow::Error> {
...
 
+    println!("=== Performing Partial Sync ===");
+    std::io::stdout().flush().expect("must flush");
+    let sync_request =
+        wallet
+            .start_sync_with_revealed_spks()
+            .inspect(|sync_item, sync_progress| {
+                let progress_percent =
+                    (100 * sync_progress.consumed()) as f32 / sync_progress.total() as f32;
+                eprintln!("[ SCANNING {:03.0}%] {}", progress_percent, sync_item)
```
oleonardolima, line 108: Good! I personally think this approach is good, but I'm wondering if it's too verbose.

examples/example_wallet_electrum/src/main.rs
```diff
@@ -92,5 +97,29 @@ fn main() -> Result<(), anyhow::Error> {
     client.transaction_broadcast(&tx)?;
     println!("Tx broadcasted! Txid: {}", tx.compute_txid());
 
+    println!("=== Performing Partial Sync ===");
+    std::io::stdout().flush().expect("must flush");
```
oleonardolima, line 101: Is this strictly needed?

oleonardolima (commented):
cACK 8e1a6ef2f74e66a0e97fcdd643adb5f4e73d72d3

I left a few comments that I think are applicable. Also, I was unable to test the electrum example locally (I guess it's a local problem on my end).

I think it's a good opportunity to also add a forced eviction, and show that to the user. WDYT?

Any reason to not add the sync usage to blocking esplora example ?

examples/example_wallet_electrum/src/main.rs
tvpeter: Oh yes, I will do that. Thank you!

examples/example_wallet_electrum/src/main.rs
tvpeter: Okay

examples/example_wallet_electrum/src/main.rs
tvpeter: I do think so because the progress update may appear before the `performing partial sync` message if it is not flushed (when using `eprintln`). I will update this, so it is consistent.

notmandatory:
I pushed a new commit to fix a couple things I found while testing:

1. for electrum, esplora examples they all now use sqlite+testnet and print mempool.space URLs for created txids
2. for esplore examples I get the fee rate estimate from explora rather than using the default 1 sat/vb  
3. for the RPC example I added a justfile to help testing with regtest bitcoind

notmandatory (approved):
ACK ce0a17e547c66a84dc9ff570c3dbee8b0ac75659

examples/example_wallet_esplora_async/src/main.rs
```diff
@@ -114,30 +108,63 @@ async fn main() -> Result<(), anyhow::Error> {
...
+        .tx_graph()
+        .full_txs()
+        .find(|full_tx| full_tx.txid == txid)
+        .map_or(0, |full_tx| full_tx.last_seen.unwrap_or(0));
+    if !evicted_txs
+        .iter()
+        .any(|(evicted_txid, _)| evicted_txid == &txid)
+    {
+        evicted_txs.push((txid, last_seen));
     }
```
oleonardolima: Any reason to not use the approach from electrum example ?
```
    for (txid, last_seen) in &sync_update.tx_update.evicted_ats {
        evicted_txs.push((*txid, *last_seen));
    }
```

oleonardolima:
I'm not sure why, but now that it's using testnet instead of signet, I keep getting timeout errors and the RBF fails because the transaction is already mined in within that timeframe.

notmandatory:
> I'm not sure why, but now that it's using testnet instead of signet, I keep getting timeout errors and the RBF fails because the transaction is already mined in within that timeframe.

I switched it to use mempool.space and testnet4 which seem to be more reliable. It's also easier to get testnet4 coins from mempool.space since all you need to do is login with your GitHub account. I also cleaned up the just file for RPC to use the `stop` command, good suggestion @oleonardolima.

tvpeter:
@ValuedMammal following your comments on the issue [here](https://github.com/bitcoindevkit/bdk_wallet/issues/253#issuecomment-3146872241), I have removed the step that call the `Wallet::apply_evicted_txs` to evict txs.

Thank you.

oleonardolima (commented):
partial tACK bb4d79bfac0cb32a70ce121faf9a2a7a372762fb

I've test for all examples, both esplora and electrum works just fine.

I found a problem in the wallet_rpc one, I've left it as a comment.

examples/example_wallet_rpc/src/main.rs
```diff
@@ -91,8 +91,7 @@ fn main() -> anyhow::Result<()> {
     );
 
     let start_load_wallet = Instant::now();
-    let (mut db, _) =
-        Store::<bdk_wallet::ChangeSet>::load_or_create(DB_MAGIC.as_bytes(), args.db_path)?;
+    let mut db = Connection::open(DB_PATH)?;
```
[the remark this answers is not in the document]
notmandatory: Oops sorry my mistake. Fixed it.

oleonardolima (approved):
tACK 11bfc2f649432ba795ce61a16f85ece7c0aab490

PR #275, opened 2025-07-02
Multipath descriptor support (BIP 389)

### LINKED ISSUE #11: Support multipath descriptors
**Describe the enhancement**  

The release of [`rust-miniscript` 10.0.0](https://github.com/rust-bitcoin/rust-miniscript/blob/master/CHANGELOG.md#1000---may-24-2023) has [multipath descriptor](https://github.com/rust-bitcoin/rust-miniscript/pull/470) support. We should support this also in bdk 1.0 and backported to a bdk 0.2X (maintenance) release.

**Use case**  

See bitcoin/bitcoin#17190

**Additional context**

This was requested by @andreasgriffin for his project [bitcoin-safe](https://github.com/andreasgriffin/bitcoin-safe).

[BIP-389](https://github.com/bitcoin/bips/blob/master/bip-0389.mediawiki)

schjonhaug:
#### Key Features:
  - New API: `Wallet::create_multipath(descriptor)` following the same pattern as `create()` and `create_single()`
  - [BIP 389](https://github.com/bitcoin/bips/blob/master/bip-0389.mediawiki) compliance with exactly 2-path validation (receive and change)
  - Robust validation with clear error messages
  - Pattern consistency with existing wallet creation methods

 #### Usage Example:
  
```rust
let multipath_desc = "wpkh([9a6a2580/84'/1'/0']tpub.../‹0;1›/*)";
let wallet = Wallet::create_multipath(multipath_desc)
      .network(Network::Testnet)
      .create_wallet_no_persist()?;

// Automatically creates separate receive and change descriptors
let receive_addr = wallet.peek_address(KeychainKind::External, 0);  // Uses path /0/*
let change_addr = wallet.peek_address(KeychainKind::Internal, 0);   // Uses path /1/*
```

#### Design Decisions:
  1. Pattern Consistency: Uses `make_multipath_descriptor_to_extract()` helper following the same pattern as existing `make_descriptor_to_extract()` function
  2. Lazy Evaluation: Descriptor parsing only happens when needed during wallet creation, not during parameter setup - this maintains performance and follows Rust's lazy evaluation patterns
  3. Strict Validation: Only allows exactly 2-path multipath descriptors to ensure proper receive/change separation
  4. API Consistency: The `create_multipath()` method returns `CreateParams` just like `create()` and `create_single()`, maintaining the fluent builder pattern

#### Implementation Notes:
  - The function calls `make_multipath_descriptor_to_extract()` twice (once for receive, once for change) which is intentional and follows the established pattern of lazy evaluation
  - Validation occurs in the descriptor extraction closures, providing clear error messages when descriptors are actually processed
  - All existing functionality remains unchanged - this is a purely additive feature

### Changelog notice

Added:
  - `Wallet::create_multipath()` method for creating wallets from BIP 389 multipath descriptors
  - `CreateParams::new_multipath()` for multipath descriptor parameter creation
  - Support for 2-path multipath descriptors with automatic receive/change separation
  - Enhanced validation for multipath descriptors with descriptive error messages
  
#### All Submissions:

#### New Features:

#### Bugfixes:

Closes #11

### DIFF
--- a/wallet/src/descriptor/error.rs
+++ b/wallet/src/descriptor/error.rs
@@ -21,7 +21,8 @@ pub enum Error {
     InvalidDescriptorChecksum,
     /// The descriptor contains hardened derivation steps on public extended keys
     HardenedDerivationXpub,
-    /// The descriptor contains multipath keys
+    /// The descriptor contains multipath keys with an invalid number of paths (must have exactly 2
+    /// paths for receive and change)
     MultiPath,
     /// Error thrown while working with [`keys`](crate::keys)
     Key(crate::keys::KeyError),
@@ -68,7 +69,7 @@ impl fmt::Display for Error {
             ),
             Self::MultiPath => write!(
                 f,
-                "The descriptor contains multipath keys, which are not supported yet"
+                "The descriptor contains multipath keys with invalid number of paths (must have exactly 2 paths for receive and change)"
             ),
             Self::Key(err) => write!(f, "Key error: {}", err),
             Self::Policy(err) => write!(f, "Policy error: {}", err),
--- a/wallet/src/descriptor/mod.rs
+++ b/wallet/src/descriptor/mod.rs
@@ -14,6 +14,7 @@
 //! This module contains generic utilities to work with descriptors, plus some re-exported types
 //! from [`miniscript`].
 
+use crate::alloc::string::ToString;
 use crate::collections::BTreeMap;
 use alloc::string::String;
 use alloc::vec::Vec;
@@ -312,7 +313,11 @@ pub(crate) fn check_wallet_descriptor(
     }
 
     if descriptor.is_multipath() {
-        return Err(DescriptorError::MultiPath);
+        return Err(DescriptorError::Miniscript(
+            miniscript::Error::BadDescriptor(
+                "`check_wallet_descriptor` must not contain multipath keys".to_string(),
+            ),
+        ));
     }
 
     // Run miniscript's sanity check, which will look for duplicated keys and other potential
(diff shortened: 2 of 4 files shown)


ValuedMammal:
Thanks for working on this @schjonhaug.

wallet/src/descriptor/error.rs
```diff
@@ -68,7 +68,7 @@ impl fmt::Display for Error {
             ),
             Self::MultiPath => write!(
                 f,
-                "The descriptor contains multipath keys, which are not supported yet"
+                "The descriptor contains multipath keys with invalid number of paths (must have exactly 2 paths for receive and change)"
```
notmandatory, line 71: Please also update the enum docs on line 24, should have more or less the same message as you have here.

wallet/src/descriptor/mod.rs
```diff
@@ -961,4 +978,36 @@ mod test {
 
         Ok(())
     }
+
+    #[test]
+    fn test_multipath_descriptor_validation() {
```
notmandatory, line 983: Doesn't this test duplicate the test logic in `test_check_wallet_descriptor()` ? If so it should be removed.

notmandatory (commented):
Thanks for taking this one on, overall your PR looks great. I especially like the detailed docs and new validation to ensure only two indexes are used. 

I have two small suggestions and you should do a rebase and then I think it's ready to go merge.

wallet/src/descriptor/mod.rs
```diff
@@ -312,7 +312,14 @@ pub(crate) fn check_wallet_descriptor(
...
     if descriptor.is_multipath() {
-        return Err(DescriptorError::MultiPath);
+        // Only allow multipath descriptors with exactly 2 paths (receive and change)
+        let paths = descriptor
+            .clone()
+            .into_single_descriptors()
+            .map_err(|_| DescriptorError::MultiPath)?;
+        if paths.len() != 2 {
+            return Err(DescriptorError::MultiPath);
+        }
```
ValuedMammal, line 322: I don't know what others think, but IMO `check_wallet_descriptor` should still reject a multipath descriptor, because we expect to be able to insert the checked descriptor directly into the Indexer, which doesn't do any extra validation.

```suggestion
         return Err(DescriptorError::Miniscript(
            miniscript::Error::BadDescriptor(
                "`check_wallet_descriptor` must not contain multipath keys".to_string(),
            ),
        ));
```

notmandatory: If the point here is that `check_wallet_descriptor` should only be used to verify that the descriptors are valid for use in the indexer then should @schjonhaug put this multipath verification directly in the `create_multipath` function? then call `check_wallet_descriptor` there too on the split out non-multipath version of the descriptors?

notmandatory:
>... My suggestion would be to use something like `create_from_two_path_descriptor` or similar to distinguish between `N` possible keychains and two.

@rustaceanrob what do you think about a little shorter name like `Wallet::create_two_path(two_path_descriptor)`?

notmandatory:
@schjonhaug do you think you'll have time this week to make the small changes ValuedMammal and Rob suggested above? If so then I think we can get this merged for our 2.1 release next week. Thanks!

thunderbiscuit:
After playing with this today to expose it in the bindings I realized that this constructor cannot be used with private keys. This is a fairly important aspect of using the method, and I think it should be made more explicit in the API docs. I will open a PR for this tomorrow.

FYI if my investigation is correct, the issue is that miniscript cannot parse an xprv (I saw a comment from Sanket somewhere but now I can't find it anymore!) if the multipath contains hardened derivation paths, and therefore miniscipt cannot ensure the xprv will be parsable into public keys in all cases? Something like that. But our arguments for bdk wallet constructors require the `IntoWalletDescriptor` trait:

```rust
pub trait IntoWalletDescriptor {
    /// Convert to wallet descriptor
    fn into_wallet_descriptor(
        self,
        secp: &SecpCtx,
        network: Network,
    ) -> Result<(ExtendedDescriptor, KeyMap), DescriptorError>;
}

pub type ExtendedDescriptor = Descriptor<DescriptorPublicKey>;
```

In other words, upon creation, the wallet will always try to parse the descriptor string into its public form, and here since it cannot be done by miniscript we fail to build.
```

PR #287, opened 2025-07-31
clippy: fix uninlined format args

### LINKED ISSUE #281: [clippy] Fix `uninlined_format_args`
This is needed in order to

- [ ] Update the `rust-version` of the repo to 1.88.0. We can either allow it everywhere or else try to find and fix every instance. Edit: Can probably be done with a simple `cargo clippy --fix`.

ValuedMammal:
Fix #281 by taking clippy's suggestion to inline the arguments to a `format!` string where applicable.

Also update `rust-version` to 1.88.0, closing #280.

### Notes to the reviewers

I've gone ahead and fixed the clippy warnings, although some have argued that the lint is quite pedantic and not entirely helpful when it comes to style and readability. Therefore I also allowed the lint to prevent clippy from warning about it in the future d75126e. That change can be reverted if/when rust lang moves `uninlined_format_args` back to the pedantic, i.e. not default group of lints https://github.com/rust-lang/rust-clippy/pull/15287.

#### All Submissions:

* [ ] ~~This pull request breaks the existing API~~

### DIFF
--- a/wallet/src/lib.rs
+++ b/wallet/src/lib.rs
@@ -7,6 +7,7 @@
 )]
 #![no_std]
 #![warn(missing_docs)]
+#![allow(clippy::uninlined_format_args)]
 
 #[cfg(feature = "std")]
 #[macro_use]


wallet/src/lib.rs
```diff
@@ -7,6 +7,7 @@
 )]
 #![no_std]
 #![warn(missing_docs)]
+#![allow(clippy::uninlined_format_args)]
```
notmandatory, line 10: I wouldn't mind leaving this lint enabled even if it's a bit pedantic. But no strong feelings about it either way so OK to allow for now.

notmandatory (approved):
ACK d75126ef07d68e99cc0cdbbf6328df88b97e9955

You'll need to update the `master` branch rules settings to version 1.88.0 before merging.

oleonardolima (approved):
ACK d75126ef07d68e99cc0cdbbf6328df88b97e9955

PR #294, opened 2025-08-06
chore: bump dev to `2.2.0-alpha.0`

oleonardolima:
#### All Submissions:

#### New Features:

#### Bugfixes:

### DIFF
--- a/wallet/Cargo.toml
+++ b/wallet/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "bdk_wallet"
 homepage = "https://bitcoindevkit.org"
-version = "2.1.0"
+version = "2.2.0-alpha.0"
 repository = "https://github.com/bitcoindevkit/bdk_wallet"
 documentation = "https://docs.rs/bdk_wallet"
 description = "A modern, lightweight, descriptor-based wallet library"


thunderbiscuit:
I would also suggest using the `3.0.0-alpha.0` version instead, letting users know where we are thinking of going next.

thunderbiscuit:
Yeah you absolutely can (just push fixes to the latest feature release of a major release), but I think in general having branches for each allows you to do it if you wanted to. It also makes it explicit that patches only go on the latest feature release if that's the only branch that gets a bump (say `release/2.2`) and all other don't get the patch (`release/2.1` and `release/2.0` don't get commits added). The other thing it does is simplify digging into code when people say they use "X" release. I always have a branch with all commits for that release ready. Checking out the tag does the same really, but branches are nice to work with.

Of course this is just my 2 sats. On bdk-ffi you can find all feature releases get their own branches. But honestly it doesn't impact the workflow immensely, I just figured I'd share/point it out.

ValuedMammal:
I'm fine with renaming the branch to `release/2.2`, in fact just calling it `release/2.x` would make the most sense to me.

> On bdk-ffi you can find all feature releases get their own branches

Yes but releases below 1.0 are semver _incompatible_ and can be treated as the major version for all intents and purposes.

ValuedMammal:
Ok I see you have a `release/1.1` and `release/1.2` on bdk-ffi. Like you said devs can also check out a tag. Still, in this instance I'm not sure it would makes sense to backport a fix to both 2.1 and 2.2 branches.

ValuedMammal:
ACK 8b97a08aeae85fe2189e5be9b63fdd25b762780f

Done in e43e59752d0cfe4f474097ba68080afa84fb8cc9

thunderbiscuit:
Yep agree with all your points above. It's mostly cosmetic, and in reality you're unlikely to port patches back. I like our workflow, but don't take my comment + approach as the only one!

Issue #66, opened 2024-10-18
Consider not enabling RBF by default


thunderbiscuit:
I know I know 🤣🤣 we just merged this.

But our timing for this PR (the one that made RBF signalling by default) was in a way quite late to the game. Bitcoin Core 0.28 now ships with full transaction replacement regardless of the RBF flag. This on its own doesn't really change anything for us, but after speaking with Murch on the Optech live podcast, his take on the RBF signalling is that in time, signalling _for_ RBF will actually fingerprint your wallet (because only few libraries will do that by default).

This is not really a pressing issue, but it's good to get the conversation going early. Over time if all transactions can always be RBF-ed and the ratio of RBF-signalling transaction keeps going down, it would probably make sense to revert the default to _not_ signalling for RBF.

BTW does anyone know if there is an easy place to find data on this? Like the ratio of transactions that signal for RBF over a given period of time? I wonder what's the proportion of tx in 2024 that do signal for it.

thunderbiscuit:
Adding to this discussion by linking to a bitcoin mailing list thread: https://gnusha.org/pi/bitcoindev/178162039498.7.16213177200751410070.1421232845@slmail.me/T/#t

PR #297, opened 2025-08-12
Implement `create_psbt` for Wallet

### LINKED ISSUE #164: Redesign `TxBuilder`
### Description

We can design a more intuitive transaction-building flow by splitting tx-building into the follow stages:

1. Coin Control
    - Determine which coins can be candidates, and which coins to filter out.
    - Determine coin groups (which coins must be spent together, and which coins must NOT be spent together).
    - Add plan to each coin (from planning module).
2. Coin Selection
    - Select coins for the actual transaction.
    - Determine whether we need a drain output (change policy).
3. Transaction Finalization
    - Formulate unsigned transaction.
    - Reordering/shuffling of inputs/outputs.
4. PSBT Preparation
    - Maps the candidates chosen by coin selection to their plans and uses the plan to create the PSBT input.
    - Maps the outputs (including drain if exists) into the PSBT.
5. Signing

### Implementation Details

- For "must be spent together", group them somehow like: `Vec<(Plan, TxOut, OutPoint)>`.

### Related

- Meeting minutes: https://hackmd.io/@evanlinjin/bdk_minutes_20221005
- Old ticket: https://github.com/LLFourn/bdk_core_staging/issues/40

### LINKED ISSUE #204: Flexible RBF construction
**Describe the enhancement**  
In bdk 0.x   could construct PSBTs using spent utxos, which was essential to construct flexible RBF transaction.
In bdk 1.x only non-spent UTXOs are allow to be added as inputs. This creates the need for another method to allow creating RBF transaction. `BumpFeeTxBuilder` however is extremely limited as it doesn't allow adding/removing inputs or outputs. Both is required in doing anything except pure fee-bumping. 

**Use case**  

Use cases for more advanced RBF:
- sending the outputs back to yourself (if for some reason the original transaction is undesireable)
- dynamic transaction batching
- using different utxos (with at least 1 input conflict) to have enough funds for a higher fee, while most outputs remain constant
- there are probably a lot more for 2. layer protocols

**Additional context**
The necessity for more flexibility was discussed in https://github.com/bitcoindevkit/bdk/issues/1374  and lead to https://github.com/bitcoindevkit/bdk/pull/1799 , which was not merged. https://github.com/bitcoindevkit/bdk-tx might also be a way to address this except `TxBuilder`. 

Remarks (from my personal point of view):
- Constructing (complex) RBF transactions is in my point of view very important and wallets like Sparrow can do these RBF constructions.

ValuedMammal:
### Description

This PR introduces a new PSBT construction flow and associated wallet APIs.

Primary additions:

- New `PsbtParams<Ctx>` API with typed contexts (`CreateTx` and `ReplaceTx`) for transaction creation and replacement.
- New wallet entry points for PSBT construction and RBF:
	- `Wallet::create_psbt` and `Wallet::create_psbt_with_rng`
	- `Wallet::replace_by_fee`, `Wallet::replace_by_fee_with_rng`, and `Wallet::replace_by_fee_and_recipients`
- New error types for the new create-PSBT and RBF paths.
- Supporting wallet helpers for transaction insertion and canonicalization-aware transaction listing.
- New examples and expanded test coverage for expected PSBT/RBF behavior.
- Generic `TxOrdering<In, Out>` updates used by the new selection/PSBT path

This work is part of the broader migration of transaction construction to `bdk_tx`, which uses miniscript's `plan` module under the hood. A key long-term goal is the move away from legacy policy-based transaction construction internals (#4).

fix #204 

### Notes to the reviewers

Suggested review order for this PR:

1. Start with `examples/psbt.rs` and `examples/replace_by_fee.rs` for a high-level API walkthrough.
2. Review `src/psbt/params.rs` in detail (types, state transitions, params semantics, and so on).
3. Review the new wallet implementation in `src/wallet/mod.rs` (create-PSBT and RBF execution paths plus shared helpers).
4. Review tests in `tests/psbt.rs`, `tests/wallet.rs` to ensure expected behavior and edge cases are covered.

### Changed

- `TxOrdering::Custom` is generic over the input and output types.

### Added

- `psbt::params` module
- `PsbtParams` struct
- `CoinSelectionStrategy` enum
- `UtxoFilter` struct

Added `pub` methods on `Wallet`

- `Wallet::create_psbt{_with_rng}`
- `Wallet::replace_by_fee{_with_rng}`
- `Wallet::replace_by_fee_and_recipients`

Added examples
- `examples/psbt.rs`
- `examples/replace_by_fee.rs`

### Checklists

* [ ] ~~This pull request breaks the existing API~~

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -83,3 +83,9 @@ name = "esplora_blocking"
 
 [[example]]
 name = "bitcoind_rpc"
+
+[[example]]
+name = "psbt"
+
+[[example]]
+name = "replace_by_fee"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -79,3 +79,9 @@ name = "esplora_blocking"
 
 [[example]]
 name = "bitcoind_rpc"
+
+[[example]]
+name = "psbt"
+
+[[example]]
+name = "replace_by_fee"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -33,9 +33,15 @@ bdk_file_store = { version = "0.22.0", optional = true }
 bip39 = { version = "2.2.2", optional = true }
 tempfile = { version = "3.26.0", optional = true }
 
+[dependencies.bdk_tx]
+version = "0.2.0"
+default-features = false
+git = "https://github.com/valuedmammal/bdk-tx"
+branch = "release/0_2_0"
+
 [features]
 default = ["std"]
-std = ["bitcoin/std", "bitcoin/rand-std", "miniscript/std", "bdk_chain/std"]
+std = ["bitcoin/std", "bitcoin/rand-std", "miniscript/std", "bdk_chain/std", "bdk_tx/std"]
 compiler = ["miniscript/compiler"]
 all-keys = ["keys-bip39"]
 keys-bip39 = ["bip39"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,6 +21,7 @@ unexpected_cfgs = { level = "warn", check-cfg = ['cfg(coverage_nightly)'] }
 
 [dependencies]
 bdk_chain = { version = "0.23.3", features = ["miniscript", "serde"], default-features = false }
+bdk_tx = { version = "0.2.0", default-features = false }
 bitcoin = { version = "0.32.8", features = ["serde", "base64"], default-features = false }
 miniscript = { version = "12.3.5", features = ["serde"], default-features = false }
 rand_core = { version = "0.6.4" }
@@ -33,9 +34,10 @@ bdk_file_store = { version = "0.22.0", optional = true }
 bip39 = { version = "2.2.2", optional = true }
 tempfile = { version = "3.26.0", optional = true }
 
+
 [features]
 default = ["std"]
-std = ["bitcoin/std", "bitcoin/rand-std", "miniscript/std", "bdk_chain/std"]
+std = ["bitcoin/std", "bitcoin/rand-std", "miniscript/std", "bdk_chain/std", "bdk_tx/std"]
 compiler = ["miniscript/compiler"]
 all-keys = ["keys-bip39"]
 keys-bip39 = ["bip39"]
--- a/examples/psbt.rs
+++ b/examples/psbt.rs
@@ -1,0 +1,117 @@
+#![allow(clippy::print_stdout)]
+
+use std::collections::HashMap;
+use std::str::FromStr;
+
+use bdk_chain::BlockId;
+use bdk_chain::ConfirmationBlockTime;
+use bdk_wallet::psbt::{PsbtParams, SelectionStrategy::*};
+use bdk_wallet::test_utils::*;
+use bdk_wallet::{KeychainKind::External, Wallet};
+use bitcoin::{
+    bip32, consensus,
+    secp256k1::{self, rand},
+    Address, Amount, TxIn, TxOut,
+};
+use rand::Rng;
+
+// This example shows how to create a PSBT using BDK Wallet.
+
+const NETWORK: bitcoin::Network = bitcoin::Network::Signet;
+const SEND_TO: &str = "tb1pw3g5qvnkryghme7pyal228ekj6vq48zc5k983lqtlr2a96n4xw0q5ejknw";
+const AMOUNT: Amount = Amount::from_sat(42_000);
+const FEERATE: f64 = 2.0; // sat/vb
+
+fn main() -> anyhow::Result<()> {
+    let (desc, change_desc) = get_test_wpkh_and_change_desc();
+    let secp = secp256k1::Secp256k1::new();
+
+    // Xpriv to be used for signing the PSBT
+    let xprv = bip32::Xpriv::from_str("tprv8ZgxMBicQKsPdy6LMhUtFHAgpocR8GC6QmwMSFpZs7h6Eziw3SpThFfczTDh5rW2krkqffa11UpX3XkeTTB2FvzZKWXqPY54Y6Rq4AQ5R8L")?;
+
+    // Create wallet and fund it.
+    let mut wallet = Wallet::create(desc, change_desc)
+        .network(NETWORK)
+        .create_wallet_no_persist()?;
+
+    fund_wallet(&mut wallet)?;
+
+    let utxos = wallet
+        .list_unspent()
+        .map(|output| (output.outpoint, output))
+        .collect::<HashMap<_, _>>();
+
+    // Build params.
+    let mut params = PsbtParams::default();
+    let addr = Address::from_str(SEND_TO)?.require_network(NETWORK)?;
+    let feerate = feerate_unchecked(FEERATE);
+    params
+        .add_recipients([(addr, AMOUNT)])
+        .fee_rate(feerate)
+        .coin_selection(SingleRandomDraw);
+
+    // Create PSBT (which also returns the Finalizer).
+    let (mut psbt, finalizer) = wallet.create_psbt(params)?;
+
+    dbg!(&psbt);
+
+    let tx = &psbt.unsigned_tx;
+    for txin in &tx.input {
+        let op = txin.previous_output;
+        let output = utxos.get(&op).unwrap();
+        println!("TxIn: {}", output.txout.value);
+    }
+    for txout in &tx.output {
+        println!("TxOut: {}", txout.value);
+    }
+
+    let _ = psbt.sign(&xprv, &secp);
+    println!("Signed: {}", !psbt.inputs[0].partial_sigs.is_empty());
+    let finalize_res = finalizer.finalize(&mut psbt);
+    println!("Finalized: {}", finalize_res.is_finalized());
+
+    let tx = psbt.extract_tx()?;
+    let feerate = wallet.calculate_fee_rate(&tx)?;
+    println!("Fee rate: {} sat/vb", bdk_wallet::floating_rate!(feerate));
+
+    println!("{}", consensus::encode::serialize_hex(&tx));
+
+    Ok(())
+}
+
+fn fund_wallet(wallet: &mut Wallet) -> anyhow::Result<()> {
+    let anchor = ConfirmationBlockTime {
+        block_id: BlockId {
+            height: 260071,
+            hash: "000000099f67ae6469d1ad0525d756e24d4b02fbf27d65b3f413d5feb367ec48".parse()?,
+        },
+        confirmation_time: 1752184658,
+    };
+    insert_checkpoint(wallet, anchor.block_id);
+
+    let mut rng = rand::thread_rng();
+
+    // Fund wallet with several random utxos
+    for i in 0..21 {
+        let addr = wallet.reveal_next_address(External).address;
+        let value = 10_000 * (i + 1) + (100 * rng.gen_range(0..10));
+        let tx = bitcoin::Transaction {
+            lock_time: bitcoin::absolute::LockTime::ZERO,
+            version: bitcoin::transaction::Version::TWO,
+            input: vec![TxIn::default()],
+            output: vec![TxOut {
+                script_pubkey: addr.script_pubkey(),
+                value: Amount::from_sat(value),
+            }],
+        };
+        insert_tx_anchor(wallet, tx, anchor.block_id);
+    }
+
+    let tip = BlockId {
+        height: 260171,
+        hash: "0000000b9efb77450e753ae9fd7be9f69219511c27b6e95c28f4126f3e1591c3".parse()?,
+    };
+    insert_checkpoint(wallet, tip);
+
+    Ok(())
+}
--- a/examples/psbt.rs
+++ b/examples/psbt.rs
@@ -8,11 +8,7 @@ use bdk_chain::ConfirmationBlockTime;
 use bdk_wallet::psbt::{PsbtParams, SelectionStrategy::*};
 use bdk_wallet::test_utils::*;
 use bdk_wallet::{KeychainKind::External, Wallet};
-use bitcoin::{
-    bip32, consensus,
-    secp256k1::{self, rand},
-    Address, Amount, TxIn, TxOut,
-};
+use bitcoin::{consensus, secp256k1::rand, Address, Amount, TxIn, TxOut};
 use rand::Rng;
 
 // This example shows how to create a PSBT using BDK Wallet.
@@ -24,10 +20,6 @@ const FEERATE: f64 = 2.0; // sat/vb
 
 fn main() -> anyhow::Result<()> {
     let (desc, change_desc) = get_test_wpkh_and_change_desc();
-    let secp = secp256k1::Secp256k1::new();
-
-    // Xpriv to be used for signing the PSBT
-    let xprv = bip32::Xpriv::from_str("tprv8ZgxMBicQKsPdy6LMhUtFHAgpocR8GC6QmwMSFpZs7h6Eziw3SpThFfczTDh5rW2krkqffa11UpX3XkeTTB2FvzZKWXqPY54Y6Rq4AQ5R8L")?;
 
     // Create wallet and fund it.
     let mut wallet = Wallet::create(desc, change_desc)
@@ -35,6 +27,14 @@ fn main() -> anyhow::Result<()> {
         .create_wallet_no_persist()?;
 
     fund_wallet(&mut wallet)?;
+
+    // Create PSBT Signer, external to the wallet
+    let signer = {
+        let secp = wallet.secp_ctx();
+        let (_, external_keymap) = miniscript::Descriptor::parse_descriptor(secp, desc)?;
+        let (_, internal_keymap) = miniscript::Descriptor::parse_descriptor(secp, change_desc)?;
+        bdk_tx::Signer(external_keymap.into_iter().chain(internal_keymap).collect())
+    };
 
     let utxos = wallet
         .list_unspent()
@@ -65,7 +65,10 @@ fn main() -> anyhow::Result<()> {
         println!("TxOut: {}", txout.value);
     }
 
-    let _ = psbt.sign(&xprv, &secp);
+    let _ = psbt
+        .sign(&signer, wallet.secp_ctx())
+        .map_err(|(_, errors)| anyhow::anyhow!("failed to sign PSBT: {errors:?}"))?;
+
     println!("Signed: {}", !psbt.inputs[0].partial_sigs.is_empty());
     let finalize_res = finalizer.finalize(&mut psbt);
     println!("Finalized: {}", finalize_res.is_finalized());
--- a/examples/psbt.rs
+++ b/examples/psbt.rs
@@ -53,8 +53,6 @@ fn main() -> anyhow::Result<()> {
     // Create PSBT (which also returns the Finalizer).
     let (mut psbt, finalizer) = wallet.create_psbt(params)?;
 
-    dbg!(&psbt);
-
     let tx = &psbt.unsigned_tx;
     for txin in &tx.input {
         let op = txin.previous_output;
--- a/examples/psbt.rs
+++ b/examples/psbt.rs
@@ -1,0 +1,120 @@
+#![allow(clippy::print_stdout)]
+
+use std::collections::HashMap;
+use std::str::FromStr;
+
+use bdk_chain::BlockId;
+use bdk_chain::ConfirmationBlockTime;
+use bdk_wallet::psbt::{PsbtParams, SelectionStrategy::*};
+use bdk_wallet::test_utils::*;
+use bdk_wallet::{KeychainKind::External, Wallet};
+use bitcoin::{consensus, secp256k1::rand, Address, Amount, TxIn, TxOut};
+use rand::Rng;
+
+// This example shows how to create a PSBT using BDK Wallet.
+
+const NETWORK: bitcoin::Network = bitcoin::Network::Signet;
+const SEND_TO: &str = "tb1pw3g5qvnkryghme7pyal228ekj6vq48zc5k983lqtlr2a96n4xw0q5ejknw";
+const AMOUNT: Amount = Amount::from_sat(42_000);
+const FEERATE: f64 = 2.0; // sat/vb
+
+fn main() -> anyhow::Result<()> {
+    let (desc, change_desc) = get_test_wpkh_and_change_desc();
+
+    // Create wallet and fund it.
+    let mut wallet = Wallet::create(desc, change_desc)
+        .network(NETWORK)
+        .create_wallet_no_persist()?;
+
+    fund_wallet(&mut wallet)?;
+
+    // Create PSBT Signer, external to the wallet
+    let signer = {
+        let secp = wallet.secp_ctx();
+        let (_, external_keymap) = miniscript::Descriptor::parse_descriptor(secp, desc)?;
+        let (_, internal_keymap) = miniscript::Descriptor::parse_descriptor(secp, change_desc)?;
+        bdk_tx::Signer(external_keymap.into_iter().chain(internal_keymap).collect())
+    };
+
+    let utxos = wallet
+        .list_unspent()
+        .map(|output| (output.outpoint, output))
+        .collect::<HashMap<_, _>>();
+
+    // Build params.
+    let mut params = PsbtParams::default();
+    let addr = Address::from_str(SEND_TO)?.require_network(NETWORK)?;
+    let feerate = feerate_unchecked(FEERATE);
+    params
+        .add_recipients([(addr, AMOUNT)])
+        .fee_rate(feerate)
+        .coin_selection(SingleRandomDraw);
+
+    // Create PSBT (which also returns the Finalizer).
+    let (mut psbt, finalizer) = wallet.create_psbt(params)?;
+
+    dbg!(&psbt);
+
+    let tx = &psbt.unsigned_tx;
+    for txin in &tx.input {
+        let op = txin.previous_output;
+        let output = utxos.get(&op).unwrap();
+        println!("TxIn: {}", output.txout.value);
+    }
+    for txout in &tx.output {
+        println!("TxOut: {}", txout.value);
+    }
+
+    let _ = psbt
+        .sign(&signer, wallet.secp_ctx())
+        .map_err(|(_, errors)| anyhow::anyhow!("failed to sign PSBT: {errors:?}"))?;
+
+    println!("Signed: {}", !psbt.inputs[0].partial_sigs.is_empty());
+    let finalize_res = finalizer.finalize(&mut psbt);
+    println!("Finalized: {}", finalize_res.is_finalized());
+
+    let tx = psbt.extract_tx()?;
+    let feerate = wallet.calculate_fee_rate(&tx)?;
+    println!("Fee rate: {} sat/vb", bdk_wallet::floating_rate!(feerate));
+
+    println!("{}", consensus::encode::serialize_hex(&tx));
+
+    Ok(())
+}
+
+fn fund_wallet(wallet: &mut Wallet) -> anyhow::Result<()> {
+    let anchor = ConfirmationBlockTime {
+        block_id: BlockId {
+            height: 260071,
+            hash: "000000099f67ae6469d1ad0525d756e24d4b02fbf27d65b3f413d5feb367ec48".parse()?,
+        },
+        confirmation_time: 1752184658,
+    };
+    insert_checkpoint(wallet, anchor.block_id);
+
+    let mut rng = rand::thread_rng();
+
+    // Fund wallet with several random utxos
+    for i in 0..21 {
+        let addr = wallet.reveal_next_address(External).address;
+        let value = 10_000 * (i + 1) + (100 * rng.gen_range(0..10));
+        let tx = bitcoin::Transaction {
+            lock_time: bitcoin::absolute::LockTime::ZERO,
+            version: bitcoin::transaction::Version::TWO,
+            input: vec![TxIn::default()],
+            output: vec![TxOut {
+                script_pubkey: addr.script_pubkey(),
+                value: Amount::from_sat(value),
+            }],
+        };
+        insert_tx_anchor(wallet, tx, anchor.block_id);
+    }
+
+    let tip = BlockId {
+        height: 260171,
+        hash: "0000000b9efb77450e753ae9fd7be9f69219511c27b6e95c28f4126f3e1591c3".parse()?,
+    };
+    insert_checkpoint(wallet, tip);
+
+    Ok(())
+}
--- a/src/psbt/params.rs
+++ b/src/psbt/params.rs
@@ -1,0 +1,632 @@
+//! Parameters for creating a PSBT.
+
+use alloc::sync::Arc;
+use alloc::vec::Vec;
+use core::fmt;
+
+use bdk_chain::{BlockId, CanonicalizationParams, ConfirmationBlockTime, FullTxOut, TxGraph};
+use bdk_tx::{ChangeScript, Input, Output};
+use bitcoin::{
+    absolute, psbt::PsbtSighashType, transaction::Version, Amount, FeeRate, OutPoint, ScriptBuf,
+    Sequence, Transaction, Txid,
+};
+use miniscript::plan::Assets;
+
+use crate::collections::HashSet;
+use crate::TxOrdering;
+
+/// Marker type representing the PSBT creation state.
+#[derive(Debug)]
+pub struct CreateTx;
+
+/// Marker type representing the Replace-By-Fee (RBF) state.
+#[derive(Debug)]
+pub struct ReplaceTx;
+
+/// Alias for [`ReplaceTx`] context marker.
+pub type Rbf = ReplaceTx;
+
+/// Parameters to create a PSBT.
+// TODO: Can we derive `Clone` for this?
+#[derive(Debug)]
+pub struct PsbtParams<C> {
+    /// Set of selected UTXO outpoints.
+    pub(crate) set: HashSet<OutPoint>,
+    /// List of UTXO outpoints to spend.
+    pub(crate) utxos: Vec<OutPoint>,
+    /// List of planned transaction [`Input`]s.
+    pub(crate) inputs: Vec<Input>,
+    /// List of recipient script/amount pairs.
+    pub(crate) recipients: Vec<(ScriptBuf, Amount)>,
+    /// Optional script or descriptor designated for change.
+    pub(crate) change_script: Option<ChangeScript>,
+    /// Optional assets for creating a spend plan.
+    pub(crate) assets: Option<Assets>,
+    /// Target fee rate.
+    pub(crate) fee_rate: FeeRate,
+    /// Whether to spend all available coins.
+    pub(crate) drain_wallet: bool,
+    /// Coin selection strategy to use.
+    pub(crate) coin_selection: SelectionStrategy,
+    /// Parameters for transaction canonicalization.
+    pub(crate) canonical_params: CanonicalizationParams,
+    /// UTXO filtering function.
+    pub(crate) utxo_filter: UtxoFilter,
+    /// Optional height for evaluating coinbase maturity.
+    pub(crate) maturity_height: Option<u32>,
+    /// Only allow spending UTXOs which are selected manually.
+    pub(crate) manually_selected_only: bool,
+    /// Optional transaction [`Version`].
+    pub(crate) version: Option<Version>,
+    /// Optional transaction [`LockTime`](absolute::LockTime).
+    pub(crate) locktime: Option<absolute::LockTime>,
+    /// Optional fallback [`Sequence`] for inputs.
+    pub(crate) fallback_sequence: Option<Sequence>,
+    /// Ordering of the transaction's inputs and outputs.
+    pub(crate) ordering: TxOrdering<Input, Output>,
+    /// Only set the [`witness_utxo`](bitcoin::psbt::Input::witness_utxo) in PSBT inputs. This
+    /// allows opting out of setting the
+    /// [`non_witness_utxo`](bitcoin::psbt::Input::non_witness_utxo).
+    pub(crate) only_witness_utxo: bool,
+    /// Optional PSBT sighash type.
+    pub(crate) sighash_type: Option<PsbtSighashType>,
+    /// Whether to try filling in the PSBT global xpubs from the wallet's descriptors.
+    pub(crate) add_global_xpubs: bool,
+    /// Whether to enable BIP326 Sequence based anti-fee sniping.
+    pub(crate) enable_anti_fee_sniping: bool,
+    /// Set of txids being replaced if this is a RBF transaction.
+    pub(crate) replace: HashSet<Txid>,
+    /// The context in which the params are used.
+    pub(crate) marker: core::marker::PhantomData<C>,
+}
+
+impl Default for PsbtParams<CreateTx> {
+    fn default() -> Self {
+        Self {
+            set: Default::default(),
+            utxos: Default::default(),
+            inputs: Default::default(),
+            assets: Default::default(),
+            recipients: Default::default(),
+            change_script: Default::default(),
+            fee_rate: FeeRate::BROADCAST_MIN,
+            drain_wallet: Default::default(),
+            coin_selection: Default::default(),
+            canonical_params: Default::default(),
+            utxo_filter: Default::default(),
+            maturity_height: Default::default(),
+            manually_selected_only: Default::default(),
+            version: Default::default(),
+            locktime: Default::default(),
+            fallback_sequence: Default::default(),
+            ordering: Default::default(),
+            only_witness_utxo: Default::default(),
+            sighash_type: Default::default(),
+            add_global_xpubs: Default::default(),
+            enable_anti_fee_sniping: Default::default(),
+            replace: Default::default(),
+            marker: core::marker::PhantomData,
+        }
+    }
+}
+
+impl PsbtParams<CreateTx> {
+    /// Create a new [`PsbtParams`].
+    pub fn new() -> Self {
+        Self::default()
+    }
+
+    /// Add UTXOs by outpoint to fund the transaction.
+    ///
+    /// A single outpoint may appear at most once in the list of UTXOs to spend. The caller is
+    /// responsible for ensuring that items of `outpoints` correspond to outputs of previous
+    /// transactions and are currently unspent.
+    ///
+    /// If an outpoint doesn't correspond to an indexed script pubkey, a [`UnknownUtxo`]
+    /// error will occur. See [`Wallet::create_psbt`] for more.
+    ///
+    /// To add a UTXO that did not originate from this wallet (i.e. a "foreign" UTXO), see
+    /// [`PsbtParams::add_planned_input`].
+    ///
+    /// [`UnknownUtxo`]: crate::wallet::error::CreatePsbtError::UnknownUtxo
+    /// [`Wallet::create_psbt`]: crate::Wallet::create_psbt
+    pub fn add_utxos(&mut self, outpoints: &[OutPoint]) -> &mut Self {
+        self.utxos
+            .extend(outpoints.iter().copied().filter(|&op| self.set.insert(op)));
+        self
+    }
+
+    /// Replace spends of the provided `txs` and return a [`PsbtParams`] populated with the
+    /// inputs to spend.
+    ///
+    /// This merges all of the spends into a single transaction while retaining the parameters
+    /// of `self`. Note that any previously added UTXOs are removed. Call
+    /// [`replace_by_fee_with_rng`](crate::Wallet::replace_by_fee_with_rng) to finish
+    /// building the PSBT.
+    ///
+    /// ## Note
+    ///
+    /// There should be no ancestry linking the elements of `txs`, since replacing an
+    /// ancestor necessarily invalidates the descendant.
+    pub fn replace_txs(self, txs: &[Arc<Transaction>]) -> PsbtParams<Rbf> {
+        let mut params = self.into_replace_params();
+        params.replace(txs);
+        params
+    }
+
+    /// Transition this [`PsbtParams`] to the [`Rbf`] state.
+    fn into_replace_params(self) -> PsbtParams<Rbf> {
+        PsbtParams {
+            set: self.set,
+            utxos: self.utxos,
+            inputs: self.inputs,
+            assets: self.assets,
+            recipients: self.recipients,
+            change_script: self.change_script,
+            fee_rate: self.fee_rate,
+            drain_wallet: self.drain_wallet,
+            coin_selection: self.coin_selection,
+            canonical_params: self.canonical_params,
+            utxo_filter: self.utxo_filter,
+            maturity_height: self.maturity_height,
+            manually_selected_only: self.manually_selected_only,
+            version: self.version,
+            locktime: self.locktime,
+            fallback_sequence: self.fallback_sequence,
+            ordering: self.ordering,
+            only_witness_utxo: self.only_witness_utxo,
+            sighash_type: self.sighash_type,
+            add_global_xpubs: self.add_global_xpubs,
+            enable_anti_fee_sniping: self.enable_anti_fee_sniping,
+            replace: self.replace,
+            marker: core::marker::PhantomData,
+        }
+    }
+}
+
+impl<C> PsbtParams<C> {
+    /// Get the currently selected spends.
+    pub fn utxos(&self) -> &HashSet<OutPoint> {
+        &self.set
+    }
+
+    /// Remove a UTXO from the currently selected inputs.
+    pub fn remove_utxo(&mut self, outpoint: &OutPoint) -> &mut Self {
+        if self.set.remove(outpoint) {
+            self.utxos.retain(|op| op != outpoint);
+        }
+        self
+    }
+
+    /// Only include inputs that are selected manually using [`add_utxos`] or [`add_planned_input`].
+    ///
+    /// Since the wallet will skip coin selection for additional candidates, the manually selected
+    /// inputs must be enough to fund the transaction or else an error will be thrown due to
+    /// insufficient funds.
+    ///
+    /// [`add_utxos`]: PsbtParams::add_utxos
+    /// [`add_planned_input`]: PsbtParams::add_planned_input
+    pub fn manually_selected_only(&mut self) -> &mut Self {
+        self.manually_selected_only = true;
+        self
+    }
+
+    /// Add the spend [`Assets`].
+    ///
+    /// Assets are required to create a spending plan for an output controlled by the wallet's
+    /// descriptors. If none are provided here, then we assume all of the keys are equally likely
+    /// to sign.
+    ///
+    /// This may be called multiple times to add additional assets, however only the last
+    /// absolute or relative timelock is retained.
+    pub fn add_assets(&mut self, assets: Assets) -> &mut Self {
+        let mut new = match self.assets {
+            Some(ref existing) => {
+                let mut new = Assets::new();
+                new.extend(existing);
+                new
+            }
+            None => Assets::new(),
+        };
+        new.extend(&assets);
+        self.assets = Some(new);
+        self
+    }
+
+    /// Add outgoing recipients to the transaction.
+    ///
+    /// - `recipients`: An iterator of `(S, Amount)` tuples where `S` can be a [`bitcoin::Address`],
+    ///   a script pubkey, or anything that can be converted straight into a [`ScriptBuf`].
+    pub fn add_recipients<I, S>(&mut self, recipients: I) -> &mut Self
+    where
+        I: IntoIterator<Item = (S, Amount)>,
+        S: Into<ScriptBuf>,
+    {
+        self.recipients
+            .extend(recipients.into_iter().map(|(s, amt)| (s.into(), amt)));
+        self
+    }
+
+    /// Set the transaction `nLockTime`.
+    ///
+    /// This can be used as a fallback in case none of the inputs to the transaction require an
+    /// absolute locktime. If no locktime is required and nothing is specified here, then the
+    /// locktime is set to the last known chain tip.
+    pub fn locktime(&mut self, locktime: absolute::LockTime) -> &mut Self {
+        self.locktime = Some(locktime);
+        self
+    }
+
+    /// Set the height to be used when evaluating the maturity of coinbase outputs during coin
+    /// selection.
+    pub fn maturity_height(&mut self, height: absolute::Height) -> &mut Self {
+        self.maturity_height = Some(height.to_consensus_u32());
+        self
+    }
+
+    /// Set the target [`FeeRate`].
+    ///
+    /// If not set, defaults to [`FeeRate::BROADCAST_MIN`].
+    pub fn fee_rate(&mut self, fee_rate: FeeRate) -> &mut Self {
+        self.fee_rate = fee_rate;
+        self
+    }
+
+    /// Set the strategy to be used when selecting coins.
+    pub fn coin_selection(&mut self, strategy: SelectionStrategy) -> &mut Self {
+        self.coin_selection = strategy;
+        self
+    }
+
+    /// Set the parameters for modifying the wallet's view of canonical transactions.
+    ///
+    /// The `params` can be used to resolve conflicts manually, or to assert that a particular
+    /// transaction should be treated as canonical for the purpose of building the current PSBT.
+    /// Refer to [`CanonicalizationParams`] for more.
+    pub fn canonicalization_params(
+        &mut self,
+        params: bdk_chain::CanonicalizationParams,
+    ) -> &mut Self {
+        self.canonical_params = params;
+        self
+    }
+
+    /// Set the [`Descriptor`] or raw [`Script`] to be used for generating the change output.
+    ///
+    /// [`Descriptor`]: ChangeScript::Descriptor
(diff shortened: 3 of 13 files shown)


thunderbiscuit (commented):
So this is more of a conceptual review/question set, as I'm getting acquainted with the PR, and it also requires knowledge and understanding of the bdk-tx crate/workflow.

One general question I have is do you think is missing in functionality between this and the current TxBuilder? Can we make a todo list that compares functionality with the current TxBuilder to better visualize how close of a replacement this is, or if it only provides part of the functionality for now (and if so which parts)?

I haven't had time to look/test the examples and my day is over, but I'll come back to this on Monday.

wallet/Cargo.toml
```diff
@@ -22,12 +22,14 @@ bitcoin = { version = "0.32.6", features = [ "serde", "base64" ], default-featur
...
-anyhow = { version = "1.0.98", optional = true }
-tempfile = { version = "3.20.0", optional = true }
+bdk_tx = { version = "0.1.0" }
+bdk_coin_select = { version = "0.4.0" }
 
 # Optional dependencies
 bip39 = { version = "2.0", optional = true }
 bdk_file_store = { version = "0.21.1", optional = true }
+anyhow = { version = "1.0.98", optional = true }
+tempfile = { version = "3.20.0", optional = true }
```
oleonardolima, line 32: I guess this can be removed, as it's not being used.

wallet/src/wallet/error.rs
```diff
@@ -256,3 +256,68 @@ impl fmt::Display for BuildFeeBumpError {
...
 impl std::error::Error for BuildFeeBumpError {}
+
+/// Error when creating a PSBT.
+#[derive(Debug)]
+#[non_exhaustive]
+pub enum CreatePsbtError {
+    /// No Bnb solution.
+    Bnb(bdk_coin_select::NoBnbSolution),
+    /// Non-sufficient funds
+    InsufficientFunds(bdk_coin_select::InsufficientFunds),
```
oleonardolima, line 267: Can't we have a generic CS error instead ?

wallet/src/wallet/mod.rs
```diff
@@ -2635,6 +2687,495 @@ impl Wallet {
...
+    ///     .add_utxos(&[outpoint])
+    ///     .add_recipients([(address, amount)])
+    ///     .coin_selection(SelectionStrategy::LowestFee)
+    ///     .feerate(FeeRate::BROADCAST_MIN);
+    ///
+    /// let (psbt, finalizer) = wallet.create_psbt(params)?;
+    /// # Ok::<_, anyhow::Error>(())
+    /// ```
+    #[cfg(feature = "std")]
+    pub fn create_psbt(&self, params: PsbtParams) -> Result<(Psbt, Finalizer), CreatePsbtError> {
```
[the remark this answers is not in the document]
oleonardolima: I still need to take a deeper look into bdk-tx, but I think the name could be misleading with `Input Finalizer` from the PSBT BIP-174 (if it's not following the same proposed idea).

oleonardolima (commented):
I did an initial review, with simple comments/questions. Still need to take a deeper look into bdk-tx and do a more thorough review here.

wallet/src/wallet/mod.rs
```diff
@@ -2635,6 +2687,495 @@ impl Wallet {
...
+
+        Input::from_prev_tx(plan, tx, op.vout as usize, tx_status).ok()
+    }
+
+    /// Attempt to create a spending plan for the UTXO of the given `outpoint`
+    /// with the provided `assets`.
+    ///
+    /// Return `None` if `outpoint` doesn't correspond to an indexed txout, or
+    /// if the assets are not sufficient to create a plan.
+    fn try_plan(&self, outpoint: OutPoint, assets: &Assets) -> Option<Plan> {
```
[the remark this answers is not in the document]
ValuedMammal: https://hackmd.io/@valuedmammal/B1lxzomj0

wallet/src/psbt/params.rs
```diff
@@ -0,0 +1,335 @@
...
+        }
+    }
+}
+
+impl PsbtParams {
+    /// Add UTXOs by outpoint to fund the transaction.
+    ///
+    /// A single outpoint may appear at most once in the list of UTXOs to spend. The caller is
+    /// responsible for ensuring that elements of `outpoints` correspond to outputs of previous
+    /// transactions and are currently unspent.
```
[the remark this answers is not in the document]
ValuedMammal: 🤔

wallet/src/psbt/params.rs
```diff
@@ -0,0 +1,335 @@
...
+        ReplaceParams::new(txs, self)
+    }
+}
+
+// TODO: Bring back `TxOrdering`
+
+/// `ReplaceParams` provides a thin wrapper around [`PsbtParams`] and is intended for
+/// crafting Replace-By-Fee transactions (RBF).
+#[derive(Debug, Default)]
+pub struct ReplaceParams {
```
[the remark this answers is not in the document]
ValuedMammal: > Naively I would have thought you'd just call `PsbtParams::replace` and you'd get a sort of pre-populated PsbtParams ready for replacing your tx

That works too.

wallet/src/psbt/params.rs
```diff
@@ -0,0 +1,335 @@
...
+    }
+
+    /// Set the strategy to be used when selecting coins.
+    pub fn coin_selection(&mut self, strategy: SelectionStrategy) -> &mut Self {
+        self.coin_selection = strategy;
+        self
+    }
+
+    /// Set the definite descriptor used for generating the change output.
+    pub fn change_descriptor(&mut self, desc: DefiniteDescriptor) -> &mut Self {
```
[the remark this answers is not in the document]
thunderbiscuit: Ok awesome thanks for confirming. In this case, I'd just mention this in the docs (that this method is basically "if you want to send change elsewhere than your default change location)"

wallet/src/psbt/params.rs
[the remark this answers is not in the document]
thunderbiscuit: If that works that's my preferred approach, unless there is something I'm not seeing.

ValuedMammal:
### TxBuilder vs PsbtParams feature comparison

*subject to change*

| Feature                              | TxBuilder  | PsbtParams | Status    |
| ---------                            | ---------- | ---------- |---------- |
| internal_policy_path                 |    ✅      |     ❌     | NA    |
| external_policy_path                 |    ✅      |     ❌     | NA    |
| manually_selected_only               |    ✅      |     ✅     | Done  |
| include_output_redeem_witness_script |    ✅      |     ❌     | NA    |
| current_height                       |    ✅      |     ✅     |  Done |
| fee_rate                             |    ✅      |     ✅     |  Done |
| fee_absolute                         |    ✅      |     ❌     |  _    |
| sighash (type)                       |    ✅      |     ✅     |  Done |
| ordering                             |    ✅      |     ✅     |  Done |
| add_global_xpubs                     |    ✅      |     ✅     |  Done |
| allow_dust                           |    ✅      |     ❌     |  _    |
| unspendable                          |    ✅      |     ❌     |  NA   |
| recipients                           |    ✅      |     ✅     |  Done |
| utxos                                |    ✅      |     ✅     |  Done |
| drain_wallet                         |    ✅      |     ✅     |  Done |
| drain_to/change_script               |    ✅      |     ✅     |  Done |
| locktime                             |    ✅      |     ✅     |  Done |
| version                              |    ✅      |     ✅     |  Done |
| fallback_sequence                    |    ✅      |     ✅     |  Done |
| bumping_fee                          |    ✅      |     ✅     |  Done |
| only_witness_utxo                    |    ✅      |     ✅     |  Done |
| change_policy                        |    ✅      |     ✅     |  Done |
| add_foreign_utxo / planned_input     |    ✅      |     ✅     |  Done |
| filter_utxos                         |    ❌      |     ✅     |  Done |

N/A:
    - `*_policy_path`: deprecated, use `add_assets` instead
    - `include_output_redeem_witness_script`: deprecated, no longer used
    - `unspendable`: NA, refer to `filter_utxos`
    - `change_policy`: NA, refer to `filter_utxos`

ValuedMammal:
- We need a way to implement `fee_absolute` or find out if this is possible using `bdk_coin_select`.
- Note that `UtxoFilter` can accomplish a number of things from the old interface, such as "change policy" and "unspendable".

ValuedMammal:
> We need a way to implement `fee_absolute` or find out if this is possible using `bdk_coin_select`.

I think a straightforward approach is to set a target feerate of `0` and add the fee amount to the output [`value_sum`](https://docs.rs/bdk_coin_select/latest/bdk_coin_select/struct.TargetOutputs.html#structfield.value_sum). The effect is to raise the cost of meeting the target by the amount of the target fee.

I can see it being relevant to CPFP, since there we're less concerned about the feerate of an individual tx and more so with the fee needed to bump a package to a target feerate.

wallet/Cargo.toml
ValuedMammal: These are currently being used by the `test-utils` feature, but there could be a way to convert them back to dev dependencies.

src/psbt/params.rs
```diff
@@ -0,0 +1,335 @@
...
+        ReplaceParams::new(txs, self)
+    }
+}
+
+// TODO: Bring back `TxOrdering`
+
+/// `ReplaceParams` provides a thin wrapper around [`PsbtParams`] and is intended for
+/// crafting Replace-By-Fee transactions (RBF).
+#[derive(Debug, Default)]
+pub struct ReplaceParams {
```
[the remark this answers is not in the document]
ValuedMammal: Removed `ReplaceParams`. Now calling `PsbtParams::replace_txs` will return a new `PsbtParams` that is marked by the "RBF" context. I think it's a cleaner approach overall without compromising type safety.

src/psbt/params.rs
```diff
@@ -0,0 +1,693 @@
...
+    /// List of planned transaction [`Input`]s.
+    pub(crate) inputs: Vec<Input>,
+    /// List of recipient script/amount pairs.
+    pub(crate) recipients: Vec<(ScriptBuf, Amount)>,
+    /// Optional script or descriptor designated for change.
+    pub(crate) change_script: Option<ScriptSource>,
+    /// Optional assets for creating a spend plan.
+    pub(crate) assets: Option<Assets>,
+    /// Fee targeting strategy.
+    pub(crate) fee_strategy: FeeStrategy,
```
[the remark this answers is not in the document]
DanGould: This is what our API looks like to solve our fee problems for receivers: https://docs.rs/payjoin/1.0.0-rc.1/payjoin/receive/v2/struct.Receiver.html#impl-Receiver%3CWantsFeeRange%3E

Internally we enforce the minimum absolute fee and we enable a receiver to specify their minimum fee rate when the initialize the receiver (though I suppose in time it should really be a function rather than a value) as well as a maximum fee rate to prevent overspending by a counterparty who insists on a higher feerate.

very much a nit, but '"ensures" the [transaction] gets received in the next block' is necessarily interactive, and still only a best effort, since all we can do is estimate in this function afaiu.

src/psbt/params.rs
```diff
@@ -0,0 +1,697 @@
...
+    /// Assets are required to create a spending plan for an output controlled by the wallet's
+    /// descriptors. If none are provided here, then we assume all of the keys are equally likely
+    /// to sign.
+    ///
+    /// This may be called multiple times to add additional assets, however only the last
+    /// absolute or relative timelock is retained.
+    ///
+    /// If adding a time-based absolute timelock here, you must also set a [`PsbtParams::locktime`]
+    /// which is time-based (i.e. UNIX timestamp) to use as a fallback, otherwise a
+    /// `LockTypeMismatch` error will occur.
```
ValuedMammal: I find this slightly annoying and there could be an alternative way to handle it. Edit: Fixed by bitcoindevkit/bdk-tx#43.

noahjoeris:
One thing I noticed is that bdk-tx test coverage is a bit thin right? Might be good to add some more before merging. I’ll have a look and see if I can help :)

ValuedMammal:
- Changed the `examples/replace_by_fee.rs` to a more practical example of a sweep transaction that gets replaced with a higher fee

ValuedMammal:
In the last push

- Removed `PsbtParams::change_policy` member
- Improved `SelectionStrategy::LowestFee` by exposing longterm_feerate, max_rounds
- Added `ChangeScript` support to `PsbtParams`. Since `ChangeScript` is not yet Clone-able, had to add a separate `Wallet::change_script` helper, and update `parse_params` to handle only the assets and indexed txouts

src/wallet/mod.rs
```diff
@@ -2778,6 +2844,586 @@ impl Wallet {
...
+
+        let version = params.version.unwrap_or(transaction::Version::TWO);
+        let fallback_locktime = params
+            .locktime
+            .unwrap_or(absolute::LockTime::from_consensus(
+                self.chain.tip().height(),
+            ));
+        let fallback_sequence = params
+            .fallback_sequence
+            .unwrap_or(Sequence::ENABLE_LOCKTIME_NO_RBF);
```
noahjoeris: Why default to `ENABLE_LOCKTIME_NO_RBF` here? Before we used `ENABLE_RBF_NO_LOCKTIME`

examples/psbt.rs
```diff
@@ -0,0 +1,117 @@
...
+    for txin in &tx.input {
+        let op = txin.previous_output;
+        let output = utxos.get(&op).unwrap();
+        println!("TxIn: {}", output.txout.value);
+    }
+    for txout in &tx.output {
+        println!("TxOut: {}", txout.value);
+    }
+
+    let _ = psbt.sign(&xprv, &secp);
```
noahjoeris: Should we show the real signing process here by using the wallet instead of hardcoded xpriv?

Something like
```rust
    let external_keymap = wallet.get_signers(External).as_key_map(wallet.secp_ctx());
    let internal_keymap = wallet.get_signers(Internal).as_key_map(wallet.secp_ctx());
    let signer = Signer(external_keymap.into_iter().chain(internal_keymap).collect());

    let _ = psbt
        .sign(&signer, wallet.secp_ctx())
        .map_err(|(_, errors)| anyhow::anyhow!("failed to sign PSBT: {errors:?}"))?;
```

src/wallet/mod.rs
```diff
@@ -2778,6 +2844,586 @@ impl Wallet {
...
+
+        Assets::new().add(pks)
+    }
+
+    /// Obtain the next change script (descriptor) from the wallet.
+    fn change_script(&self, change_script: Option<ChangeScript>) -> ChangeScript {
+        change_script.unwrap_or_else(|| {
+            let change_keychain = self.map_keychain(KeychainKind::Internal);
+            let descriptor = self.public_descriptor(change_keychain);
+            let next_index = self.next_derivation_index(change_keychain);
```
noahjoeris: This can reuse the same change script on repeated calls, since this only peeks the next index without revealing it, right? That would be bad for privacy

src/psbt/params.rs
```diff
@@ -0,0 +1,673 @@
...
+    /// Get the currently selected spends.
+    pub fn utxos(&self) -> &HashSet<OutPoint> {
+        &self.set
+    }
+
+    /// Remove a UTXO from the currently selected inputs.
+    pub fn remove_utxo(&mut self, outpoint: &OutPoint) -> &mut Self {
+        if self.set.remove(outpoint) {
+            self.utxos.retain(|op| op != outpoint);
+        }
```
noahjoeris, line 271: Should this also remove matching planned inputs?

src/wallet/mod.rs
```diff
@@ -2778,6 +2844,586 @@ impl Wallet {
...
+        let input_candidates = InputCandidates::new(must_spend, may_spend);
+        if input_candidates.inputs().next().is_none() {
+            let target_amount: Amount = target_outputs.iter().map(|output| output.value).sum();
+            let err = bdk_coin_select::InsufficientFunds {
+                missing: target_amount.to_sat(),
+            };
+            return Err(CreatePsbtError::InsufficientFunds(err))?;
+        }
+
+        let original_txs: Vec<OriginalTxStats> = txids_to_replace
```
noahjoeris: `build_fee_bump` rejected confirmed originals before bumping and also checked for RBF signaling. Should we reject confirmed txs here too? Otherwise we could return a PSBT that cannot be accepted as an RBF replacement.

noahjoeris (commented):
Looking very good overall ✨  I left some small comments

ValuedMammal:
Thank you @noahjoeris

src/wallet/mod.rs
ValuedMammal: I'll take another look at it. I'm thinking the behavior should be the "next unused address", so if every revealed internal address has been used, we reveal the next one. But until it's used in a transaction and applied to the wallet, it'll be "reused" for the change address. This prevents a runaway in the internal derivation index if many txs are crafted but never go on chain. bitcoindevkit/bdk#1578.

In a multi-keychain world it'd be best to avoid making this decision internally and have the user provide the change script, although it's a bit more hands-on.

src/wallet/mod.rs
ValuedMammal: The default was chosen on the assumption that most of the network is supporting `-mempoolfullrbf` (so no explicit signaling required), and in the future signaling BIP125 can make it easier to fingerprint wallets. #66

src/wallet/mod.rs
ValuedMammal: Yes I think it should reject confirmed txs.

src/wallet/mod.rs
ValuedMammal: Done in 3154047a

examples/psbt.rs
ValuedMammal: I opted to keep the keymaps external to the wallet thinking that `wallet.get_signers()` might be discouraged in the future.

src/wallet/mod.rs
```diff
@@ -2778,6 +2844,705 @@ impl Wallet {
...
+        &mut self,
+        mut params: PsbtParams<Rbf>,
+        rng: &mut impl RngCore,
+    ) -> Result<(Psbt, Finalizer), ReplaceByFeeError> {
+        let (change_info, change_script) = self.peek_change_info(params.change_script.take());
+
+        let (assets, txouts) = self.parse_params(&params);
+
+        let PsbtParams {
+            replace: txids_to_replace,
```
reez, line 3338: Edge case, but should we explicitly guard that at least one selected input overlaps with the inputs of the transactions being replaced? 

The normal path does add those original inputs as `must_spend`, but I think `replace_txs` can be called with an empty set, and `PsbtParams<Rbf>` still allows those UTXOs to be removed via `remove_utxo` before `replace_by_fee` is called. 

I wrote a quick test for that case and it looks like the wallet can still build a PSBT funded by unrelated wallet UTXOs. And even with Full RBF the replacement depends on spending at least one same input, so I think it could be checked directly.

Definitely let me know if I'm missing something though

reez (commented):
Overall LGTM, added one comment/question on an edge case

ValuedMammal:
Changes

- Update `bdk_tx` dependency from git ref to 0.2.0
- Align `PsbtParams` with the updated `bdk_tx` API by supporting `min_locktime`, add per-input Sequence overrides, drop `sighash_type`, change `enable_anti_fee_sniping: bool` to `anti_fee_sniping_height: Option<absolute::Height>`
- `filter_spendable` uses `params.set.contains()` instead of `params.utxos.contains()` to benefit from O(1) HashSet lookup
- `replace_by_fee_with_rng` uses `params.canonical_params` consistently when building the chain-positions map for the confirmation check, since the same `CanonicalizationParams` are used to build the set of indexed txouts
- Use `core::error::Error` for `CreatePsbtError` and `ReplaceByFeeError` to match the style of the code base
- Sort inputs and outputs on the Selection directly instead of going through TxOrdering
- Enforce replacement semantics in `replace_by_fee_with_rng` h/t @reez, throw `ReplaceByFeeError::NoInputsFromOriginal(Txid)` otherwise

reez:
> Changes
> 
> * Update `bdk_tx` dependency from git ref to 0.2.0
> * Align `PsbtParams` with the updated `bdk_tx` API by supporting `min_locktime`, add per-input Sequence overrides, drop `sighash_type`, change `enable_anti_fee_sniping: bool` to `anti_fee_sniping_height: Option<absolute::Height>`
> * `filter_spendable` uses `params.set.contains()` instead of `params.utxos.contains()` to benefit from O(1) HashSet lookup
> * `replace_by_fee_with_rng` uses `params.canonical_params` consistently when building the chain-positions map for the confirmation check, since the same `CanonicalizationParams` are used to build the set of indexed txouts
> * Use `core::error::Error` for `CreatePsbtError` and `ReplaceByFeeError` to match the style of the code base
> * Sort inputs and outputs on the Selection directly instead of going through TxOrdering
> * Enforce replacement semantics in `replace_by_fee_with_rng` h/t @reez, throw `ReplaceByFeeError::NoInputsFromOriginal(Txid)` otherwise

ACK 43adb6ea46ebf87fdfc9e5ff078815e8a5d5edc5

Thanks for looking at that RBF edge case

src/wallet/mod.rs
reez: resolved in https://github.com/bitcoindevkit/bdk_wallet/pull/297#issuecomment-4617634935

src/psbt/params.rs
```diff
@@ -0,0 +1,706 @@
...
+    /// ## Note
+    ///
+    /// There should be no ancestry linking the elements of `txs`, since replacing an
+    /// ancestor necessarily invalidates the descendant.
+    ///
+    /// # Panics
+    ///
+    /// Panics if `txs` is empty.
+    pub fn replace_txs(self, txs: &[Arc<Transaction>]) -> PsbtParams<Rbf> {
+        assert!(
```
notmandatory: Can we return a result here instead of this assert? Would rather not have a panic caused by user input validation.

tests/psbt.rs
```diff
@@ -1,11 +1,909 @@
...
+        .manually_selected_only()
+        .sequence_override(op, Sequence(3)); // CSV requires >= 6
+    let result = wallet.create_psbt(params);
+    assert!(matches!(result, Err(CreatePsbtError::Sequence(_))));
+}
+
+// Test that replacing two unconfirmed txs A, B results in a transaction
+// that spends the inputs of both A and B.
+#[test]
+fn test_replace_by_fee_and_recpients() {
```
notmandatory: nit ⛏️ 
```suggestion
fn test_replace_by_fee_and_recipients() {
```

src/psbt/params.rs
```diff
@@ -0,0 +1,706 @@
...
+impl fmt::Debug for UtxoFilter {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "UtxoFilter")
+    }
+}
+
+impl PsbtParams<Rbf> {
+    /// Replace spends of the provided `txs`. This will internally set the list of UTXOs
+    /// to be spent.
+    fn replace(&mut self, txs: &[Arc<Transaction>]) {
```
notmandatory: Why does the `txs` param need to be an `&[Arc<Transaction>]` ?  can it be `&[Transaction]` since the code below on 528 is cloning the inner ref and not the `Arc` so there is no additional efficiency.

Same comment for the callers `PsbtParams<>::replace_txs()` and `Wallet::replace_by_fee_and_recipients()`.

notmandatory (commented):
There's a lot of great work here, I mainly focused my review on the examples, APIs and confirmed no breaking changes, I only have a couple questions and a small nit to fix.  I also ran a couple different LLM models at it and they gave it high marks without finding any significant issues.

ValuedMammal:
Thanks @notmandatory

src/psbt/params.rs
```diff
@@ -0,0 +1,665 @@
...
+    /// ```
+    ///
+    /// [`Plan`]: miniscript::plan::Plan
+    /// [`psbt::Input`]: bitcoin::psbt::Input
+    pub fn add_planned_input(&mut self, input: Input) -> &mut Self {
+        if self.set.insert(input.prev_outpoint()) {
+            self.inputs.push(input);
+        }
+        self
+    }
```
oleonardolima: I got a bit confused on the usage of both `self.utxos` and `self.set`, shouldn't the `input` outpoint be addded to `self.utxos` here too ?

examples/psbt.rs
```diff
@@ -0,0 +1,120 @@
...
+    let feerate = feerate_unchecked(FEERATE);
+    params
+        .add_recipients([(addr, AMOUNT)])
+        .fee_rate(feerate)
+        .coin_selection(SingleRandomDraw);
+
+    // Create PSBT (which also returns the Finalizer).
+    let (mut psbt, finalizer) = wallet.create_psbt(params)?;
+
+    dbg!(&psbt);
```
110CodingP: nit: is this intentional?

src/wallet/mod.rs
```diff
@@ -2491,6 +2527,36 @@ impl Wallet {
...
+    /// If the transaction was newly inserted and an output matches a derived script pubkey, then
+    /// the index is updated with the relevant outpoints. If no outputs are relevant, the
+    /// transaction is kept and the index remains unchanged. If `tx` already exists in the wallet
+    /// under the same txid, then the effect is a no-op.
+    ///
+    /// **You must persist the change set staged as a result of this call.**
+    ///
+    /// [relevance]: Indexer::is_tx_relevant
+    /// [indexer]: Self::spk_index
+    pub fn insert_tx<T>(&mut self, tx: T)
```
oleonardolima: Is this a MUST to go together with `create_psbt` ? I'd move this to another PR instead.

Does it also need a matching (or existing) `WalletEvent` (by using apply_update_events) ?

Also, having the `.expect()` in a public API feels a bit odd.

tests/psbt.rs
```diff
@@ -1,11 +1,909 @@
...
+    insert_tx(&mut wallet, unconfirmed_tx.clone());
+
+    // Now confirm that tx.
+    let confirmed_txid = unconfirmed_tx.compute_txid();
+    let confirm_block = BlockId {
+        height: 1001,
+        hash: Hash::hash(b"1001"),
+    };
+    insert_tx_anchor(&mut wallet, unconfirmed_tx.clone(), confirm_block);
+    insert_checkpoint(&mut wallet, confirm_block);
```
110CodingP: nit: we don't need this right? since `insert_tx_anchor` calls `insert_checkpoint`.

src/wallet/mod.rs
```diff
@@ -2778,6 +2844,779 @@ impl Wallet {
...
+                    longterm_feerate,
+                    max_rounds,
+                } => {
+                    selector
+                        .select_with_algorithm(selection_algorithm_lowest_fee_bnb(
+                            longterm_feerate,
+                            max_rounds,
+                        ))
+                        .map_err(CreatePsbtError::Bnb)?;
+                }
```
oleonardolima, line 3158: question: the user does not have the option of using a customized CS algorithm anymore ?

src/wallet/mod.rs
```diff
@@ -2778,6 +2844,779 @@ impl Wallet {
...
 
+/// Maps a chain position to tx confirmation status, if `pos` is the confirmed
+/// variant.
+///
+/// - Returns None if the confirmation height or time is not a valid absolute [`Height`] or
+///   [`Time`].
+///
+/// [`Height`]: bitcoin::absolute::Height
+/// [`Time`]: bitcoin::absolute::Time
+fn status_from_position(pos: ChainPosition<ConfirmationBlockTime>) -> Option<ConfirmationStatus> {
```
oleonardolima, line 2815: nit: `confirmation_status()` would be just fine too, also it could live in `utils.rs` instead

src/wallet/mod.rs
```diff
@@ -789,6 +802,19 @@ impl Wallet {
             .map(|((k, i), full_txo)| new_local_utxo(k, i, full_txo))
     }
 
+    /// List indexed [`FullTxOut`]s.
+    fn list_indexed_txouts(
```
oleonardolima, line 806: nit: I'd suggest renaming it to `list_txouts` and reuse it in the already existing `list_output` method.

src/wallet/tx_builder.rs
nymius, line 1: What's the feature unlocked by this change?

src/psbt/params.rs
```diff
@@ -0,0 +1,706 @@
...
+#[derive(Debug)]
+pub struct ReplaceTx;
+
+/// Alias for [`ReplaceTx`] context marker.
+pub type Rbf = ReplaceTx;
+
+/// Parameters to create a PSBT.
+// TODO: Can we derive `Clone` for this?
+#[derive(Debug)]
+pub struct PsbtParams<C> {
```
nymius, line 32: Looking in perspective to the motivation for `bdk-tx`, why this is better than `TxBuilder`? What's the difference?

src/psbt/params.rs
nymius: Why we cannot create the `PsbtParams` here, i.e., `fn replace(txs: &[Arc<Transaction>]) -> PsbtParams<Rbf>`? Is there any parameter that we would like to keep from whatever we had before? I've checked the example and we are not reusing parameters from `psbt1`, we should?

nymius: `test_replace_params` repeats the pattern, it doesn't reuse the old `PsbtParams`.

tests/psbt.rs
```diff
@@ -1,11 +1,909 @@
...
+    assert!(matches!(
+        result,
+        Err(CreatePsbtError::InsufficientFunds(
+            bdk_coin_select::InsufficientFunds { missing: 10_000 }
+        )),
+    ));
+}
+
+#[test]
+fn test_create_psbt_maturity_height() {
```
nymius, line 124: A test with no assertions? Shouldn't we check at least the outpoint used belongs to the coinbase tx?

tests/add_foreign_utxo.rs
```diff
@@ -323,3 +323,56 @@ fn test_add_foreign_utxo_rejects_wrong_non_witness_utxo_even_with_witness_utxo()
         "should reject non_witness_utxo with wrong txid even when witness_utxo is present"
     );
 }
+
+#[test]
+fn test_add_planned_psbt_input() -> anyhow::Result<()> {
```
nymius, line 328: Is `anyhow::Result<()>` really needed here? It is first introduced by this test. Why not replacing the `?` operators by `.expect("...")`?

nymius (commented):
I need another round of review yet, but these are the things I've found so far.

src/psbt/params.rs
nymius: +1

src/psbt/params.rs
```diff
@@ -0,0 +1,706 @@
...
+    /// Set of selected UTXO outpoints.
+    pub(crate) set: HashSet<OutPoint>,
+    /// List of UTXO outpoints to spend.
+    pub(crate) utxos: Vec<OutPoint>,
+    /// List of planned transaction [`Input`]s.
```
nymius, line 37: Why we need both?

tests/psbt.rs
```diff
@@ -1,11 +1,909 @@
...
+        let (psbt, _) = wallet.create_psbt(params).unwrap();
+        assert_eq!(psbt.unsigned_tx.lock_time.to_consensus_u32(), 200_000);
+    }
+}
+
+#[test]
+fn test_create_psbt_cltv_timestamp() {
+    use absolute::LockTime;
+
+    let lock_time = LockTime::from_consensus(1734230218);
```
nymius, line 264: Here, I would use a constant relative to the unix timepstamp threshold, e.g., `LOCKTIME_TIMESTAMP_FLOOR = 500 millions` then `LockTime::from_consensus(LOCKTIME_TIMESTAMP_FLOOR + 1)`

tests/psbt.rs
```diff
@@ -1,11 +1,909 @@
...
+            .add_utxos(&[op])
+            .add_assets(Assets::new().after(lock_time))
+            .add_recipients([(addr.script_pubkey(), Amount::from_btc(0.42).unwrap())]);
+        let (psbt, _) = wallet.create_psbt(params).unwrap();
+        assert_eq!(psbt.unsigned_tx.lock_time, lock_time);
+    }
+
+    // Locktime greater than required
+    {
+        let new_lock_time = 1772167108;
```
nymius, line 302: Same here than above comment.

src/psbt/params.rs
110CodingP: If I understand correctly then, `utxos` corresponds to local UTXOs and `inputs` corresponds to foreign UTXOs in the old TxBuilder and `set` is a hashset of all manually selected utxos (local + foreign) for quick membership checks.

But then do we plan to provide the old invariants? By old invariants I mean,
"Foreign UTXOs are not prioritized over local UTXOs. If a local UTXO is added to the manually selected list, it will replace any conflicting foreign UTXOs. However, a foreign UTXO cannot replace a conflicting local UTXO."

src/psbt/params.rs
```diff
@@ -0,0 +1,706 @@
...
+        for tx_node in tx_graph.full_txs() {
+            let tx = &tx_node.tx;
+            if tx.is_coinbase()
+                || tx_graph
+                    .walk_ancestors(Arc::clone(tx), |_, tx| Some(tx.compute_txid()))
+                    .any(|ancestor_txid| txids_to_replace.contains(&ancestor_txid))
+            {
+                txids_to_replace.remove(&tx_node.txid);
+            } else {
+                utxos.extend(tx.input.iter().map(|txin| txin.previous_output));
```
110CodingP, line 573: This also adds the inputs which were foreign UTXOs to `utxos`  field?
Hacked together a test to check this.
```rust
#[test]
fn test_replace_tx_with_planned_input() {
    use KeychainKind::*;

    let (desc, change_desc) = get_test_wpkh_and_change_desc();
    let mut wallet = Wallet::create(desc, change_desc)
        .network(Network::Regtest)
        .create_wallet_no_persist()
        .unwrap();

    let addr = wallet.reveal_next_address(External).address;

    // Fund the wallet with an unconfirmed output.
    let funding_tx = Transaction {
        input: vec![TxIn {
            previous_output: OutPoint::new(Hash::hash(b"funding_parent"), 0),
            ..Default::default()
        }],
        output: vec![TxOut {
            value: Amount::from_sat(200_000),
            script_pubkey: addr.script_pubkey(),
        }],
        ..new_tx(0)
    };
    let funding_op = OutPoint::new(funding_tx.compute_txid(), 0);
    insert_tx(&mut wallet, funding_tx);

    // Create an unconfirmed tx spending the funded UTXO.
    let recip =
        ScriptBuf::from_hex("5120e8f5c4dc2f5d6a7595e7b108cb063da9c7550312da1e22875d78b9db62b59cd5")
            .unwrap();
    let op2 = OutPoint::new(Hash::hash(b"txid"), 2);
    let txout = TxOut {
        value: Amount::ZERO,
        script_pubkey: ScriptBuf::new_p2a(),
    };
    let psbt_input = bitcoin::psbt::Input {
        witness_utxo: Some(txout),
        ..Default::default()
    };
    let planned_input = bdk_tx::Input::from_psbt_input(
        op2,
        Sequence::ENABLE_LOCKTIME_NO_RBF,
        psbt_input,
        /* satisfaction_weight: */ 0,
        /* status: */ None,
        /* is_coinbase: */ false,
        /* absolute_timelock: */ None,
    ).unwrap();

    let mut params = PsbtParams::default();
    params
        .add_utxos(&[funding_op])
        .add_planned_input(planned_input.clone())
        .add_recipients([(recip, Amount::from_sat(100_000))]);
    let unconfirmed_tx = wallet.create_psbt(params).unwrap().0.unsigned_tx;
    insert_tx(&mut wallet, unconfirmed_tx.clone());
    let params = PsbtParams::default().replace_txs(&[Arc::new(unconfirmed_tx)]);

    // this fails
    let psbt = wallet.replace_by_fee(params).unwrap().0;
}
```

src/psbt/params.rs
```diff
@@ -0,0 +1,706 @@
...
+        write!(f, "UtxoFilter")
+    }
+}
+
+impl PsbtParams<Rbf> {
+    /// Replace spends of the provided `txs`. This will internally set the list of UTXOs
+    /// to be spent.
+    fn replace(&mut self, txs: &[Arc<Transaction>]) {
+        self.utxos.clear();
+        self.set.clear();
```
110CodingP: Why do we not clear the `inputs`? 

Also clearing the `set` can allow duplicate entries in `inputs`.
```rust
let (desc, change_desc) = get_test_wpkh_and_change_desc();
    let mut wallet = Wallet::create(desc, change_desc)
        .network(Network::Regtest)
        .create_wallet_no_persist()
        .unwrap();

    let addr = wallet.reveal_next_address(External).address;

    // Fund the wallet with an unconfirmed output.
    let funding_tx = Transaction {
        input: vec![TxIn {
            previous_output: OutPoint::new(Hash::hash(b"funding_parent"), 0),
            ..Default::default()
        }],
        output: vec![TxOut {
            value: Amount::from_sat(200_000),
            script_pubkey: addr.script_pubkey(),
        }],
        ..new_tx(0)
    };
    let funding_op = OutPoint::new(funding_tx.compute_txid(), 0);
    insert_tx(&mut wallet, funding_tx);

    // Create an unconfirmed tx spending the funded UTXO.
    let recip =
        ScriptBuf::from_hex("5120e8f5c4dc2f5d6a7595e7b108cb063da9c7550312da1e22875d78b9db62b59cd5")
            .unwrap();
    let op2 = OutPoint::new(Hash::hash(b"txid"), 2);
    let txout = TxOut {
        value: Amount::ZERO,
        script_pubkey: ScriptBuf::new_p2a(),
    };
    let psbt_input = bitcoin::psbt::Input {
        witness_utxo: Some(txout),
        ..Default::default()
    };
    let planned_input = bdk_tx::Input::from_psbt_input(
        op2,
        Sequence::ENABLE_LOCKTIME_NO_RBF,
        psbt_input,
        /* satisfaction_weight: */ 0,
        /* status: */ None,
        /* is_coinbase: */ false,
        /* absolute_timelock: */ None,
    ).unwrap();

    let mut params = PsbtParams::default();
    params
        .add_utxos(&[funding_op])
        .add_recipients([(recip, Amount::from_sat(100_000))]);
    let unconfirmed_tx = wallet.create_psbt(params).unwrap().0.unsigned_tx;
    insert_tx(&mut wallet, unconfirmed_tx.clone());
    let mut params = PsbtParams::default();
    params.add_planned_input(planned_input.clone());
    let mut params = params.replace_txs(&[Arc::new(unconfirmed_tx)]);
    params.add_planned_input(planned_input.clone());

    dbg!(&params); // contains duplicate `Input`s
```

src/psbt/params.rs
```diff
@@ -0,0 +1,706 @@
...
+    ///     is_coinbase,
+    ///     None,
+    /// )?;
+    /// params.add_planned_input(input);
+    /// # Ok::<_, anyhow::Error>(())
+    /// ```
+    ///
+    /// [`Plan`]: miniscript::plan::Plan
+    /// [`psbt::Input`]: bitcoin::psbt::Input
+    pub fn add_planned_input(&mut self, input: Input) -> &mut Self {
```
110CodingP: Why are we allowed to add a planned_input (using `add_planned_input`) but not a utxo (using `add_utxos`) to a replacement transaction? Is this because we cannot recover the `planned_input`s from the transactions to be replaced?

tests/psbt.rs
```diff
@@ -1,11 +1,909 @@
...
+    };
+    insert_checkpoint(&mut wallet, anchor.block_id);
+    let op = receive_output(
+        &mut wallet,
+        Amount::from_sat(100_000),
+        ReceiveTo::Block(anchor),
+    );
+
+    let addr = wallet.next_unused_address(KeychainKind::External);
+    let rel_locktime = relative::LockTime::from_consensus(6).unwrap();
```
nymius, line 425: I would use a different value here, as 6 is the default if nothing set, as [line 381 above](https://github.com/ValuedMammal/bdk_wallet/blob/43adb6ea46ebf87fdfc9e5ff078815e8a5d5edc5/tests/psbt.rs#L381) shows

tests/psbt.rs
```diff
@@ -1,11 +1,909 @@
...
+            },
+            confirmation_time: 1234567000,
+        };
+        insert_checkpoint(&mut wallet, anchor.block_id);
+        let mut params = PsbtParams::default();
+        params
+            .add_utxos(&[op])
+            .add_recipients([(addr.script_pubkey(), Amount::from_btc(0.42).unwrap())]);
+        let (psbt, _) = wallet.create_psbt(params).unwrap();
+        assert_eq!(psbt.unsigned_tx.input[0].sequence, Sequence(6));
```
nymius, line 380: I would use a local named constant to point to the place where this value is set.

tests/psbt.rs
```diff
@@ -1,11 +1,909 @@
...
+        }],
+        ..new_tx(0)
+    };
+    let fee_c = wallet.calculate_fee(&tx_c).unwrap();
+    insert_tx(&mut wallet, tx_c.clone());
+
+    // The replacement must pay at least the combined fee of all three transactions
+    // (Bitcoin Core RBF Rule 3).
+    let total_original_fee = fee_a + fee_b + fee_c;
+    assert_eq!(total_original_fee.to_sat(), 90_000);
```
nymius, line 716: Why not checking `replacement_fee >= fee_a + fee_b + fee_c` instead of checking they equal `90_000`? Is there any value on doing this? Could the fee be different than `90_000` after how we have set up the txs above?

tests/psbt.rs
```diff
@@ -1,11 +1,909 @@
...
+    let change_script = ChangeScript::from_descriptor(
+        wallet
+            .public_descriptor(KeychainKind::Internal)
+            .at_derivation_index(0)
+            .unwrap(),
+    );
+    params.change_script(change_script);
+    params.drain_wallet();
+    let (psbt, _) = wallet.create_psbt(params).unwrap();
+    assert_eq!(psbt.unsigned_tx.input.len(), 2);
```
nymius, line 1012: We can check the inputs used are `600` and `1000` sats each.

src/wallet/mod.rs
```diff
@@ -2778,6 +2844,779 @@ impl Wallet {
...
+        fee_rate: FeeRate,
+        recipients: Vec<(ScriptBuf, Amount)>,
+    ) -> Result<(Psbt, Finalizer), ReplaceByFeeError> {
+        let params = PsbtParams {
+            fee_rate,
+            recipients,
+            ..Default::default()
+        }
+        .replace_txs(txs);
+        self.replace_by_fee_with_rng(params, &mut rand::thread_rng())
```
nymius, line 3267: Is this really an ergonomic improvement?

tests/psbt.rs
```diff
@@ -1,11 +1,909 @@
...
+    let confirmed_txid = unconfirmed_tx.compute_txid();
+    let confirm_block = BlockId {
+        height: 1001,
+        hash: Hash::hash(b"1001"),
+    };
+    insert_tx_anchor(&mut wallet, unconfirmed_tx.clone(), confirm_block);
+    insert_checkpoint(&mut wallet, confirm_block);
+
+    // Attempting to replace the now-confirmed tx should return TransactionConfirmed.
+    let result = wallet.replace_by_fee_and_recipients(
```
nymius, line 788: Why not reusing the old params here?

`params.replace(&[Arc::new(unconfirmed_tx)]).fee_rate(fee_rate)`?

This goes in hand with https://github.com/bitcoindevkit/bdk_wallet/pull/297/changes#r3383792188.

tests/wallet.rs
```diff
@@ -25,6 +25,89 @@ use rand::SeedableRng;
 
 mod common;
 
+// Test we can select and spend an indexed but not-yet-canonical utxo
+#[test]
+fn test_spend_non_canonical_txout() -> anyhow::Result<()> {
```
nymius: I disagree with this, where it's said that unconfirmed txs are non canonical? I would consider this different if there existed a conflicting tx, or even a confirmed conflicting tx, but in this case, the tests is checking unconfirmed tx2 and parents belong to the canonical view.

src/psbt/params.rs
notmandatory: The two advantages I see are: 
1. `bdk-tx` uses the `rust-miniscript` planning module instead of the hard to use `Policy` approach in TxBuilder.
2. By extracting `bdk-tx` into a new crate it can be used by any projects using `rust-bitcoin` and `rust-miniscript`.

nymius: Thanks Steve, In the change to `bdk-tx`, I initially assigned the same weight to the "code complexity" and lack of "flexibility" issues than to `Policy`, I might have been shortsighted.
Rephrasing my question: how this approach is not going to lead us to the same level of complexity of the old `TxBuilder`? Or have we assumed the complexity is insurmountable? (that's a valid answer too).

ValuedMammal:
Thanks @110CodingP 

Moved to 3.2.0 as it's still under testing.

src/psbt/params.rs
```diff
@@ -0,0 +1,843 @@
...
+    pub fn only_witness_utxo(&mut self) -> &mut Self {
+        self.only_witness_utxo = true;
+        self
+    }
+
+    /// Drain wallet.
+    ///
+    /// This will force selection of the available input candidates. As such, the option is only
+    /// applied to inputs that meet the spending criteria.
+    pub fn drain_wallet(&mut self) -> &mut Self {
```
ValuedMammal, line 430: Perhaps `select_all` would be a better name for describing what it actually does.

src/psbt/params.rs
evanlinjin: @nymius I guess what you mean by "complexity" is "invalid state representation"?

src/psbt/params.rs
```diff
@@ -0,0 +1,843 @@
...
+    /// List of recipient script/amount pairs.
+    pub(crate) recipients: Vec<(ScriptBuf, Amount)>,
+    /// Optional script or descriptor designated for change.
+    pub(crate) change_script: Option<ChangeScript>,
+    /// Optional assets for creating a spend plan.
+    pub(crate) assets: Option<Assets>,
+    /// Target fee rate.
+    pub(crate) fee_rate: FeeRate,
+    /// Whether to spend all available coins.
+    pub(crate) drain_wallet: bool,
```
evanlinjin, line 48: Looking at how `drain_wallet` functions, I think `DrainAll` should be a `SelectionStrategy` (or as you suggested, `SelectAll`).

src/psbt/params.rs
```diff
@@ -0,0 +1,843 @@
...
+/// Parameters to create a PSBT.
+// TODO: Can we derive `Clone` for this?
+#[derive(Debug)]
+pub struct PsbtParams<C> {
+    /// Set of selected UTXO outpoints, `HashSet` ensures uniqueness
+    pub(crate) set: HashSet<OutPoint>,
+    /// List of UTXO outpoints to spend.
+    pub(crate) utxos: Vec<OutPoint>,
+    /// List of planned transaction [`Input`]s.
+    pub(crate) inputs: Vec<Input>,
```
evanlinjin, line 38: Instead of having both `utxos` (resolved by wallet) and `inputs` (external UTXOs), I think there should be an unified enum:

```rust
enum InputSpec {
    Owned(OutPoint),   // to be resolved by wallet
    Planned(Input),    // taken as-is
}
```

This will fix an ordering bug with `Untouched` - where the order does not respect the order the UTXOs were introduced if we have both wallet-resolved and external inputs.

src/psbt/params.rs
nymius: No, I meant code complexity: the numerous fields, the setters, the long `create_tx` method and all that.
Or the shape of the code didn't matter that much and the main concern were these invalid state representations.

evanlinjin (commented):
Also, what do you think about a staged API? 

```rust
// Stage 1 — wallet resolves a snapshot of spendable inputs
let coins = wallet.candidates()       // -> InputCandidates
    .with_assets(assets)
    .maturity_height(h)
    .filter(|txo| ...)
    .add_foreign(input)               // planned/foreign inputs
    .replace(txs)?;                   // RBF seeds forced inputs here

// Stage 2 — wallet builds the tx from that set
let (psbt, finalizer) = wallet.create_psbt(coins, CreateTxParams)?;
let (psbt, finalizer) = wallet.create_rbf(coins, RbfParams)?;
```

Cargo.toml
```diff
@@ -21,6 +21,7 @@ unexpected_cfgs = { level = "warn", check-cfg = ['cfg(coverage_nightly)'] }
 
 [dependencies]
 bdk_chain = { version = "0.23.3", features = ["miniscript", "serde"], default-features = false }
+bdk_tx = { version = "0.2.0", default-features = false }
```
evanlinjin, line 24: Not sure if this is related to this PR, but we should pin bitcoin-units to 0.1.3 or higher. Any lower, and `FeeRate::from_sat_per_vb_u32` would not be available.

src/psbt/params.rs
```diff
@@ -0,0 +1,843 @@
...
+        self
+    }
+}
+
+/// Coin select strategy.
+#[derive(Debug, Clone, Copy, Default)]
+#[non_exhaustive]
+pub enum SelectionStrategy {
+    /// Single random draw.
+    #[default]
```
evanlinjin, line 499: I would argue that `LowestFee` should be the default strategy - as it creates better results.

src/psbt/params.rs
```diff
@@ -0,0 +1,843 @@
...
+    ///
+    /// If an outpoint doesn't correspond to an indexed script pubkey, an [`UnknownUtxo`]
+    /// error will occur. See [`Wallet::create_psbt`] for more.
+    ///
+    /// To add a UTXO that did not originate from this wallet (i.e. a "foreign" UTXO), see
+    /// [`PsbtParams::add_planned_input`].
+    ///
+    /// [`UnknownUtxo`]: crate::wallet::error::CreatePsbtError::UnknownUtxo
+    /// [`Wallet::create_psbt`]: crate::Wallet::create_psbt
+    pub fn add_utxos(&mut self, outpoints: &[OutPoint]) -> &mut Self {
```
evanlinjin, line 137: I think that a consuming builder pattern has better ergonomics. The current (non-consuming) version, forces the caller to instantiate on a separate line and keep a mutable variable around.

```suggestion
    pub fn add_utxos(mut self, outpoints: &[OutPoint]) -> Self {
```

`PsbtParams` will then have `#[must_use]`.

evanlinjin: I guess a non-consuming builder is slightly better API for someone who needs conditional construction. I'll be happy either way.

src/wallet/mod.rs
```diff
@@ -2778,6 +2804,809 @@ impl Wallet {
...
+            .collect::<Result<_, _>>()?;
+
+        // Sum fees from all descendants known to the tx graph. This assumes every
+        // descendant is currently in the mempool, which could slightly overestimate
+        // the fee floor if a descendant was evicted or never relayed.
+        let descendant_fee: Amount = descendants
+            .iter()
+            .filter_map(|&txid| {
+                let tx = self.tx_graph.graph().get_tx(txid)?;
+                self.calculate_fee(&tx).ok()
```
evanlinjin, line 3474: If the fee cannot be calculated, we may undershoot the RBF fee floor. Maybe we should return an error?

src/psbt/params.rs
```diff
@@ -0,0 +1,843 @@
...
+    /// Refer to [`CanonicalizationParams`] for more.
+    pub fn canonicalization_params(
+        &mut self,
+        params: bdk_chain::CanonicalizationParams,
+    ) -> &mut Self {
+        self.canonical_params = params;
+        self
+    }
+
+    /// Set the [`Descriptor`] or raw [`Script`] to be used for generating the change output.
```
evanlinjin, line 372: We should probably note that if this is not explicitly specified, the wallet will still derive a change address.

evanlinjin:
Thanks for pushing this forward. The implementation is solid and surfaced a lot of the tricky cases. Before we commit to an API shape, I want to put an alternative on the table. I built a PoC of it in #502.

I don't think a single `create_psbt(PsbtParams)` is the right approach — it puts the complexity in the wrong place. `bdk_tx` already models transaction construction as a **staged flow** (resolve candidates → select coins → shape a `TxTemplate` → emit). That shape is deliberate and correct. A single-call wrapper collapses those stages back into one large params type, and that has three consequences worth avoiding:

1. **Invalid or conflicting options only surface at the final call.** A single params bag only sorts two things out when `create_psbt` finally runs: *conflicts* — an explicit `locktime` alongside `anti_fee_sniping_height`, both setting `nLockTime`, with the precedence hidden in docs — and *invalid values* — a sequence that violates an input's CSV requirement isn't rejected when you set it; it surfaces only at build time, failing the whole process. The staged flow applies operations in explicit order, and each one (e.g. `template.input_mut(op).set_sequence(..)`) validates and errors *at the call site*, leaving the rest of the template intact. It's also a better fit for **frontend / interactive** use: a UI can resolve candidates once and let the user adjust selection and outputs against a live, inspectable `CandidateSet`/`TxTemplate`, surfacing each error inline at the control that caused it instead of failing a whole form at submit.
2. **The wallet ends up re-declaring `bdk_tx`.** Version, locktime, anti-fee-sniping, ordering, shuffling, per-input/fallback sequence — these already exist on `bdk_tx`'s `TxTemplate` (this PR carries `sequence_override`/`fallback_sequence`, for instance). Re-exposing them as wallet options means we wrap and maintain each one, and the wallet's surface grows to mirror a dependency it should just be *using*. Every `bdk_tx` change becomes a `bdk_wallet` change.
3. **It only exposes what the wallet authors thought to wrap.** Hiding `bdk_tx` behind a single call means anything `bdk_tx` can do that `create_psbt` didn't surface is simply out of reach.

So `bdk_wallet`'s wrappers should **respect `bdk_tx`'s flow** rather than hide it: the wallet's real job is to make it easy to fill `bdk_tx`'s structs with what only the wallet knows (its UTXOs, descriptors/plans, change derivation, canonicalization), and then get out of the way. #502 does exactly that — `candidates()` → `select()` → `finish()`, returning a `bdk_tx::TxTemplate` you shape with `bdk_tx`'s own methods.

Concretely, because the candidate set and template are real values you hold, the staged design unlocks things a single opaque call can't: input grouping, post-resolution filtering, a direct `bdk_tx` escape hatch, and **resolve-once / select-many** — the last of which a large `bdk_wallet` user has explicitly asked for.

If anything, the staged design is *more* ergonomic — and more transparent. The single-call approach hands you one large params struct where it's hard to see how the options relate; staging scopes each set of options to its stage, so what's valid where is obvious. The one thing it gives up is the single-call one-liner for the trivial "send X to Y" case — and that convenience can always be layered *on top of* composable stages, whereas composability can't be retrofitted onto a sealed call. #502 is a draft PoC for this comparison; I'd love your eyes on it, and to hear where the staged shape falls short.

src/psbt/params.rs
evanlinjin: Note that some complexity in `create_tx` cannot be avoided right now. I'm assuming it can be simplified once `CanonicalView`, it's extract subgraph method and the `bdk_tx_core` glue lands.

src/wallet/mod.rs
oleonardolima: cc @notmandatory

evanlinjin:
Another objection I have is I don't think this integration should live in `bdk_wallet` while `bdk_tx` is still pre-stable. Putting `bdk_tx` types in `bdk_wallet`'s public API means every breaking `bdk_tx` release forces a `bdk_wallet` major release. This couples our stable crate to a fast-moving one.

I'd rather put it in a separate crate (say `bdk_wallet_tx`) that depends on both. Then neither base crate depends on the other: `bdk_wallet` stays stable, `bdk_tx` stays free to move, and the bridge crate absorbs the coupling - opt-in, and versioned on its own. This is the idea from bitcoindevkit/bdk-tx#12, but in its own crate rather than inside `bdk_tx` (which would invert the dependency).

https://github.com/bitcoindevkit/bdk-tx/pull/13 implemented https://github.com/bitcoindevkit/bdk-tx/issues/12 and got a Concept NACK without a rationale, so I couldn't address it. I'd like to understand the objection so we can weigh it properly.

evanlinjin:
There is a problem with the change-address lifecycle. If the caller decides to discard the created PSBT, they can't easily `unmark_used` on it as `create_psbt` only returns a PSBT.

The fix would be to return `Option<AddressInfo>` alongside the PSBT if a change address is derived.

src/wallet/error.rs
```diff
@@ -365,3 +366,112 @@ impl fmt::Display for BuildFeeBumpError {
...
+#[non_exhaustive]
+pub enum ReplaceByFeeError {
+    /// There was a problem creating the PSBT
+    CreatePsbt(CreatePsbtError),
+    /// Failed to compute the fee of an original transaction
+    PreviousFee(bdk_chain::tx_graph::CalculateFeeError),
+    /// Original transaction could not be found
+    MissingTransaction(Txid),
+    /// One of the transactions to be replaced is already confirmed
+    TransactionConfirmed(Txid),
```
evanlinjin, line 434: I don't think this error variant earns it's keep.

* Most callers would want to check whether the transaction is still unconfirmed before trying to replace it. A caller that doesn't do that obviously wants to create a transaction that conflicts with a confirmed one.

* The guard is partial. It only catches "already confirmed at build time". The original transaction can be confirmed before the replacement is ready to broadcast.

src/wallet/mod.rs
notmandatory: I would prefer setting the coin selection to any struct that implements `CoinSelectionAlgorithm` so that a custom one can be used. That's how `TxBuilder` currently does it and I don't see a compelling reason to change it.

ValuedMammal:
Closed in favor of #516 as the thread became too long.

PR #306, opened 2025-09-04
Add docs + test on required use of xpub for multipath constructor

thunderbiscuit:
This PR makes explicit a limitation of the `Wallet::create_from_two_path_descriptor` method, in that you must use public descriptors to build the wallet using this constructor, meaning you can only build watch-only wallets.

I also added a test to showcase this and the error returned.

See [my comment here](https://github.com/bitcoindevkit/bdk_wallet/pull/275#issuecomment-3251334629) for more:

From what I can tell the issue is that miniscript cannot parse an xprv (I saw a comment from Sanket about this somewhere but now I can't find it anymore) if the multipath contains hardened derivation paths, and therefore miniscipt cannot ensure the xprv will be parsable into public keys in all cases? Something like that. But our arguments for bdk wallet constructors require the `IntoWalletDescriptor` trait:

```rust
pub trait IntoWalletDescriptor {
    /// Convert to wallet descriptor
    fn into_wallet_descriptor(
        self,
        secp: &SecpCtx,
        network: Network,
    ) -> Result<(ExtendedDescriptor, KeyMap), DescriptorError>;
}

pub type ExtendedDescriptor = Descriptor<DescriptorPublicKey>;
```

In other words, upon creation, the wallet will always try to parse the descriptor string into its public form, and here since it cannot be done by miniscript we fail to build.

### Changelog notice

No changelog notice.

#### All Submissions:

### DIFF
--- a/wallet/src/wallet/mod.rs
+++ b/wallet/src/wallet/mod.rs
@@ -399,7 +399,8 @@ impl Wallet {
     /// Build a new [`Wallet`] from a two-path descriptor.
     ///
     /// This function parses a multipath descriptor with exactly 2 paths and creates a wallet
-    /// using the existing receive and change wallet creation logic.
+    /// using the existing receive and change wallet creation logic. Note that you can only use this
+    /// method with public extended keys (`xpub` prefix) to create watch-only wallets.
     ///
     /// Multipath descriptors follow [BIP 389] and allow defining both receive and change
     /// derivation paths in a single descriptor using the `<0;1>` syntax.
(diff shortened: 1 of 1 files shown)


ValuedMammal:
ACK e647ed2692cfb9234801d72a88f690dbd73d5777

Further, the presence of hardened steps doesn't seem to make a difference, as [this test](https://github.com/bitcoindevkit/bdk_wallet/blob/0cf9f280abad973ebe0efa7394d2096bbedcf520/wallet/src/descriptor/mod.rs#L962-L968) shows. It's important to document the limitation as this PR does but to be honest I haven't found a good explanation for why it exists.

ValuedMammal (approved):

oleonardolima (approved):
utACK e647ed2692cfb9234801d72a88f690dbd73d5777

PR #317, opened 2025-09-24
chore: bump MSRV to 1.85.0

### LINKED ISSUE #289: Upgrade MSRV to 1.85
We should update the `bdk_wallet` MSRV to 1.85 with rust edition 2024.  The rust 1.85 version is included in the next Debian stable release 13 (trixie) which should be released this week.  

**Use case**  

Upgrading the MSRV will resolve #101. 

To support enterprise users I think we should continue to support as an MSRV what ever is released with Debian stable (see #203). 

**Additional context**

Changing our MSRV is also a good time to implement the suggestion in #79.

https://lists.debian.org/debian-devel-announce/2025/07/msg00003.html
https://packages.debian.org/search?keywords=rust

ValuedMammal:
Bump `bdk_wallet` MSRV to 1.85.0.

Restructure repository to be a single rust project.

Small cleanups to `.github/workflows` by moving away from unmaintained `actions-rs/toolchain` in favor of `dtolnay/rust-toolchain`.

Follow-ups

- [x] Address zizmor code scanning results
- [ ] Use rust edition 2024. May have to remove some "unnecessary binding modifiers"
- [ ] Consider updating `.github/CODEOWNERS`

fix #289 

### Changelog notice

```md
Changed

- Raise project MSRV to 1.85.0
- ci: Migrate to `dtolnay/rust-toolchain`
- ci: Add zizmor.yml config
```

#### All Submissions:

### DIFF
--- a/.github/zizmor.yml
+++ b/.github/zizmor.yml
@@ -1,0 +1,7 @@
+# Zizmor config
+rules:
+  unpinned-uses:
+    config:
+      policies:
+        # Allow pin by ref/tag
+        dtolnay/rust-toolchain: ref-pin


Cargo.toml
```diff
@@ -2,11 +2,8 @@
 resolver = "2"
 members = [
     "wallet",
-    "examples/example_wallet_electrum",
-    "examples/example_wallet_esplora_blocking",
-    "examples/example_wallet_esplora_async",
-    "examples/example_wallet_rpc",
```
thunderbiscuit, line 8: Because you now only have one crate in this workspace, I wonder if it's cleaner to remove this workspace entirely. As far as I understand, single crates typically don't use workspaces.

You'd have a `src/` directory right at the root and could remove the `wallet/` directory. Your cargo commands would work directly from the root of the repository.

thunderbiscuit (commented):
Concept ACK. See comment on workspace.

Cargo.toml
ValuedMammal: If you want I can try to do this in the same PR.

thunderbiscuit (commented):
I'm personally not a fan of the "pin the hash on the actions" that zizmor recommends here.

If you want to address the zizmor warning you're getting here, you could [do what we do for bdk-ffi](https://github.com/bitcoindevkit/bdk-ffi/blob/master/.github/zizmor.yml). You add a file called zizmor.yml at the root of the `.github/` directory with the following:

```yaml
# This is a configuration file for the zizmor action defined in .github/workflows/zizmor.yaml

rules:
  # Disable the rule requiring all actions be pinned to a specific hash
  unpinned-uses:
    config:
      policies:
        "*": any
```

thunderbiscuit (approved):
Tested ACK 6260274a56c9442362378df7f112c6410121b4a2.

I'm getting one little clippy warning locally:

```sh
error: the following explicit lifetimes could be elided: 'a
    --> src/wallet/mod.rs:1192:25
     |
1192 |     pub fn transactions<'a>(&'a self) -> impl Iterator<Item = WalletTx<'a>> + 'a {
     |                         ^^   ^^                                        ^^     ^^
     |
     = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#needless_lifetimes
     = note: `-D clippy::needless-lifetimes` implied by `-D warnings`
     = help: to override `-D warnings` add `#[allow(clippy::needless_lifetimes)]`
help: elide the lifetimes
     |
1192 -     pub fn transactions<'a>(&'a self) -> impl Iterator<Item = WalletTx<'a>> + 'a {
1192 +     pub fn transactions(&self) -> impl Iterator<Item = WalletTx<'_>> + '_ {
```

But this is related to my default toolchain being 1.85. The warning disappears if I build using 1.89. See #315 for my recommendation on how to fix that sort of issue. Not required for this PR, just pointing it out.

README.md
```diff
@@ -1,21 +1,19 @@
-# The Bitcoin Dev Kit Wallet
-
 <div align="center">
-  <h1>BDK Wallet</h1>
+  <h1>BDK</h1>
```
ValuedMammal, line 2: 📌 Accidentally clobbered when I merged the two READMEs together. Same on line 4.

Issue #227, opened 2025-05-05
Support for Any Number of Descriptors/Keychains


thunderbiscuit:
This issue is the starting point for discussion on the idea of adding support for any number of keychains in the `Wallet` type in a future major/breaking release.

I have an PR (#226) digging into what such an API _could_ look like, but it's really not the only way to do it and I invite others to also poke into this if they're interested.

### Pros and Cons

✅ The wallet does away with the idea of main/change (external/internal) descriptors, which are fairly narrow-focused and less generally-applicable to all situations and requirements.
✅ Counter-point to the above: Our wallet methods are currently sometimes not explicit enough, and produce unexpected results (for example calling `Wallet::peek_address` with KeychainKind::Internal will actually return an address on your External keychain if your wallet doesn't have an Internal keychain).
✅ Feature-parity for the Wallet type with application software in the industry that offers this feature. Examples: Blockstream Green, Sparrow Wallet, Blue Wallet, Samourai. All of those can provide a unified balance for the user over a number of descriptors/keychains.
❌ This potentially adds complexity to all wallet methods that deal with the keychains, because users need to handle the multiple keychains more explicitly. I'm not actually sure this is necessarily the case for all methods and under all API designs, just pointing it out here. More keychains = maybe more complexity at the call sites.
❌ Because the workflow of external/internal descriptors is embedded deep into the current Wallet API, the rewrite and review cost for shipping this is likely fairly high. The good thing is (at least so far in my exploration in #226), the reach of the changes is really limited to the wallet module. The separation with bdk_chain structures is really well done, and the inner logic is really built to handle this use case and so work out of the box really well (as far as I can tell). The real changes are mostly all at the Wallet and user-facing API level (not so much at the domain-logic level).

### Some questions that still need answering

1. Who needs this? How would they use it?
2. New changesets and persistence. How problematic is that?

evanlinjin:
The main concern I have is with the changeset design — this might be partly my oversight. I originally designed `bdk_wallet::ChangeSet` under the assumption that we wouldn't introduce flexible descriptor counts later on.

One question: for users who need this flexible descriptor count feature, do they also require backward compatibility with the existing 2-descriptor-only `Wallet`?

thunderbiscuit:
[...]

PR #318, opened 2025-09-24
Support For Arbitrary Number (1+) of Keychains

thunderbiscuit:
# Description

This PR allows the `Wallet` type to track any number of keychains. It is a breaking change.

Related Issues: #188, #227
Related PRs: #230, #226
Exploratory Crate: [multi-keychain-wallet](https://github.com/ValuedMammal/multi-keychain-wallet)

Follow along the todos and PR development on [this HackMD file](https://hackmd.io/@bdk/318).

<br>

# Commits

| Commit | Changes |
| ------ | ------ |
| 1 | Comment out `Wallet` dependent code. |
| 2 | Add `KeyRing` and related types. |
| 3 | Modify `Wallet` to levergage the `KeyRing`. |
| 4 | Add docs and examples. |
| 5 | Add tests. |

<br>

# Todos

Search for the following pattern to find Todos in the codebase:

```rust
// TODO PR #318: We should fix this because...
```

<br>

# Changelog notice

TODO


thunderbiscuit (commented):
Quick review on 6c50f002f1ffaaea58db25fc65bf5b189ec4896f: given your commit message I thought you were adding the Wallet::new constructor here, but the commit only contains the addition of the KeyRing to the Wallet type. Mind you this might be enough for this commit. Feel free to add a new commit with the constructor, or add it to this commit; whatever you think works best.

Also note that fa984e6c0ea6d59ff68bc5dc8d984bea8de190ea and 6c50f002f1ffaaea58db25fc65bf5b189ec4896f don't compile because of 

```rust
impl AsRef<bdk_chain::tx_graph::TxGraph<ConfirmationBlockTime>> for Wallet {
    fn as_ref(&self) -> &bdk_chain::tx_graph::TxGraph<ConfirmationBlockTime> {
        self.keychain_tx_graph.graph()
    }
}
```

on line 2638 of wallet/mod.rs. You could comment it out with the rest to ensure everything builds.

wallet/src/wallet/changeset.rs
```diff
@@ -101,58 +103,47 @@ type IndexedTxGraphChangeSet =
...
-    pub change_descriptor: Option<Descriptor<DescriptorPublicKey>>,
-    /// Stores the network type of the transaction data.
-    pub network: Option<bitcoin::Network>,
+#[derive(Debug, Clone, PartialEq, Deserialize, Serialize)]
+pub struct ChangeSet<K: Ord> {
+    /// Stores the `KeyRing` containing the network and descriptor data.
+    pub keyring: keyring::changeset::ChangeSet<K>,
     /// Changes to the [`LocalChain`](local_chain::LocalChain).
-    pub local_chain: local_chain::ChangeSet,
+    pub chain: local_chain::ChangeSet,
```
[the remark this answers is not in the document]
110CodingP: I did the rename since the corresponding field in `Wallet` is so. reverted the rename in [588121b](https://github.com/bitcoindevkit/bdk_wallet/pull/318/commits/588121b2f269c8bf9a4a6269b7170644e786440f).

110CodingP:
I am so sorry :cry: . I completely messed dividing the commits into two. Things should be fixed now!

thunderbiscuit:
This one now needs a big ol' rebase @110CodingP.

110CodingP:
Wow! this one was HUGE! For the first few commits I did something like `git rebase --onto master HEAD~7 HEAD~6` and for the later ones I cherry-picked each commit onto master and finally did a `reset --hard`, wondering if there's a simpler way to do difficult rebases :sweat_smile: ?

thunderbiscuit:
Adding an example like this fails:

```rust
    // Simple KeyRing, allowing us to build a standard 2-descriptor wallet.
    let external_descriptor: &str = "tr(tpubD6NzVbkrYhZ4WyC5VZLuSJQ14uwfUbus7oAFurAFkZA5N3groeQqtW65m8pG1TT1arPpfWu9RbBsc5rSBncrX2d84BAwJJHQfaRjnMCQwuT/86h/1h/0h/0/*)";
    let internal_descriptor: &str = "tr(tpubD6NzVbkrYhZ4WyC5VZLuSJQ14uwfUbus7oAFurAFkZA5N3groeQqtW65m8pG1TT1arPpfWu9RbBsc5rSBncrX2d84BAwJJHQfaRjnMCQwuT/86h/1h/0h/1/*)";

    let mut keyring: KeyRing<KeychainKind> = KeyRing::new(Network::Regtest, KeychainKind::External, external_descriptor);
    keyring.add_descriptor(KeychainKind::Internal, internal_descriptor, false);

    let mut wallet = Wallet::new(keyring);
```

With a stacktrace containing:
```sh
thread 'main' panicked at /Users/user/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/miniscript-12.3.5/src/descriptor/key.rs:687:14:
The key should not contain any wildcards at this point
```

After a bit of poking around I found that this is coming from the descriptor_id() method called inside insert_descriptor():

```rust
    pub fn insert_descriptor(
        &mut self,
        keychain: K,
        descriptor: Descriptor<DescriptorPublicKey>,
    ) -> Result<bool, InsertDescriptorError<K>> {
        let did = descriptor.descriptor_id();
```

I don't have time to finish the investigation today, but just pointing it out. I'm not sure yet why this worked well in the multi-keychain-wallet crate and not here, nor why the code for the constructor is so different than the other previous constructors like `Wallet::create_with_params`.

thunderbiscuit:
Total facepalm. Sorry @110CodingP! Ok so I simplified the example a little bit just to keep a really neat version of how to build a wallet as close as you can from the 2.0 approach.

110CodingP:
Also is `balance_per_keychain` api desirable? ig it would help with this [bdk#57](https://github.com/bitcoindevkit/bdk/issues/57#issue-700190681).

luisschwab:
I'm thinking about how this will interact with the wallet's birthday, since different keychains can possibly have different birthdays (e.g. adding a keychain with older transactions to a wallet with a keychain that has newer transactions).

110CodingP:
> I noticed you have the sync examples commented out for now. When making synching a wallet, are you planning to add all revealed spk's into an "spk pool" and sync everything, or are we going to have to specify which keychain we want to sync? 

Currently we sync all keychains in [one go](https://github.com/thunderbiscuit/bdk_wallet/blob/25a41f33a01f2dd46c24fddeb40822db79effe41/src/wallet/mod.rs#L791) . But ig having an API to sync a particular keychain would also be great in case of large wallets with multiple keychains? 

> I ask this because with the first option, I can just take the first transaction from the "spk pool" and set the wallet's birthday as the transaction's `BlockId`, instead of possibly having to overwrite it if another keychain has an older birthday and we sync it later.

Yup, I agree this should work if we are not allowing addition of keychains!

evanlinjin:
@110CodingP thanks for the reply. This comment is my understanding of the situation. That said, I shouldn't have the final say with bdk_wallet. I will say that I never got a "this approach is bad" response to my suggestion — but maybe there are flaws with my thinking here which aren't voiced. It'd be worth checking with the other maintainers.

src/wallet/mod.rs
```diff
@@ -583,36 +615,58 @@ where
...
-    // }
-
     /// Finds how the wallet derived the script pubkey `spk`.
     ///
     /// Will only return `Some(_)` if the wallet has given out the spk.
     pub fn derivation_of_spk(&self, spk: ScriptBuf) -> Option<(K, u32)> {
         self.indexed_graph.index.index_of_spk(spk).cloned()
     }
+
+    /// Get unbounded script pubkey iterators for both `Internal` and `External` keychains.
```
110CodingP: nit: `Internal` and `External` should be removed?

src/wallet/persisted.rs
```diff
@@ -380,7 +376,7 @@ pub enum LoadWithPersistError<E, K> {
     /// Error from persistence.
     Persist(E),
     /// Occurs when the loaded changeset cannot construct [`Wallet`].
-    InvalidChangeSet(LoadError<K>),
+    InvalidChangeSet(crate::LoadError<K>),
```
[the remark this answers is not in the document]
thunderbiscuit: Yep thanks!

thunderbiscuit (commented):
I found some methods that are cut out from the codebase because of commit 5df393d4e1329e9dd9e46c5dad099ed8a857797b, we should make sure we bring those back (or remove only the ones we really want to remove).

src/wallet/mod.rs
```diff
@@ -2572,217 +1437,1009 @@ impl Wallet {
...
-    /// let func = |wallet: &mut Wallet| wallet.apply_update(update);
-    /// let events = wallet.events_helper(func)?;
-    /// # Ok::<(), anyhow::Error>(())
-    /// ```
-    ///
-    /// # Errors
-    ///
-    /// If `f` returns an error, then returns `E` of a type defined by the function
-    /// passed in.
-    pub fn events_helper<F, T, E>(&mut self, f: F) -> Result<Vec<WalletEvent>, E>
```
thunderbiscuit, line 2652: Some methods are removed entirely from the module, for example here `Wallet::events_helper`.

110CodingP: Thanks for catching this! This shouldn't have happened though because this is just a rebase over master. Trying to figure out the reason.

ValuedMammal:
## How to preserve the automatic change derivation mechanism

### Problem

In the 2-keychain wallet, the change address can be automatically derived because we know which keychain is the "internal" one. However once we allow potentially many keychains we can no longer assume which keychain may be designated for change. It would be annoying, and arguably a regression, for the user to specify the change keychain on the PSBT params every time.

Proposed solution is to add an internal flag to each wallet descriptor at wallet creation, persisted via the ChangeSet. Knowing which keychain(s) can be used for change not only allows us to derive the change script automatically, but match the change script to the dominant output script type (for privacy reasons).

### Example

```rs
/// Represents a wallet descriptor
pub struct WalletDescriptor<D> {
    descriptor: D,
    internal: bool,
}

impl WalletDescriptor<D> {
    /// A new wallet descriptor for generating receive addresses.
    pub fn new(descriptor: D) -> Self { ... }

    /// A new wallet descriptor with the internal flag set indicating it may be used to generate change addresses.
    pub fn new_internal(descriptor: D) -> Self { ... }
}

impl KeyRing<K> {
    /// Create new from (keychain, wallet_descriptor) tuples.
    pub fn new<D>(descriptors: impl IntoIterator<Item = (K, WalletDescriptor<D>)>) -> Self
    where
        D: IntoWalletDescriptor,
    { ... }
}
```

thunderbiscuit:
I like the idea of simplifying the work for users (particularly if they have a given keychain that's _always_ the change keychain for a given descriptor, in which case they would as you say just always have to set it as their change keychain on every PSBT build).

Looking at the pseudocode above though, I think it might need adjusting just tad. In this scenario (at least in my mind), the keychains should be provided in pairs (just a marker that says "this is internal" is not enough if you actually have 6 keychains, and some of them have an internal, sister-keychain but they're all different, some don't have change keychain altogether, etc.).

Also tiny nit: I have seen enough users misunderstand the meaning of internal/external as to scare me for a lifetime. I don't know if this PR is the place to start this journey, but I think we should promote the receive/change taxonomy, or maybe just main/change.

ValuedMammal:
The main design hurdle is the ability of the wallet to derive a change address during transaction creation. For that we need to encode whether a keychain is eligible to "receive" change. In the process, to remove ambiguity in keychain selection, we should finally move away from using `map_keychain`.

In the single-keychain case, the user can mark their only keychain as also eligible for change addresses and the process should "just work" for deriving addresses in all cases. Of course the user can override it by specifying the change script directly on `PsbtParams`. If not explicitly provided, and if no keychains are marked "internal" by the wallet then `create_psbt` should return a "MissingChangeScript" error or similar.

The need to quickly access descriptor metadata in the wallet (e.g. the internal/change flag) again supports the argument for having `KeyRing` as a member of the `Wallet` structure.

thunderbiscuit:
> we should finally move away from using map_keychain

Yep 100% agree!

> If not explicitly provided, and if no keychains are marked "internal" by the wallet then create_psbt should return a "MissingChangeScript" error or similar.

Also agree. One thing we need to consider however, is that there is potentially multiple keychains marked as candidates for change, and then again you'd need a way to discriminate. A naive version of this could use keychain pairs, but then what happens when the wallet creates a transaction that pulls in UTXOs from different keychains? They each would potentially have a change keychain and then we haven't quite solved our issue. Maybe requiring users provide a change keychain on all transaction creation seems like a good basic first step, and then we can relax this by providing further shortcuts and helpers?

> supports the argument for having KeyRing as a member of the Wallet structure. 

Darn! I kind of agree here, metadata is useful of course and we need access. But one of the weird aspects of the keyring as part of the wallet is that the descriptors then exist in two places (in the tx_graph and the keyring field). If there is a way to just use the keychains in the tx_graph I think that's maybe a good way to not duplicate or split the source of truth for the wallet. I've honestly been going back and forth on this (now after thinning the KeyRing as much as I have I kind of even find myself wondering if it's important enough to keep altogether... need to deep dive on this a bit more).

## Update on this PR

I worked on the rebase of this yesterday all day and today, and it's become clear to me that the 15k green 15k red line diff is super hard to (1) manage + rebase, and (2) review (too much is just pre-keychain work and too little is actual API changes).

I'm testing out a different approach:
1. Create a branch that's called `pre-multi-keychain`. This contains work that's required but not directly related to multi-keychain. For example do we just cut the TxBuilder out entirely? Do it there. If so, the tests all need updating. Many of the wallet tests can migrate to Wallet::create_psbt. Then remove the signers. Stuff like that. This one is sort of loosely defined and I understand could be contentious, because it's sort of assuming a 4.0 world. My current work on this
(a) uses basically the separation between TxBuilder and signers explored in #505, and (b) removes the signers.
2. Open the multi-keychain PR against that branch! The PR becomes simpler and easier to review conceptually, even if we don't necessarily know what the shape of 4.0 on master will be yet.

[Example of this here](https://github.com/thunderbiscuit/bdk_wallet/pull/6).

ValuedMammal:
> Also agree. One thing we need to consider however, is that there is potentially multiple keychains marked as candidates for change, and then again you'd need a way to discriminate. A naive version of this could use keychain pairs, but then what happens when the wallet creates a transaction that pulls in UTXOs from different keychains? They each would potentially have a change keychain and then we haven't quite solved our issue.

Just pick one. To reiterate, the privacy focused approach is to choose the change keychain based on the output type of the *outputs* of the transaction to be created, not the keychain of the utxos pulled in as inputs.

> Darn! I kind of agree here, metadata is useful of course and we need access. But one of the weird aspects of the keyring as part of the wallet is that the descriptors then exist in two places (in the tx_graph and the keyring field).

You can have a list of internal keychains, or a map of keychain to "descriptor metadata" or something similar, but the cleanest way to define the change keychain is at wallet creation time, because it's tied directly to a wallet descriptor. Supplying the change keychains another way means you'll have to handle a possibly non-existing keychain. We're already changing the persistence layer by supporting multiple keychains. The internal flag would be 1 additional column in SQLite for example

| keychain | descriptor | internal |
| --- | --- | --- | 
| "keychain_1" | desc_1 | 0 |
| "keychain_2" | desc_2 | 1 |

This is analogous to the `importdescriptors` RPC which asks you to set the "internal" flag for change descriptors. This way we can realize notmandatory's dream of importing a Core wallet into BDK.

So for PSBT creation the order of precedence is

1. User provided change-script/descriptor via the params
2. Auto derive from a change keychain the Wallet/KeyRing knows about
3. A third option could be a single globally configurable `K` value, but the real benefit of this comes from persisting it, so you can truly set it and forget it

110CodingP:
> Supplying the change keychains another way means you'll have to handle a possibly non-existing keychain. We're already changing the persistence layer by supporting multiple keychains. The internal flag would be 1 additional column in SQLite for example.

Agreed!

> This is analogous to the `importdescriptors` RPC which asks you to set the "internal" flag for change descriptors. This way we can realize notmandatory's dream of importing a Core wallet into BDK.


But Mammal, Core could tell us about which keychains are `internal` and then we can encode that information in `K` itself? But I agree this doesn't tell the wallet about the keychains that can be used for change.

PR #319, opened 2025-09-26
Return wallet events when applying updates and blocks (3.0 milestone)

notmandatory:
### Description

Cherry-pick of commits from #310 and #336.
       
### Notes to the reviewers

See https://github.com/bitcoindevkit/bdk_wallet/pull/319#issuecomment-3829852077 for proposed follow-up work.

### Changelog notice

No changes from `release/2.x` branch.

BREAKING:

- `wallet::event` module is made private. `WalletEvent` can now be imported from the root level.

#### All Submissions:

#### New Features:


wallet/src/wallet/mod.rs
```diff
@@ -76,6 +77,7 @@ use crate::wallet::{
 };
 
 // re-exports
+use crate::event::{wallet_events, WalletEvent};
```
ValuedMammal, line 80: This would be a good place to re-export the `WalletEvent` type. That would also help keep the `use` statements organized.

wallet/src/wallet/mod.rs
```diff
@@ -2370,6 +2372,126 @@ impl Wallet {
...
+            .map(|wtx| {
+                (
+                    wtx.tx_node.txid,
+                    (wtx.tx_node.tx.clone(), wtx.chain_position),
+                )
+            })
+            .collect::<BTreeMap<Txid, (Arc<Transaction>, ChainPosition<ConfirmationBlockTime>)>>();
+
+        // apply update
+        self.apply_update(update)?;
```
ValuedMammal, line 2472: Any reason not to apply the update and use the resulting changeset to produce a summary of what changed in the Wallet?

wallet/src/wallet/event.rs
```diff
@@ -0,0 +1,182 @@
...
+    /// a [`WalletEvent::TxUnconfirmed`] event.
+    TxDropped {
+        /// Transaction id.
+        txid: Txid,
+        /// Transaction.
+        tx: Arc<Transaction>,
+    },
+}
+
+pub(crate) fn wallet_events(
```
ValuedMammal, line 85: 📌 A code comment somewhere would help to explain what this function is doing.

wallet/src/wallet/event.rs
```diff
@@ -0,0 +1,182 @@
...
+    wallet_txs2: BTreeMap<Txid, (Arc<Transaction>, ChainPosition<ConfirmationBlockTime>)>,
+) -> Vec<WalletEvent> {
+    let mut events: Vec<WalletEvent> = Vec::new();
+
+    if chain_tip1 != chain_tip2 {
+        events.push(WalletEvent::ChainTipChanged {
+            old_tip: chain_tip1,
+            new_tip: chain_tip2,
+        });
+    }
```
ValuedMammal, line 99: I think it would be helpful to know which blocks were connected and/or disconnected.

wallet/src/wallet/event.rs
```diff
@@ -0,0 +1,182 @@
...
+
+    wallet_txs2.iter().for_each(|(txid2, (tx2, cp2))| {
```
ValuedMammal, line 101: nit: Please use a `for` loop for this.

wallet/src/wallet/event.rs
```diff
@@ -0,0 +1,182 @@
...
+        if let Some((tx1, cp1)) = wallet_txs1.get(txid2) {
```
ValuedMammal, line 102: `pos1` or `chain_position1`. "cp" usually signifies something else.

wallet/tests/wallet_event.rs
```diff
@@ -0,0 +1,300 @@
+use bdk_chain::{BlockId, CheckPoint, ConfirmationBlockTime};
+use bdk_wallet::event::WalletEvent;
+use bdk_wallet::test_utils::{get_test_wpkh_and_change_desc, new_wallet_and_funding_update};
+use bdk_wallet::{SignOptions, Update};
```
ValuedMammal, line 4: This will be more future proof if you can do it using the unsigned txs.

wallet/src/wallet/event.rs
```diff
@@ -0,0 +1,182 @@
...
+            assert_eq!(tx1.compute_txid(), *txid2);
```
ValuedMammal, line 103: Maybe just a `debug_assert` here.

ValuedMammal (changes requested):
Thanks @notmandatory

wallet/src/wallet/event.rs
notmandatory: Do you mean a list of connected/disconnected block ids? I didn't add any other info here because @tnull didn't need it for LDK's use  case. Do you have some use or user in mind for the block ids?

wallet/src/wallet/mod.rs
notmandatory: The resulting `Wallet::ChangeSet` only contains the new blocks, tx, and anchors found after applying the sync update but don't show how this new data changes the canonical status of existing tx.

src/wallet/event.rs
```diff
@@ -0,0 +1,182 @@
...
+    wallet_txs2: BTreeMap<Txid, (Arc<Transaction>, ChainPosition<ConfirmationBlockTime>)>,
+) -> Vec<WalletEvent> {
+    let mut events: Vec<WalletEvent> = Vec::new();
+
+    if chain_tip1 != chain_tip2 {
+        events.push(WalletEvent::ChainTipChanged {
+            old_tip: chain_tip1,
+            new_tip: chain_tip2,
+        });
+    }
```
ValuedMammal: Yeah I mean the block IDs. It's just a more complete way of describing how the chain tip changed.

notmandatory:
I will cherry-pick and incorporate commits from #336 for `apply_block_events` and `apply_block_connected_to_events` when it's merged.

notmandatory:
Rebased on `master` branch.

src/wallet/mod.rs
```diff
@@ -2370,6 +2372,126 @@ impl Wallet {
...
+            .map(|wtx| {
+                (
+                    wtx.tx_node.txid,
+                    (wtx.tx_node.tx.clone(), wtx.chain_position),
+                )
+            })
+            .collect::<BTreeMap<Txid, (Arc<Transaction>, ChainPosition<ConfirmationBlockTime>)>>();
+
+        // apply update
+        self.apply_update(update)?;
```
ValuedMammal: We make a logical assumption that the information in the events matches the data that is persisted via the `ChangeSet`, so my thinking is that an adequate test for this feature should include checking that the events do in fact agree with the newly staged changes. I may look into adding a test for it.

src/wallet/event.rs
notmandatory: @ValuedMammal are you thinking of something like adding `connected` and `disconnected` fields to this variant:

```rust
    /// The latest chain tip known to the wallet changed.
    ChainTipChanged {
        /// Previous chain tip.
        old_tip: BlockId,
        /// New chain tip.
        new_tip: BlockId,
        /// Newly connected blocks.
        connected: Vec<BlockId>,
        /// Newly disconnected blocks.
        disconnected: Vec<BlockId>,
    },
```

I'm  still not clear on how this info would be used by a wallet user since it's more of an internal detail.

notmandatory:
Refactored to return events from original apply_update, apply_block* functions, removed *_events functions, see 66cedb47425a255100b0102fb33c471928c71459.

src/wallet/event.rs
ValuedMammal: I've seen a similar thing used in LDK, for example [`Cache`](https://docs.rs/lightning-block-sync/latest/lightning_block_sync/trait.Cache.html) and [`Listen`](https://docs.rs/lightning/latest/lightning/chain/trait.Listen.html), but don't know if it's also applicable in this context.

notmandatory:
Added one more commit with some docs improvements and refactored out helper method based on suggestions from Claude LLM 🤖.

notmandatory:
I haven't added an example for getting affected UTXOs/addresses from a Transaction related WalletEvents. This is a bit out of scope for this PR and should get it's own PR, either to add an example or add a helper function to Wallet to do it.

notmandatory:
> I haven't added an example for getting affected UTXOs/addresses from a Transaction related WalletEvents. This is a bit out of scope for this PR and should get it's own PR, either to add an example or add a helper function to Wallet to do it.

My thinking is to provide a new helper method `Wallet::send_and_received_txouts` to accomplish this. See: https://github.com/bitcoindevkit/bdk/pull/2081

evanlinjin (commented):
Thanks for working on this!

Since this is a 3.0 milestone, we will have `CanonicalView` available. I would store the latest `CanonicalView` as a field in `Wallet` and update it every time an update is applied.

Events can be obtained by diff-ing the new view against the old view.

```rust
impl CanonicalView {
    /// Not implemented.
    pub fn diff(other: &Self) -> Diff { todo!() }
}
```

If `apply_update` returns the old view, we can do this:

```rust
let old_view = wallet.apply_update(update)?;
let diff = old_view.diff(wallet.view() /* current view */);
```

Then `diff` can be treated as "events".

Here are some benefits to this:

* This feature will be available for callers that don't use `bdk_wallet`.
* You aren't forced to do the event processing (it's opt-in).
* Only canonicalize once (after applying update).

In other words, I propose implementing this in `bdk_chain` as a method on `CanonicalView`. Let me know what you think.

evanlinjin:
> > In other words, I propose implementing this in `bdk_chain` as a method on `CanonicalView`. Let me know what you think.
> 
> Hmm, do I understand correctly that you propose to revert the API changes from #310, i.e., no longer expose `WalletEvents`? While surely doable, I want to note that that would mean a good bit of churn for us, as we _just_ started migrating our internals to the APIs introduced in `bdk_wallet` v2.2 / v2.3.

No. Essentially moving `WalletEvents` to `bdk_chain` (probably renaming it to something else). `WalletEvents` will be a diff between two `CanonicalView`s.

However, I did suggest some API changes - one benefit of my suggestion is you can choose exactly where and when you diff.

```rust
// Look you can apply multiple things and then diff later!
let old_view = wallet.apply_update(update1)?;
let _ = wallet.apply_update(update2)?;
// Look, we switch to a block-based chain source in the middle.
let _ = wallet.apply_block(&block, height)?;

// We diff now! Capturing the changes made by update1 + update2 + block.
let diff = old_view.diff(wallet.view() /* current view */);
```

Maybe we don't even need `apply_x` methods to return the old `CanonicalView`.

```rust
let old_view = wallet.view();
wallet.apply_update(update1)?;
wallet.apply_update(update2)?;
wallet.apply_block(&block, height)?;
let diff = old_view.diff(wallet.view());
```

I would argue this is probably even cleaner.

If the `bdk_chain::CanonicalViewDiff` thing departs too much from the current `WalletEvent` enum, we can have a helper method to convert it to `WalletEvent`.

src/wallet/mod.rs
```diff
@@ -52,6 +52,7 @@ use rand_core::RngCore;
 mod changeset;
 pub mod coin_selection;
 pub mod error;
+pub mod event;
```
ValuedMammal: This doesn't really need to be `pub`.

ValuedMammal:
ACK fcdc00653fec50f28cc3cbde69970f47d155eaa1

PR #337, opened 2025-10-29
fix(wallet): Don't fail in `build_fee_bump` for missing parent txid 

### LINKED ISSUE #325: Can't build a fee bump transaction if spending a P2A output
**Describe the bug**  
In [Bark](https://codeberg.org/ark-bitcoin/bark/src/commit/1b715575c41da6f5ed2c449bd5748e172cef8155/bitcoin-ext/src/bdk.rs#L126) we create TRUC transactions using ephemeral anchors to perform unilateral exits. The parent transaction in the TRUC originates outside of the wallet and we use BDK's tx builder to create a child transaction that spends the P2A output. Currently the fee system can't calculate the effective fee rate of the TRUC package and that is fine, we work around it by looping and calculating an absolute fee based on the generated child transaction. 

The problem we do have is that if we wish to RBF the child transaction from the same users wallet we are unable to use `build_fee_bump` because the P2A spend input doesn't originate from the users wallet. Unfortunately I can't find a method to remove unconfirmed TRUC child from the wallet to then generate a new one and reuse the existing UTXOs. This puts us in a situation where we have to recommend users keep multiple UTXOs spare for the purpose of fee bumping which isn't necessarily ideal in an emergency unilateral exit situation.

**To Reproduce**  
 1. Create a transaction with a ephemeral anchor output outside the wallet
 2. Use `build_tx` to start building a new transaction with the wallet
 3. Use `add_foreign_utxo` to add the P2A spend as an input into the child transaction
 4. Finish creating the transaction and persist it in the wallet
 5. Use `build_fee_bump` to create an RBF transaction with the TX ID of the newly created child transaction
 6. Notice that an error is returned because the origin of the P2A spend input isn't the wallet itself

**Expected behavior**  
If I use a foreign UTXO when building a transaction I should be able to use `build_fee_bump` to RBF the newly built transaction. Having some way to specify an input as foreign to bypass the UTXO checks would be ideal, alternatively if I could remove the unconfirmed transaction from the history of the wallet entirely that would also be helpful.

**Build environment**  
- BDK tag/commit: v2.1.0
- OS+version: MacOS 15.5
- Rust/Cargo version 1.88
- Rust/Cargo target: aarch64-apple-darwin

- [ ] Electrum
- [ ] Esplora
- [ ] Bitcoin Core RPC
- [ ] Other (please specify): `____`

- [ ] Yes
- [x] No

[Second](https://second.tech/)

**Additional context**  
Some logs showing the error:
```
Attempting to bump the fee of wallet tx b7572a182ccf392634a5178010cfd4e48bee7ccb72b58a0e8d086c5e87810e7e for exit 48a0caba9f5ec585d1c26e5c9edf3c36cf26e7f66b7254e044e33f1ed85233e4

Error creating signed P2A CPFP: Failed to bump wallet tx b7572a182ccf392634a5178010cfd4e48bee7ccb72b58a0e8d086c5e87810e7e: UTXO not found in the internal database with txid: 48a0caba9f5ec585d1c26e5c9edf3c36cf26e7f66b7254e044e33f1ed85233e4, vout: 1
```

Links to our code:
- [make_signed_p2a_cpfp function](https://codeberg.org/ark-bitcoin/bark/src/commit/1b715575c41da6f5ed2c449bd5748e172cef8155/bitcoin-ext/src/bdk.rs#L126)
- [add_fee_anchor_spend extension](https://codeberg.org/ark-bitcoin/bark/src/commit/1b715575c41da6f5ed2c449bd5748e172cef8155/bitcoin-ext/src/bdk.rs#L22)
- [create_exit_tx containing P2A output](https://codeberg.org/ark-bitcoin/bark/src/commit/1b715575c41da6f5ed2c449bd5748e172cef8155/lib/src/vtxo/mod.rs#L212)

ValuedMammal:
### Description

Previously `build_fee_bump` could error with `UnknownUtxo` if a parent of the tx being fee-bumped wasn't found in the wallet. This made it impossible to use `build_fee_bump` on a transaction created using `add_foreign_utxo`. This PR is a refactor of `build_fee_bump` that manages to avoid the error by instead querying the tx graph for the txout specifically. This is a reasonable assumption because the previous txouts are necessary to compute the fee of the original tx.

I've included a test in 992a08fd6f2aacf58232d9c27d0790378f9f78c3 to demonstrate the old behavior which now passes.

may resolve #325.

### Changelog notice

- Fix `Wallet::build_fee_bump` to enable fee-bumping a tx which contains a foreign UTXO.

#### All Submissions:

#### Bugfixes:

* ~~[ ] This pull request breaks the existing API~~

### DIFF
--- a/src/wallet/mod.rs
+++ b/src/wallet/mod.rs
@@ -1711,15 +1711,15 @@ impl Wallet {
         &mut self,
         txid: Txid,
     ) -> Result<TxBuilder<'_, DefaultCoinSelectionAlgorithm>, BuildFeeBumpError> {
-        let graph = self.indexed_graph.graph();
+        let tx_graph = self.indexed_graph.graph();
         let txout_index = &self.indexed_graph.index;
         let chain_tip = self.chain.tip().block_id();
-        let chain_positions = graph
+        let chain_positions: HashMap<Txid, ChainPosition<_>> = tx_graph
             .list_canonical_txs(&self.chain, chain_tip, CanonicalizationParams::default())
             .map(|canon_tx| (canon_tx.tx_node.txid, canon_tx.chain_position))
-            .collect::<HashMap<Txid, _>>();
+            .collect();
 
-        let mut tx = graph
+        let mut tx = tx_graph
             .get_tx(txid)
             .ok_or(BuildFeeBumpError::TransactionNotFound(txid))?
             .as_ref()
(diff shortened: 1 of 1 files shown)


src/wallet/mod.rs
```diff
@@ -1746,73 +1746,63 @@ impl Wallet {
         let fee = self
             .calculate_fee(&tx)
             .map_err(|_| BuildFeeBumpError::FeeRateUnavailable)?;
-        let fee_rate = self
-            .calculate_fee_rate(&tx)
-            .map_err(|_| BuildFeeBumpError::FeeRateUnavailable)?;
+        let fee_rate = fee / tx.weight();
```
nymius: What's the rationale to not use `calculate_fee_rate`?

tests/add_foreign_utxo.rs
```diff
@@ -290,3 +290,48 @@ fn test_taproot_foreign_utxo() {
...
+fn test_add_foreign_utxo_bump_fee() {
+    // Create tx spending a p2a output
+    let (mut wallet, _) = get_funded_wallet_wpkh();
+
+    let drain_spk = wallet
+        .next_unused_address(KeychainKind::External)
+        .script_pubkey();
+    let witness_utxo = TxOut {
+        value: Amount::ZERO,
+        script_pubkey: ScriptBuf::new_p2a(),
```
nymius: `new_p2a` is not in `rust-bitcoin 0.32.6`, but in `rust-bitcoin 0.32.7`. Missing `Cargo.toml` update I guess

tests/add_foreign_utxo.rs
```diff
@@ -290,3 +290,48 @@ fn test_taproot_foreign_utxo() {
         "foreign_utxo should be in there"
     );
 }
+
+#[test]
+fn test_add_foreign_utxo_bump_fee() {
```
nymius: I would mention the type of script used in the first tx input, as I think is important for discoverability and also to illustrate the context of what this test is doing.

src/wallet/mod.rs
```diff
@@ -1746,73 +1746,63 @@ impl Wallet {
...
-                    // Get chain position.
-                    .and_then(|prev_tx| {
+                let outpoint = txin.previous_output;
+                let prev_txout = tx_graph.get_txout(outpoint).cloned();
+                match prev_txout.as_ref().and_then(|prev_txout| {
+                    txout_index.index_of_spk(prev_txout.script_pubkey.clone())
+                }) {
+                    // This is a local utxo.
+                    Some(&(keychain, derivation_index)) => {
+                        let txout = prev_txout.ok_or(BuildFeeBumpError::UnknownUtxo(outpoint))?;
```
nymius: I think the error here is impossible to reach.
If the spk is present on the index, then this wallet collaborated on the creation of the transaction with the signature of that input, then that UTXO should have been already known by this wallet, right?
Also, we are getting this from the `TxGraph` itself and using an `and_then` call. All `UnknownUtxos` will be transformed into `None` and parsed by the other match arm before reaching this line.

nymius (commented):
Concept ACK 8a08bf53d10134cb6dae9f8fbd8ecef23c934363

notmandatory:
I see this PR is in the "Wallet 2.3.0" milestone but merging to the `master` branch. Is the plan to merge it to `master` then back port it to `release/2.x` ?

nymius:
I've been thinking about calling `Wallet.insert_txout` from `TxBuilder.add_foreign_utxo` and I think is going to make the API much more entangled to what I would like.
I prefer to document the workflow rather than embedding this logic into `TxBuilder`.
Something that may be needed though in `Wallet` is a method to insert transactions, as the only available now is `insert_txout` but for non segwit foreign UTXOs the full transaction will be needed at the moment of building the tx RBF.

tests/build_fee_bump.rs
```diff
@@ -944,3 +944,51 @@ fn test_legacy_bump_fee_absolute_add_input() {
...
+// Test that we can fee-bump a tx containing a foreign (p2a) utxo.
+#[test]
+fn test_bump_fee_pay_to_anchor_foreign_utxo() {
+    let (mut wallet, _) = get_funded_wallet_wpkh();
+    let drain_spk = wallet
+        .next_unused_address(KeychainKind::External)
+        .script_pubkey();
+
+    let witness_utxo = TxOut {
+        value: Amount::ONE_SAT,
```
notmandatory: ⛏️ nit, in a typical P2A tx this would be zero.

notmandatory (approved):
ACK f15582df8e70fb5ee0a7cfe622b0328f1da8f705

src/wallet/mod.rs
```diff
@@ -1711,15 +1711,15 @@ impl Wallet {
...
         let chain_tip = self.chain.tip().block_id();
-        let chain_positions = graph
+        let chain_positions: HashMap<Txid, ChainPosition<_>> = tx_graph
             .list_canonical_txs(&self.chain, chain_tip, CanonicalizationParams::default())
             .map(|canon_tx| (canon_tx.tx_node.txid, canon_tx.chain_position))
-            .collect::<HashMap<Txid, _>>();
+            .collect();
 
-        let mut tx = graph
+        let mut tx = tx_graph
```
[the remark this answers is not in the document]
ValuedMammal: Can you share a snippet or code suggestion so I understand better what you're referring to?

PR #343, opened 2025-11-07
feat(persist_test_utils)!: Add async suite to test AsyncWalletPersister

ValuedMammal:
### Description

Refactors the `persist_test_utils` module and extends it with an async test suite for `AsyncWalletPersister`.

903bd848be3f157747c3f690ae7bd020064012ed: Refactor WalletPersister suite + update deps

The public test functions previously took a `filename: &str` and a path-based closure, managed `tempfile::tempdir()` internally, and panicked on failure. This redesigns the API so callers supply a `FnOnce() -> Result<P, P::Error>` closure that owns the store directly, and all functions return `Result<(), PersistError>` instead of panicking. A new `PersistError` enum exposes failures as structured values with `ChangeSetMismatch { got, expected }` and `Persister(Box<dyn core::error::Error>)` variants. `persist_single_keychain` is removed, and its coverage is absorbed into `persist_keychains`, which now runs two rounds: external descriptor only (the single-keychain case), then external plus change descriptor with merge verification. `anyhow` and `tempfile` are removed from the `test-utils` feature's optional dependencies and kept only as `dev-dependencies`.

6dc8fe5580f3d2dda670f30aca108b419d96280e: Add async suite to test AsyncWalletPersister

Adds async counterparts to every public function in the `WalletPersister` test suite so implementors of `AsyncWalletPersister` can exercise the same scenarios: `persist_wallet_changeset_async`, `persist_keychains_async`, and `persist_network_async`. Two private helpers, `init_async_wallet_persister` and `persist_changeset_verify_async`, mirror the sync suite API. All entry points use the `F: AsyncFnOnce() -> Result<P, P::Error>` bound, consistent with the sync API pattern.

Potential follow-ups:

- [ ] Add async counterpart for `persist_multiple_wallet_changesets`

### Changelog notice

**Added**
- `persist_wallet_changeset_async`, `persist_keychains_async`, `persist_network_async` to `persist_test_utils` for testing `AsyncWalletPersister` implementations

**Changed**
- Refactored `persist_test_utils` public API: callers own store construction; functions now return `Result<(), PersistError>` instead of panicking
- Added `PersistError` enum with `ChangeSetMismatch` and `Persister` variants
- `persist_keychains` now covers the single-keychain case; `persist_single_keychain` removed
- Removed optional `anyhow` and `tempfile` dependencies from the `test-utils` feature

#### All Submissions:

#### New Features:

### DIFF
--- a/src/persist_test_utils.rs
+++ b/src/persist_test_utils.rs
@@ -14,7 +14,7 @@ use bitcoin::{
 };
 use miniscript::{Descriptor, DescriptorPublicKey};
 
-use crate::{locked_outpoints, ChangeSet, WalletPersister};
+use crate::{locked_outpoints, AsyncWalletPersister, ChangeSet, WalletPersister};
 
 macro_rules! block_id {
     ($height:expr, $hash:literal) => {{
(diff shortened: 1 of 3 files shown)


src/persist_test_utils.rs
```diff
@@ -74,23 +77,322 @@ fn spk_at_index(descriptor: &Descriptor<DescriptorPublicKey>, index: u32) -> Scr
...
+///
+/// We create a dummy [`ChangeSet`] for first wallet and persist it then we create a dummy
+/// [`ChangeSet`] for second wallet and persist that. Finally we load these two [`ChangeSet`]s and
+/// check if they were persisted correctly.
+pub fn persist_multiple_wallet_changesets<F, P>(create_stores: F) -> Result<(), PersistError>
+where
+    F: Fn() -> Result<(P, P), P::Error>,
+    P: WalletPersister,
+    P::Error: StdErr + 'static,
+{
```
ValuedMammal: @110CodingP I changed this by removing the `&Path` arg from the `create_stores` `Fn`, which means this function is no longer in control of the path in which the stores are created. After thinking about it I'm not sure this is a feature that needs to be tested. Any thought about whether to keep or improve it somehow?

tests/persisted_wallet.rs
```diff
@@ -429,42 +429,78 @@ fn single_descriptor_wallet_persist_and_recover() {
...
-    persist_wallet_changeset("store.db", |path| {
-        Ok(bdk_file_store::Store::create(DB_MAGIC, path)?)
-    });
-    persist_wallet_changeset::<bdk_chain::rusqlite::Connection, _>("store.sqlite", |path| {
-        Ok(bdk_chain::rusqlite::Connection::open(path)?)
-    });
+    let tmpdir = tempfile::tempdir().unwrap();
+
+    // Test file_store
+    #[allow(clippy::result_large_err)]
```
ValuedMammal: Should be fixed as of file_store-0.22.0 (found 4 instances).
```suggestion
```

KrishnaShuk:
Concept ACK

The sync side has 5 test utilities but only persist_wallet_changeset got an async equivalent. 
persist_network, persist_keychains, persist_single_keychain, and persist_multiple_wallet_changesets are missing async versions.
A backend could pass the full changeset round-trip but fail on partial writes. Is this intentional or planned for follow-up?

ValuedMammal:
> The sync side has 5 test utilities but only persist_wallet_changeset got an async equivalent.
> persist_network, persist_keychains, persist_single_keychain, and persist_multiple_wallet_changesets are missing async versions.

Added `persist_network_async`. `persist_keychains_async` also covers the single keychain case.

bladed-operative:
Replaced

PR #348, opened 2025-12-02
Add wallet birthday to `CreateParams`

rustaceanrob:
There has been discussion offline on how to improve the experience for recovering wallets using block-by-block scanning methods. A birthday has come up multiple times, so I introduce one here in `CreateParams`.

I believe the `expect` may only be hit when attempting to over-write the genesis block. If we want to return the error, it would require a break in the error type. IMO this seems unnecessary, as a birthday should not attempt to replace genesis.

This may be used immediately in `Wallet::latest_checkpoint` or similar.

### Changelog notice

- Added `birthday` parameter to `CreateParams`

#### All Submissions:

#### New Features:

#### Bugfixes:

### DIFF
--- a/src/wallet/params.rs
+++ b/src/wallet/params.rs
@@ -1,6 +1,6 @@
 use alloc::boxed::Box;
 
-use bdk_chain::keychain_txout::DEFAULT_LOOKAHEAD;
+use bdk_chain::{keychain_txout::DEFAULT_LOOKAHEAD, BlockId};
 use bitcoin::{BlockHash, Network, NetworkKind};
 use miniscript::descriptor::KeyMap;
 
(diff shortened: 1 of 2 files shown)


src/wallet/params.rs
```diff
@@ -162,6 +166,13 @@ impl CreateParams {
         self
     }
 
+    /// Begin wallet scanning from the specified birthday. Particularly useful for block-based
+    /// scanning sources.
+    pub fn birthday(mut self, birthday: BlockId) -> Self {
+        self.birthday = Some(birthday);
```
tnull: Hmm, it's kind of odd to having to always having to remember the birthday outside of the wallet and setting it on `CreateParams`. Shouldn't the wallet itself also remember its birthday, and would it need to be persisted as part of the `ChangeSet`?

luisschwab:
I was thinking that it should be possible for a chain source's result to automatically set the wallet birthday, instead of having to set it manually based on a result. Not sure if this would require going the `ChangeSet` route though. cc @oleonardolima

PR #352, opened 2025-12-05
feat: implement new tx builder crate for bip353 payment instructions

### LINKED ISSUE #13: Support DNS payment instructions, BIP353
**Describe the enhancement**  

A new transaction builder crate should add an optional feature to enable DNS payment instructions support. If this option is enabled the builder will include a new function to add a recipient via their DNS payment instructions name, i.e. `₿user@domain` or `user@domain`, instead of a bitcoin address. 

An error must be thrown if the DNS payment instructions lookup does not return valid payment instructions that include an on-chain address. An error should also be thrown if the user supplied payment amount does not equal the returned payment instructions fixed amount or is not in the min/max range for a configurable amount. 

If a DNS payment instructions derived address is used for an output then the `PSBT_OUT_DNSSEC_PROOF` must also be included in the resulting PSBT.

**Use case**  

This feature will allow a wallet developer to enable paying recipients via their human readable DNS payment instructions name. It will also allow hardware wallets that support BIP353 to validate and display the recipients payment instructions name. 

Initially this will only support static on-chain addresses which are likely to be reused and thus be less private. But when BDK is able to support Silent Payments this feature can be used in a way that avoids address reuse. 

For a full run-down of the pros/cons see [BIP353](https://bips.dev/353/).

**Additional context**

[BIP 21: URI Scheme](https://bips.dev/21/)
[BIP 353: DNS Payment Instructions](https://bips.dev/353/)
[BIP 352: Silent Payments](https://bips.dev/352/)

@thebluematt created a new [bitcoin-payment-instructions](https://docs.rs/bitcoin-payment-instructions/latest/bitcoin_payment_instructions/index.html) crate to lookup, validate and parse DNS bitcoin payment instructions. It should make the implementation of this feature pretty straight forward. See [discussion on discord](https://discord.com/channels/753336465005608961/753367451319926827/1357813717051506688).

sdmg15:
### Description

This PR adds support for BIP353 payment instructions.

Closes #13 

#### All Submissions:

* [] I ran `just p` before pushing

#### New Features:

#### Bugfixes:

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -29,6 +29,7 @@ anyhow = { version = "1.0.98", optional = true }
 bdk_file_store = { version = "0.21.1", optional = true }
 bip39 = { version = "2.0", optional = true }
 tempfile = { version = "3.20.0", optional = true }
+bitcoin-payment-instructions = { version = "0.5.0", optional = true}
 
 [features]
 default = ["std"]
@@ -39,6 +40,7 @@ keys-bip39 = ["bip39"]
 rusqlite = ["bdk_chain/rusqlite"]
 file_store = ["bdk_file_store"]
 test-utils = ["std", "anyhow", "tempfile"]
+bip353 = ["bitcoin-payment-instructions"]
 
 [dev-dependencies]
 anyhow = "1"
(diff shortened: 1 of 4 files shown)


luisschwab:
This should go on the new [`bdk_tx`](https://github.com/bitcoindevkit/bdk-tx) crate, as we're moving all transaction building logic over there.

sdmg15:
Closed in favor of https://github.com/bitcoindevkit/bdk-tx/pull/33

PR #355, opened 2025-12-15
feat: public key generation from descriptors

### LINKED ISSUE #209: Add deriving `public keys` from wallet
**Describe the enhancement**  

It would be ideal for the wallet API to generate public keys in the same manner addresses are created. Currently, getting public keys from descriptors seems quite cumbersome as one has to derive the xpub from which they can get a public key. In terms of security, I have not seen any significant breaches related to this.

**Use case**  
 
Generating public keys from a wallet will be useful in multi-sig or taproot setup where a wallet implementation allows users to collaboratively create addresses and perform transactions. Signing is already managed through PSBT.

sdmg15:
### Description

This PR as a new function to the wallet API `public_key_at_index(keychain, index)` which returns the public key found at the specified index. It closes #209 

### Notes to the reviewers

I explicitly matched  all arms branches of the `Descriptor<DefiniteDescriptorKey>` so that if a new type is added in the future it signals there is a need to update this function too.

### Changelog notice

- Add a helper function `public_key_at_index(keychain, index)` which returns the public key found at the specified index.
- 
#### All Submissions:

#### New Features:

### DIFF
--- a/src/wallet/mod.rs
+++ b/src/wallet/mod.rs
@@ -41,12 +41,13 @@ use bitcoin::{
     psbt,
     secp256k1::Secp256k1,
     sighash::{EcdsaSighashType, TapSighashType},
-    transaction, Address, Amount, Block, FeeRate, Network, NetworkKind, OutPoint, Psbt, ScriptBuf,
-    Sequence, SignedAmount, Transaction, TxOut, Txid, Weight, Witness,
+    transaction, Address, Amount, Block, FeeRate, Network, NetworkKind, OutPoint, Psbt, PublicKey,
+    ScriptBuf, Sequence, SignedAmount, Transaction, TxOut, Txid, Weight, Witness,
 };
 use miniscript::{
     descriptor::KeyMap,
     psbt::{PsbtExt, PsbtInputExt, PsbtInputSatisfier},
+    ToPublicKey,
 };
 use rand_core::RngCore;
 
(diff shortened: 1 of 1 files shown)


AdamuAbba (commented):
tACK [7ac497f](https://github.com/bitcoindevkit/bdk_wallet/pull/355/commits/7ac497f514ea03eea5880442b53f19a17a9bd0ca) 

Ran the test locally and it was successful.
Screenshot 2026-01-16 at 4 01 28 PM

ValuedMammal:
Since it can be done using rust-miniscript I don't think it's a high priority. Out of curiosity are you sure there's no way to return a public key (or keys) from a Bare, Sh, or Wsh? or is there a reason why that's not desirable?

sdmg15:
> Since it can be done using rust-miniscript I don't think it's a high priority. Out of curiosity are you sure there's no way to return a public key (or keys) from a Bare, Sh, or Wsh? or is there a reason why that's not desirable?

I'm happy to review the API if we want to return multiple public keys. The initial thought was returning only a single one. For wsh, sh to get the keys should require to parse the raw miniscript looking for `pk`, `pkh` no?

busayo-OD:
> 
> I'm happy to review the API if we want to return multiple public keys. The initial thought was returning only a single one. For wsh, sh to get the keys should require to parse the raw miniscript looking for `pk`, `pkh` no?

I looked into this and found that rust-miniscript already provides exhaustive key traversal through ForEachKey, including delegation into Sh, Wsh, and Taproot script-path leaves. So it looks like we could reuse that instead of parsing the Miniscript ourselves.

I also noticed the issue discussion eventually shifted toward signature verification rather than general key enumeration. Is public_key_at_index intended for that narrower use case, or is it meant to be a more general descriptor key API?

PR #364, opened 2026-01-12
Add get_pre_1_wallet_keychains migration helper

notmandatory:
### Description

Add `wallet/migration` module and `get_pre_1_wallet_keychains()` to help migrate users from a pre-1.0 bdk sqlite database. This new function returns the last revealed index and checksum value for each keychain it finds.

### Notes to the reviewers

This PR replaces bitcoindevkit/bdk#2090 since make more sense to put it in the wallet where it will be easier to export the bindings.  

### Changelog notice

* Add `get_pre_1_wallet_keychains` to assist migration from pre-1.0 bdk wallets.

#### All Submissions:

#### New Features:

### DIFF
--- a/src/wallet/migration.rs
+++ b/src/wallet/migration.rs
@@ -1,0 +1,138 @@
+// Bitcoin Dev Kit
+//
+// Copyright (c) 2020-2026 Bitcoin Dev Kit Developers
+//
+// This file is licensed under the Apache License, Version 2.0 <LICENSE-APACHE
+// or http://www.apache.org/licenses/LICENSE-2.0> or the MIT license
+// <LICENSE-MIT or http://opensource.org/licenses/MIT>, at your option.
+// You may not use this file except in accordance with one or both of these
+// licenses.
+
+//! Migration utilities for upgrading between BDK library versions.
+//!
+//! This module provides helper functions and types to assist users in migrating wallet data
+//! when upgrading between major versions of the `bdk_wallet` crate.
+
+#[cfg(feature = "rusqlite")]
+use crate::rusqlite::{self, Connection};
+#[cfg(feature = "rusqlite")]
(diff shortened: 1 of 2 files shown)


notmandatory:
> Two things I'd like to clarify maybe:
> 
> 1. Can we confirm the versions of bdk this will work for, and add it to the docstrings?

The only sqlite tables this function uses is `last_derivation_indices` and `checksums` which existed and haven't changed for any pre-1.0 version of `bdk`. So I'd prefer to leave it as-is and let folks know that this function works for any pre-1.0 version they're on.

> 2. Was there a reason to return the descriptor checksum instead of the whole descriptor string?

Unfortunately pre-1.0 we did not persist the descriptor string in the sqlite db, all wallet constructors required you pass in your descriptor(s), see: pre-1.0 [Wallet](https://docs.rs/bdk/latest/bdk/wallet/struct.Wallet.html) docs. All we have in the pre-1.0 sqlite db are the descriptor checksums.

src/wallet/migration.rs
```diff
@@ -0,0 +1,138 @@
...
+    fn test_get_pre_1_wallet_keychains() -> anyhow::Result<()> {
+        let mut conn = Connection::open_in_memory()?;
+        let external_checksum = vec![0x01u8, 0x02u8, 0x03u8, 0x04u8];
+        let internal_checksum = vec![0x05u8, 0x06u8, 0x07u8, 0x08u8];
+        // Init tables
+        {
+            // Create pre-1.0 bdk sqlite schema
+            conn.execute_batch(
+                "CREATE TABLE last_derivation_indices (keychain TEXT, value INTEGER);
+                 CREATE UNIQUE INDEX idx_indices_keychain ON last_derivation_indices(keychain);
```
110CodingP: Just a nit, feel free to ignore:
Do we need to create an index on the keychain column?

PR #368, opened 2026-01-16
feat!: add wallet birthday

### LINKED ISSUE #110: Add birthday to the wallet database
**Describe the enhancement**  
Add a method `wallet_birthday` or `birthday` that returns the unix time of when the wallet was created.

**Use case**  
If we know the birthday of a wallet, we can avoid scanning blocks that happened before the wallet was created. If there's no birthday, we must scan the entire chain to be sure that we don't miss txs.

luisschwab:
Closes #110.

TBD

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -74,3 +74,6 @@ name = "esplora_blocking"
 
 [[example]]
 name = "bitcoind_rpc"
+
+[[example]]
+name = "bitcoind_rpc_filter"
--- a/docs/adr/0004_birthday.md
+++ b/docs/adr/0004_birthday.md
@@ -1,0 +1,56 @@
+# Introduce birthday information to the wallet
+
+* Status: TBD
+* Authors: @luisschwab
+* Date: 2026-02-XX
+* Targeted modules: wallet
+* Associated tickets/PRs: [#368](https://github.com/bitcoindevkit/bdk_wallet/pull/368)
+
+## Context and Problem Statement
+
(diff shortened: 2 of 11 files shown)


evanlinjin:
> > I was more in favor of the approach in #348 which provided an option on CreateParams to simply add a checkpoint to the local chain during Wallet creation.
> 
> Indeed, this would be a better flow, but may conflict with a possibly inexistent `BlockId` set during wallet creation. I need to do some testing to see what would happen.

**Some thoughts:**

* Birthdays are only useful for block-based chain sources.
* `bitcoind_rpc` emitters trace back from a no-longer-existing-block to a block that is currently in chain (which is the correct behavior) - so no conflict at all!

The only fear is if the chain source has never even seen the no-longer-existing-block - in which case, the chain source starts syncing from genesis (as that would be the closest checkpoint that connects).

So if the user creates a wallet just before a reorg, that would be non-ideal.

**My suggestion:**

`bitcoind_rpc::Emitter::new` has a `start_height` parameter. This is used in the case the closest block that connects is below the `start_height`. So if the tip is reorged out, and we only have tip + genesis, we will start from `start_height` instead. To convey the wallet birthday with the checkpoint tip, while still taking into account reorgs, we can use `tip_height - FINAL_DEPTH` as the `start_height` parameter. `FINAL_DEPTH` can be as conservative as you want.

`bitcoind_rpc::FilterIter::new` does not have such a parameter. We can add it?

Then we provide docs as how to convey the wallet's birthday to the chain source using the checkpoint tip.

PR #390, opened 2026-03-02
fix(coin_selection): `calculate_cs_result` returns the required UTXOs first

ValuedMammal:
### Description

Follow-up to #262 that addresses transaction input ordering when BnB finds a solution.
    
Previously `calculate_cs_result` produced a CoinSelectionResult by appending the required UTXOs onto the selected ones, which changed the expected order of transaction inputs.

`calculate_cs_result` now returns the required UTXOs before the newly selected ones. This behavior aligns with the expectation that the order of manually selected inputs should be preserved in the final transaction whenever `TxOrdering::Untouched` is specified.

For related discussion refer to https://github.com/bitcoindevkit/bdk_wallet/issues/244#issuecomment-2996583520.

### Changelog notice

Fixed

- wallet: Fixed order of selected UTXOs for `BranchAndBoundCoinSelection`, required UTXOs come first

#### All Submissions:

#### Bugfixes:

### DIFF
--- a/src/wallet/coin_selection.rs
+++ b/src/wallet/coin_selection.rs
@@ -704,16 +704,17 @@ impl CoinSelectionAlgorithm for SingleRandomDraw {
 }
 
 fn calculate_cs_result(
-    mut selected_utxos: Vec<OutputGroup>,
-    mut required_utxos: Vec<OutputGroup>,
+    selected_utxos: Vec<OutputGroup>,
+    required_utxos: Vec<OutputGroup>,
     excess: Excess,
 ) -> CoinSelectionResult {
-    selected_utxos.append(&mut required_utxos);
-    let fee_amount = selected_utxos.iter().map(|u| u.fee).sum();
-    let selected = selected_utxos
+    let mut selected = required_utxos;
+    selected.extend(selected_utxos);
+    let fee_amount = selected.iter().map(|u| u.fee).sum();
+    let selected = selected
         .into_iter()
-        .map(|u| u.weighted_utxo.utxo)
-        .collect::<Vec<_>>();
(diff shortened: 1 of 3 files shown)


Serious-Sam-Dev:
all test and clippy passed, i dont see any problem, ACK

110CodingP (approved):
ACK [`378cb24`](https://github.com/bitcoindevkit/bdk_wallet/pull/390/commits/378cb24231d9e781da76dfdbdc1a7dc594dd9f69)

Though unrelated to the PR I was wondering whether we should also have a test to check the following invariant in `add_utxos` ?
"If a UTXO is inserted multiple times, only the final insertion will take effect."

ValuedMammal:
> why you used .extend() instead of the previously used .append().

It seemed to match the style and ergonomics of the rest of the code base. I think either one is probably fine? Found a related discussion here https://github.com/rust-lang/rust-clippy/issues/4321#issuecomment-1701112204 and here https://users.rust-lang.org/t/pearl-extending-a-vec-via-append-or-extend/73456

KrishnaShuk:
ACK e020391c.

I reviewed the changes locally on a per-commit basis. I conceptually agree with preserving the integration of TxOrdering::Untouched by inserting required UTXOs before Branch and Bound selections. 
I also temporarily reverted the concatenation order locally and verified that the newly added bnb_success test failed as expected, proving the test suite properly covers the edge case. 

Great work.

110CodingP (approved):
ACK [`8a5e763`](https://github.com/bitcoindevkit/bdk_wallet/pull/390/commits/8a5e76309b54443c65d7a89aa36ab2fa9b8d3e3a).

Issue #234, opened 2025-05-19
Downgrade Support


TheBlueMatt:
For most enterprise deployments of software, the ability to roll back a change if things aren't working correctly during a deploy is a critical feature. It seems BDK's current wallet serialization logic doesn't provide this as a hard guarantee to downstream users. It would be nice if this were provided as an explicit supported feature.

notmandatory:
I agree we should support a downgrade policy but we should define more precisely what that means. 

Currently `bdk_wallet` and `bdk_chain` types that are persisted implement the `Merge: Default` trait. The docs for `Merge` could use improvement, currently says "Trait that makes an object mergeable."  In practice because this trait requires `Default` if a new version of persistence code loads old data nothing breaks because newly required data is defaulted. 

But the question here is what if a user downgrades to an old version of persistence code and tries to load and "Merge" data created with a newer version of the persistence code. In this case we need to guarantee the new persistence code didn't:
1. remove or rename struct fields or enum variants expected by the old code
2. change how a field or variant is used in a way that would break old code 
3. add new data that can't be ignored by the old code

I don't know a way to enforce the above as we do with `Default` for forward compatibility. But here's what I think we can do:

- [ ] Document somewhere that all persisted data must support the `Merge` trait
- [ ] Document for `Merge` that new fields and enum variants must not be removed between minor/patch versions
- [ ] Document a process to deprecate and then remove/rename fields/variants in future major versions
- [ ] Add CI process to verify data created with the prior release version can be loaded with a new version
- [ ] Add CI process to verify data created with new version can be loaded with the prior release version

For the deprecate policy I'm thinking we only support roll back for deprecated fields/variants for one major release before removing rollback support for it. This would mean users can upgrade or downgrade between say 1.x and 2.x, but not 1.x and 3.x. Would that be sufficient downgrade support?

PR #391, opened 2026-03-02
docs(changeset): Add section on version compatibility

ValuedMammal:
### Description

Partially addresses #234 by improving the documentation of `ChangeSet` by adding a version compatibility policy, clarifying what upgrade/downgrade paths are supported and how breaking changes should be handled. It also adds the `#[serde(default)]` attribute to `locked_outpoints` which is a newly added field in 3.0.

In summary the policy says:

- Changes to the `ChangeSet` data structure will correspond to a major version bump.
- Version N can read/deserialize data written by N-1, i.e., one major version back, but this guarantee doesn't extend to version N-2 or earlier.
- New fields introduced in version N must implement `Default` so that when deserializing, any missing fields are filled in with the default.
- Version N-1 can deserialize version N data by ignoring unknown fields, with the caveat that not all features from version N will be available.
- Outline a 3-version deprecation cycle

### Changelog notice

Changed

- Added `#[serde(default)]` to `ChangeSet::locked_outpoints`

#### All Submissions:

### DIFF
--- a/src/wallet/changeset.rs
+++ b/src/wallet/changeset.rs
@@ -76,12 +76,42 @@ type IndexedTxGraphChangeSet =
 ///
 /// Existing fields may be extended in the future with additional sub-fields. New top-level fields
 /// are likely to be added as new features and core components are implemented. Existing fields may
-/// be removed in future versions of the library.
+/// be removed in future versions of the library following the deprecation policy below.
 ///
-/// The authors reserve the right to make breaking changes to the [`ChangeSet`] structure in
-/// a major version release. API changes affecting the types of data persisted will display
-/// prominently in the release notes. Users are advised to look for such changes and update their
-/// application accordingly.
+/// ## Version Compatibility
+///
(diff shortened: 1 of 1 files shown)


src/wallet/changeset.rs
```diff
@@ -76,12 +76,42 @@ type IndexedTxGraphChangeSet =
...
+/// be removed in future versions of the library following the deprecation policy below.
 ///
-/// The authors reserve the right to make breaking changes to the [`ChangeSet`] structure in
-/// a major version release. API changes affecting the types of data persisted will display
-/// prominently in the release notes. Users are advised to look for such changes and update their
-/// application accordingly.
+/// ## Version Compatibility
+///
+/// Any change to the [`ChangeSet`] data structure MUST correlate with a major version bump per
+/// [Semantic Versioning](https://semver.org/). We guarantee that version N can read and
```
oleonardolima: nit: maybe it's a good idea to also refer to: https://doc.rust-lang.org/cargo/reference/semver.html (not a blocker though).

PR #392, opened 2026-03-03
fix(tx_builder)!: Remove `include_output_redeem_witness_script`

### LINKED ISSUE #236: Deprecate method `TxBuilder::include_output_redeem_witness_script`
As far as I can tell there is no dedicated logic for handling this TxBuilder option. I guess it may have been made obsolete somewhere along the way.

Dmenec:
### Description

Closes #236

Removes `TxBuilder::include_output_redeem_witness_script` as it currently has no effect.

Originally, this method controlled an explicit block in the transaction creation pipeline that derived the descriptor and populated `psbt::Output::witness_script` and `psbt::Output::redeem_script` for outputs requiring it (e.g. `sh(wsh(...))`). That logic was guarded behind this flag, making it opt-in.

At some point during refactoring, this responsibility was delegated to miniscript.

#### New Features:

* [x] I've added a test to verify that the redeem and witness scritps are still being populated even when the flag is not used. 

#### Workflow:

* [x] Traced all references to `include_output_redeem_witness_script` in the codebase.
* [x] Checked if the existing test `test_include_output_redeem_witness_script` passes with or without calling the method, since the scripts are populated automatically
* [x] Found the original commit that introduced the method, which showed the conditional block that no longer exists in the current code

#### All Submissions:

* [x] I ran `just fmt`, `just check` and `just test` before pushing

### Note on deprecation policy
The contribution guidelines recommend [deprecating APIs before removal](https://github.com/bitcoindevkit/bdk_wallet/blob/master/CONTRIBUTING.md#deprecation-policy). However, given the proximity of the `Wallet 3.0.0` release, it might be reasonable to include this change in that milestone. 

In any case I'll be happy to adjust it if the maintainers prefer following the standard deprectation flow.

### DIFF
--- a/src/wallet/tx_builder.rs
+++ b/src/wallet/tx_builder.rs
@@ -137,7 +137,6 @@ pub(crate) struct TxParams {
     pub(crate) change_policy: ChangeSpendPolicy,
     pub(crate) only_witness_utxo: bool,
     pub(crate) add_global_xpubs: bool,
-    pub(crate) include_output_redeem_witness_script: bool,
     pub(crate) bumping_fee: Option<PreviousFee>,
     pub(crate) current_height: Option<absolute::LockTime>,
     pub(crate) allow_dust: bool,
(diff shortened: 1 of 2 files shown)


ValuedMammal (approved):
ACK 8ca39a79a9f9e263787ba52c852a62d1fdade68c

nymius (approved):
ACK 8ca39a79a9f9e263787ba52c852a62d1fdade68c

As per #236, the logic to handle this option doesn't exist (a no-op), we can skip the deprecation flag this time.

luisschwab (approved):
ACK 8ca39a79a9f9e263787ba52c852a62d1fdade68c

PR #423, opened 2026-03-28
fix(wallet)!: fail early for unregistered foreign prevouts in add_foreign_utxo

### LINKED ISSUE #345: Inform wallet about foreign transactions/txouts before using them to fund new transactions
**Description**

> This is a follow up of #337 and #325 

`TxBuilder::add_foreign_utxo` allows the construction of transactions with UTXOs not belonging  to the wallet.
Trying to RBF those transactions up to #337 failed.
#337 is relying on the manual addition of the UTXOs to the transaction graph.
The wallet is not going to list them them as `LocalUtxos` because the `KeychainTxOutIndex` filters out any outpoint not related to a script pubkey produced by the `KeyChain`.

Something that may be needed though in `Wallet` is a method to insert transactions, as the only available now is `insert_txout` but for non segwit foreign UTXOs the full transaction will be needed at the moment of building the tx RBF.

To address the original issue from the perspective of `TxBuilder`, I had three options so far:
1. Calling `Wallet.insert_txout` from `TxBuilder.add_foreign_utxo` to make this less manual, but I think is going to make the API much more entangled to what I would like.
2. Document the workflow rather than embedding the logic into `TxBuilder`.
3. Add a method to list non local UTXOs to wallet or tx graph and error in `TxBuilder.add_foreign_utxo` if the UTXO is not listed there, forcing the TxBuilder API user to call `Wallet.insert_transaction` or `Wallet.insert_txout` before calling `TxBuilder.add_foreign_utxo`.

I think the third one is preferable, but I'm writing this issue to gather feedback and see if there are any concerns I"m overlooking.

noahjoeris:
### Description

This PR makes `TxBuilder::add_foreign_utxo` fail early when the txout/tx required by the PSBT input has not been inserted into the wallet.

- use `Wallet::insert_txout` for `witness_utxo` inputs
- use `Wallet::insert_tx` for `non_witness_utxo` inputs

This avoids building transactions that later fail during fee bumping because the wallet cannot reconstruct the required PSBT input metadata from its tx graph.

This PR also adds `Wallet::insert_tx`, introduces new registration errors in `add_foreign_utxo`, and updates the tests.

Fixes #345

### Notes to the reviewers
- The main issue here is fee bumping for transactions funded with foreign inputs. `build_fee_bump` reconstructs PSBT input metadata from the wallet tx graph. That works if the required foreign data is already in the graph, but it can fail for foreign inputs when the wallet was not informed about the parent transaction or txout ahead of time.

- This is relevant for foreign inputs that rely on `non_witness_utxo`, where the full parent transaction must be available. `insert_txout` alone is not enough in that case, which is why this PR adds `Wallet::insert_tx`.

- I chose to make `add_foreign_utxo` fail early if the transaction or txout has not been inserted into the wallet. This follows the direction discussed in #345.

- This is my first PR to the codebase 🔥 I’d appreciate any feedback on the approach and I’m happy to make follow-up changes.

### Changelog notice

`TxBuilder::add_foreign_utxo` now fails early if the transaction or txout required by the provided PSBT input has not already been inserted into the wallet. Use `Wallet::insert_txout` for `witness_utxo` inputs and `Wallet::insert_tx` for `non_witness_utxo` inputs.

#### All Submissions:

#### Bugfixes:

### DIFF
--- a/src/wallet/tx_builder.rs
+++ b/src/wallet/tx_builder.rs
@@ -366,9 +366,12 @@ impl<'a, Cs> TxBuilder<'a, Cs> {
     ///
     /// This is an **EXPERIMENTAL** feature, API and other major changes are expected.
     ///
-    /// In order to use [`Wallet::calculate_fee`] or [`Wallet::calculate_fee_rate`] for a
-    /// transaction created with foreign UTXO(s) you must manually insert the corresponding
-    /// TxOut(s) into the tx graph using the [`Wallet::insert_txout`] function.
+    /// Before calling this method, you must preregister the foreign prev data in the wallet tx
+    /// graph. Use [`Wallet::insert_txout`] if `psbt_input` only provides a `witness_utxo`, or
+    /// [`Wallet::insert_tx`] if it provides a `non_witness_utxo`.
+    ///
+    /// This preregistration is also what allows [`Wallet::calculate_fee`] or
+    /// [`Wallet::calculate_fee_rate`] to work for transactions created with foreign UTXO(s).
     ///
     /// # Errors
     ///
@@ -376,6 +379,7 @@ impl<'a, Cs> TxBuilder<'a, Cs> {
     ///
     /// 1. The `psbt_input` does not contain a `witness_utxo` or `non_witness_utxo`.
     /// 2. The data in `non_witness_utxo` does not match what is in `outpoint`.
+    /// 3. The wallet tx graph does not already contain the prev data required by `psbt_input`.
     ///
     /// Note unless you set [`only_witness_utxo`] any non-taproot `psbt_input` you pass to this
     /// method must have `non_witness_utxo` set otherwise you will get an error when [`finish`]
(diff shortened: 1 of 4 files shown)


tests/add_foreign_utxo.rs
```diff
@@ -93,9 +93,11 @@ fn test_add_foreign_utxo_invalid_psbt_input() {
         .unwrap();
 
     let mut builder = wallet.build_tx();
-    let result =
-        builder.add_foreign_utxo(outpoint, psbt::Input::default(), foreign_utxo_satisfaction);
-    assert!(matches!(result, Err(AddForeignUtxoError::MissingUtxo)));
+    let err = builder
+        .add_foreign_utxo(outpoint, psbt::Input::default(), foreign_utxo_satisfaction)
+        .unwrap_err();
+    assert!(!err.to_string().is_empty());
```
sdmg15, line 99: I'd be in favor of keeping the diff minimal and just add the line 
`assert!(!result.as_ref().unwrap_err().to_string().is_empty());`

tests/add_foreign_utxo.rs
```diff
@@ -112,16 +114,19 @@ fn test_add_foreign_utxo_requires_inserted_txout() {
...
+                utxo.outpoint,
+                psbt::Input {
+                    witness_utxo: Some(utxo.txout.clone()),
+                    ..Default::default()
+                },
+                foreign_utxo_satisfaction,
+            )
+            .unwrap_err();
+        assert!(!err.to_string().is_empty());
         assert!(
```
sdmg15, line 128: Same here we could keep the diff minimal.

sdmg15 (commented):
I've done a first round of review and haven't noticed any major issues. Left few comments on commit https://github.com/bitcoindevkit/bdk_wallet/pull/423/changes/ced965e316a3452a6f6d6f4494726817a362a21a

The third point in the associated issue, suggests adding Add a method to list non local UTXOs. I haven't seen that here.

src/wallet/tx_builder.rs
```diff
@@ -407,25 +411,32 @@ impl<'a, Cs> TxBuilder<'a, Cs> {
...
+                    input_txid: tx.compute_txid(),
+                    foreign_utxo: outpoint,
+                });
+            }
+            if tx.output.len() <= outpoint.vout as usize {
+                return Err(AddForeignUtxoError::InvalidOutpoint(outpoint));
             }
         }
 
+        if psbt_input.non_witness_utxo.is_some() {
```
nymius: We are already checking `psbt_input.non_witness_utxo` is some above. What's the reason to split the conditional?

noahjoeris: The rebase fixed it

PR #427, opened 2026-04-01
feat(wallet): add configurable min_output_value filter

### LINKED ISSUE #375: Add a global dust filter setting, idealy configurable amount
Add a configurable dust threshold to Wallet that filters outputs below a specified satoshi value from balance calculations, sent_and_received(), and UTXO iteration methods.

  Currently, to exclude dust outputs (e.g., < 1000 sats) from wallet operations, users must:
  1. Manually iterate list_unspent() and filter by value
  2. Reimplement BDK's internal logic for categorizing UTXOs into confirmed, trusted_pending, untrusted_pending, and immature
  3. Manually calculate sent_and_received() by iterating transaction inputs/outputs and looking up previous outputs

  This is error-prone and requires duplicating internal BDK logic that may change between versions.

  Proposed API

  // Option 1: Builder pattern
  let wallet = Wallet::builder()
      .descriptor(descriptor)
      .change_descriptor(change_descriptor)
      .dust_threshold(Amount::from_sat(1000))
      .network(Network::Bitcoin)
      .build()?;

  // Option 2: Runtime configuration
  wallet.set_dust_threshold(Some(Amount::from_sat(1000)));

  // These would now respect the threshold:
  let balance = wallet.balance();  // Excludes UTXOs < 1000 sats
  let (sent, received) = wallet.sent_and_received(&tx);  // Excludes dust outputs
  let utxos = wallet.list_unspent();  // Excludes dust UTXOs

  Use case

  We operate a multi-sig custody platform where dust outputs create operational overhead:
  - Users see confusing small amounts in their balance
  - Transaction history shows spam/dust attack outputs
  - Consolidating dust costs more in fees than the outputs are worth
  - Affects user privacy by triggering the common ownership heuristic

  We need consistent dust filtering across all wallet operations. Currently we're forced to bypass wallet.balance() and sent_and_received() entirely, manually reimplementing the UTXO categorization logic.

  Impact
  - Nice-to-have / UX improvement
  - Developer experience / maintainability

  Are you using BDK in a production project?
  - Yes

  Which backend(s) are relevant (if any)?
  - Electrum
  - Esplora
  - Bitcoin Core RPC

  Project or organization (optional)

  AnchorWatch - Bitcoin custody platform

  Additional context

  The 546 sat dust limit is a relay policy, but many applications want higher thresholds (1000+ sats) for UX reasons. A configurable threshold at the wallet level would be more ergonomic than forcing every consumer
  to reimplement filtering logic.

noahjoeris:
### Description

Adds a runtime-only `min_output_value: Option<Amount>` setting to `Wallet` for dust/spam filtering. It filters wallet-visible spendable value in `balance`, `list_unspent`, and automatic coin selection without changing accounting functions.

Closes #375

### Notes to the reviewers

- Named `min_output_value` rather than `dust_threshold` to avoid confusion with the existing coin selection `dust_threshold` and Bitcoin's relay dust policy.
- Not persisted; treated as a runtime filter applied on top of stored wallet state. It can be provided when creating/loading via `CreateParams::min_output_value()` / `LoadParams::min_output_value()`, or changed at runtime via `Wallet::set_min_output_value()`.
- Filtering is applied in `bdk_wallet`, not pushed down into `bdk_chain`.

### Changelog notice

- Added runtime `min_output_value` support to `Wallet`, `CreateParams`, and `LoadParams`
- `balance`, `list_unspent`, and coin selection now honor `min_output_value`

#### All Submissions:

#### New Features:

### DIFF
--- a/src/wallet/mod.rs
+++ b/src/wallet/mod.rs
@@ -117,6 +117,7 @@ pub struct Wallet {
     network: Network,
     secp: SecpCtx,
     locked_outpoints: HashSet<OutPoint>,
+    min_output_value: Option<Amount>,
 }
 
 /// An update to [`Wallet`].
@@ -374,6 +375,7 @@ impl Wallet {
             stage,
             secp,
             locked_outpoints,
+            min_output_value: params.min_output_value,
         })
     }
 
@@ -581,12 +583,23 @@ impl Wallet {
             network,
             secp,
             locked_outpoints,
+            min_output_value: params.min_output_value,
         }))
     }
 
     /// Get the [`Network`] the wallet is using.
     pub fn network(&self) -> Network {
         self.network
+    }
+
+    /// Get the minimum output value, if set.
+    pub fn min_output_value(&self) -> Option<Amount> {
+        self.min_output_value
+    }
+
+    /// Set or clear the minimum output value.
+    pub fn set_min_output_value(&mut self, min_output_value: Option<Amount>) {
+        self.min_output_value = min_output_value;
     }
 
     /// Iterator over all keychains in this wallet
@@ -777,7 +790,11 @@ impl Wallet {
     }
 
     /// Return the list of unspent outputs of this wallet
+    ///
+    /// If [`min_output_value`](Wallet::min_output_value) is set, outputs with a value below
+    /// that threshold are excluded.
     pub fn list_unspent(&self) -> impl Iterator<Item = LocalOutput> + '_ {
+        let min_value = self.min_output_value;
         self.tx_graph
             .graph()
             .filter_chain_unspents(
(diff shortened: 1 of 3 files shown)


Dmenec:
Concept ACK. Related with the `sent_and_received` / `net_value` approach, why apply `min_output_value` to transaction accounting methods? A tx already broadcast moved exactly what it moved on-chain. I'd expect the threshold to act as an operational filter for UTXO management, but not change how historical transactions are reported. Maybe I'm ignoring some use case...

Also, #431 rewrites `balance()` for trusted/untrusted pending classification, worth a look as both touch the same function.

noahjoeris:
Yes agree, thanks. I think I included it because #375 asked for it too. Bit weird in hindsight and can cause inconsistencies. So the key parts to filter are `list_unspent`, `balance`, and coin selection. I updated in latest push.
Will rebase once your PR lands

PR #428, opened 2026-04-01
Persist cached SPKs on wallet creation

### LINKED ISSUE #387: Wallet creation should persist cached SPKs
Right now we have to reveal an address before the cache takes over, [...] should wallet creation on its own persist an initial changeset that includes the lookahead scripts?

_Originally posted by @ValuedMammal in https://github.com/bitcoindevkit/bdk_wallet/pull/288#discussion_r2251803638_

j-kon:
## Summary
- flush the indexer's staged SPK cache during wallet creation so persisted wallets write the initial lookahead cache immediately
- avoid emitting empty staged cache entries when reloading wallets with persisted SPK cache enabled
- add a regression test proving a wallet created with use_spk_cache(true) does not need an initial reveal before the cache persists

Closes #387.

## Testing
- cargo test --all-features --test persisted_wallet
- cargo test --all-features

cc @ValuedMammal, @oleonardolima

### DIFF
--- a/src/wallet/mod.rs
+++ b/src/wallet/mod.rs
@@ -355,15 +355,15 @@ impl Wallet {
             ..Default::default()
         };
 
-        let tx_graph = make_indexed_graph(
-            &mut stage,
-            Default::default(),
+        let index = make_keychain_index(
             Default::default(),
             descriptor,
             change_descriptor,
             params.lookahead,
             params.use_spk_cache,
         )?;
+        stage.indexer.merge(index.initial_changeset());
+        let tx_graph = IndexedTxGraph::new(index);
 
         Ok(Wallet {
             signers,
@@ -559,10 +559,9 @@ impl Wallet {
             .map(|(op, _)| op)
             .collect();
 
-        let mut stage = ChangeSet::default();
+        let stage = ChangeSet::default();
 
-        let tx_graph = make_indexed_graph(
-            &mut stage,
+        let tx_graph = make_loaded_indexed_graph(
             changeset.tx_graph,
             changeset.indexer,
             descriptor,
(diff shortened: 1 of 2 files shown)


ValuedMammal:
The main issue is with the `spk_cache_stage` AFAICT, because it alters the `ChangeSet` semantics in subtle and surprising ways. It's not clear to the developer why indexing a null txout would be necessary to "flush the stage".

The SPK cache already acts as a kind of stage. What we want is to pass a `&mut ChangeSet` directly to `replenish_inner_index` and then ideally return it from `insert_descriptor` (or similar). At the very least we can just stage the indexer `initial_changeset` when creating a `Wallet`. Then `reveal_next_address` will only persist 1 newly derived SPK at a time.  

Second, `make_indexed_graph` appears to be doing double duty by constructing the `IndexedTxGraph` from a changeset during both the creation and loading phases. This is unnecessary since `create_with_params` just passes an empty changeset. Further, we already assert that the [stage must be empty upon loading](https://github.com/bitcoindevkit/bdk_wallet/blob/fb7681ab83634dd7cc71b4908b385e38b8db6bef/tests/persisted_wallet.rs#L165), so what is the intention of the `&mut stage` in `make_indexed_graph` 🤔

j-kon:
Thanks, this makes sense.

I reworked the patch to follow that direction and pushed an update. Instead of forcing the staged SPK cache out via `index_txout`, wallet creation now stages `index.initial_changeset()` directly, so the initial lookahead cache is persisted explicitly at creation time.

I also split the creation and loading paths: creation now builds the index directly and stages its initial changeset, while loading reconstructs the `IndexedTxGraph` from the persisted changeset without producing new staged changes. That removes the `&mut stage` coupling from the load path and keeps the "stage is empty upon load" invariant intact.

The regression test is still covering the original issue: a wallet created with `use_spk_cache(true)` can be reloaded immediately, and the first `reveal_next_address` only stages the newly derived SPK rather than restaging the initial cache.

I reran `cargo test --all-features --test persisted_wallet` and `cargo test --all-features` after the refactor.

ValuedMammal:
> a wallet created with use_spk_cache(true) can be reloaded immediately, and the first reveal_next_address only stages the newly derived SPK rather than restaging the initial cache.

I like the approach, but seems to be blocked on a resolution to bitcoindevkit/bdk#2217. If that were fixed we wouldn't need the create-and-reload-immediately workaround.

PR #430, opened 2026-04-03
fix(wallet): allow fee bump without opt-in RBF signaling

### LINKED ISSUE #65: Allow bump fee on transactions without RBF sequence
This should be an easy change, the following lines need to be removed from `Wallet::build_fee_bump`:
https://github.com/bitcoindevkit/bdk/blob/7969898c732a3a387a932a61b1719cedcb6fae3c/crates/wallet/src/wallet/mod.rs#L1636-L1644

> Also related to this discussion (probably warrants a separate issue?) is how the new rules impact the bump fee transaction builder. I'm not sure if the current builder requires a transaction signals for RBF to then be used by the replace by fee txbuilder?

_Originally posted by @thunderbiscuit in https://github.com/bitcoindevkit/bdk/issues/1653#issuecomment-2422975075_

alienx5499:
### Description

Fixes #65.

`Wallet::build_fee_bump` previously refused to build a replacement when **no** input signaled opt-in RBF (`nSequence` ≤ `0xFFFFFFFD`). That guard is removed so the wallet can **construct** a fee-bump transaction even if the original did not opt in. Whether a replacement is accepted is a **mempool / relay policy** concern (e.g. default full-RBF in recent Bitcoin Core), not something this API can enforce by requiring a specific `nSequence` on the **prior** tx.

Also removes `BuildFeeBumpError::IrreplaceableTransaction`, since that path is no longer reachable.

### Notes to the reviewers

- **API:** `BuildFeeBumpError::IrreplaceableTransaction` is removed (breaking public enum). If you prefer a deprecation cycle for the variant only, say so and we can adjust.
- **Semantics:** Callers could previously rely on an error when the original tx did not signal opt-in RBF; they must not assume that anymore.
- **Tests:** `tests/build_fee_bump.rs` — the old test expected failure for non–opt-in-RBF; it now asserts a successful fee bump with a higher fee rate.

### Changelog notice

`fix(wallet): allow build_fee_bump without opt-in RBF on the original tx; remove IrreplaceableTransaction (#65)`

#### All Submissions:

#### New Features:

#### Bugfixes:

### DIFF
--- a/src/wallet/mod.rs
+++ b/src/wallet/mod.rs
@@ -1551,9 +1551,14 @@ impl Wallet {
 
     /// Bump the fee of a transaction previously created with this wallet.
     ///
-    /// Returns an error if the transaction is already confirmed or doesn't explicitly signal
-    /// *replace by fee* (RBF). If the transaction can be fee bumped then it returns a [`TxBuilder`]
-    /// pre-populated with the inputs and outputs of the original transaction.
+    /// Returns an error if the transaction is already confirmed. If the transaction can be fee
+    /// bumped then it returns a [`TxBuilder`] pre-populated with the inputs and outputs of the
+    /// original transaction.
+    ///
+    /// Replacing an unconfirmed transaction does not require the original to signal opt-in RBF
+    /// (`nSequence` ≤ `0xFFFFFFFD`). Relay and mining acceptance depend on local policy; Bitcoin
+    /// Core 28+ defaults to full-RBF mempool relay, while other implementations or configs may
+    /// differ.
     ///
     /// ## Example
     ///
@@ -1614,16 +1619,6 @@ impl Wallet {
             .is_confirmed()
         {
             return Err(BuildFeeBumpError::TransactionConfirmed(txid));
-        }
-
-        if !tx
-            .input
-            .iter()
-            .any(|txin| txin.sequence.to_consensus_u32() <= 0xFFFFFFFD)
-        {
-            return Err(BuildFeeBumpError::IrreplaceableTransaction(
-                tx.compute_txid(),
-            ));
         }
 
         let fee = self
--- a/src/wallet/mod.rs
+++ b/src/wallet/mod.rs
@@ -1589,9 +1589,13 @@ impl Wallet {
 
     /// Bump the fee of a transaction previously created with this wallet.
     ///
-    /// Returns an error if the transaction is already confirmed or doesn't explicitly signal
-    /// *replace by fee* (RBF). If the transaction can be fee bumped then it returns a [`TxBuilder`]
-    /// pre-populated with the inputs and outputs of the original transaction.
+    /// Returns an error if the transaction is already confirmed. If the transaction can be fee
+    /// bumped then it returns a [`TxBuilder`] pre-populated with the inputs and outputs of the
+    /// original transaction.
+    ///
+    /// This builder does not require the original transaction to signal opt-in RBF
+    /// (`nSequence` <= `0xFFFFFFFD`). Whether the resulting replacement is accepted
+    /// for relay or mining depends on the local policy of the broadcasting node.
     ///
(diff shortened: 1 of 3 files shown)


alienx5499:
Hi @ValuedMammal, gentle ping for review when you have time. Thanks!

ValuedMammal:
Approach ACK

src/wallet/mod.rs
```diff
@@ -1551,9 +1551,14 @@ impl Wallet {
...
-    /// pre-populated with the inputs and outputs of the original transaction.
+    /// Returns an error if the transaction is already confirmed. If the transaction can be fee
+    /// bumped then it returns a [`TxBuilder`] pre-populated with the inputs and outputs of the
+    /// original transaction.
+    ///
+    /// Replacing an unconfirmed transaction does not require the original to signal opt-in RBF
+    /// (`nSequence` ≤ `0xFFFFFFFD`). Relay and mining acceptance depend on local policy; Bitcoin
+    /// Core 28+ defaults to full-RBF mempool relay, while other implementations or configs may
+    /// differ.
     ///
```
yan-pi: Could we make this specific to the builder? The first sentence reads like a general replacement rule.

Maybe:

```suggestion
/// This builder does not require the original transaction to signal opt-in RBF
/// (`nSequence` <= `0xFFFFFFFD`). Whether the resulting replacement is accepted
/// for relay or mining depends on the local policy of the broadcasting node.
```

tests/build_fee_bump.rs
```diff
@@ -16,19 +16,39 @@ mod common;
...
+        .expect("fee bump without opt-in RBF");
+    builder.fee_rate(feerate);
+    let psbt = builder.finish().expect("finish fee bump");
+    let fee = check_fee!(wallet, psbt);
+    assert!(
+        fee > original_fee,
+        "fee bump must increase absolute fee: {fee} > {original_fee}"
+    );
+
+    let new_tx = psbt.clone().extract_tx().expect("failed to extract tx");
```
yan-pi, line 50: Would it be worth asserting the two properties this test is covering: that the original does not explicitly signal RBF, and that the replacement shares a prevout with it?

The current assertions only check that the fee increased and the txid changed.

Maybe something like this before moving tx into the wallet:

```rust
let tx = psbt.extract_tx().expect("failed to extract tx");
assert!(!tx.is_explicitly_rbf());
let original_prevouts = tx
    .input
    .iter()
    .map(|txin| txin.previous_output)
    .collect::<Vec<_>>();
let txid = tx.compute_txid();
insert_tx(&mut wallet, tx);
```

And this after extracting `new_tx`:
```rust
assert!(
    new_tx
        .input
        .iter()
        .any(|txin| original_prevouts.contains(&txin.previous_output)),
);
```

yan-pi (commented):
Approach ACK

Removing the sequence check makes sense to me. `build_fee_bump` constructs the replacement, while acceptance depends on the node's mempool policy.

The new test passes on db7feac0:

```
cargo test --test build_fee_bump test_bump_fee_tx_without_rbf_signaling
```

`just check` on the PR branch hits the deprecations fixed in #455. To make sure those were unrelated to this change, I applied the diff on top of fc88144 and ran:

```
cargo test --test build_fee_bump
just check test doc
```

Both pass. But could you rebase the PR onto the current master?

In my local Core 30 regtest, a non-RBF original was replaced successfully with full-RBF enabled.

I left two small comments on the docs and test. Neither is blocking.

alienx5499:
Thanks for the review @yan-pi ! I've updated the docstring in `build_fee_bump` as suggested and added the assertions to the test to verify !is_explicitly_rbf() and that the replacement shares a prevout. I've also rebased the branch onto the latest master.

yan-pi:
tACK 4955feaa

Thanks, both comments are addressed.
Everything passes.

Issue #16, opened 2025-03-31
wallet::Balance::trusted_pending AND TxBuilder::only_spend_change are broken


stevenroose:
**Describe the bug**  
`TxGraph` calculates balance and splits the unconfirmed txs into "trusted" and "untrusted" based on a `trust_predicate` (`trust_predicate: impl FnMut(&OI, ScriptBuf) -> bool`). `Wallet` then uses that method and fills this argument with
```rust
            |&(k, _), _| k == KeychainKind::Internal,
```

This is too naive. While the `Internal` keychain is **intended** for change that the wallet controls, there is no guarantee that external users don't pay to change addresses (for example by observing the chain and discovering our change addresses). Hence, utxos that are considered `trusted` by the `Wallet::balance` method, can still be maliciously double-spent by external parties.

Similarly, `only_spend_change` also only checks the keychainkind of the output which does not guarantee it was actually change from one of our own txs:
```rust
impl ChangeSpendPolicy {
    pub(crate) fn is_satisfied_by(&self, utxo: &LocalOutput) -> bool {
        match self {
            ChangeSpendPolicy::ChangeAllowed => true,
            ChangeSpendPolicy::OnlyChange => utxo.keychain == KeychainKind::Internal,
            ChangeSpendPolicy::ChangeForbidden => utxo.keychain == KeychainKind::External,
        }
    }
}
```

While this `trust_predicate` seems useful, I think a "trusted" notion is confusing. In effect, the intention is already to mean "change", so I would just drop the "trusted" in favor of "change" (or equivalent the internal/external naming) (`pending_internal`, `pending_external`) or so. BUT, "internal" does not merely mean _from the internal keychain_, but actually "originating from a tx that we ourselves have created".

Once that notion is fixed, and change it not just determined based on keychainkind.

notmandatory:
Moved to `bdk_wallet` repo.

ValuedMammal:
My 2 sats after reading this and the related #273

`trust_predicate` should take into account the outpoint, not just the indexed SPK, since an indexed SPK can occur on multiple transactions which may or not be trusted. In contrast, knowledge of the outpoint means you can apply trust to all of the txouts of a transaction regardless of the associated keychain.

PR #431, opened 2026-04-04
refactor(wallet): derive balance trust from output ancestry

### LINKED ISSUE #273: Balance categorization: wallet-generated transactions to external addresses incorrectly marked as   untrusted_pending
**Describe the bug**
  BDK incorrectly categorizes unconfirmed UTXOs sent to external/receive addresses as
  `untrusted_pending` even when the transaction is generated by the wallet itself (spending wallet's own
   UTXOs to wallet's own addresses). According to BDK documentation, `trusted_pending` should include
  "Unconfirmed UTXOs generated by a wallet tx", but the current implementation only considers outputs to
   internal/change addresses as trusted, regardless of transaction origin.

  **To Reproduce**
  1. Create a wallet with both external (receive) and internal (change) addresses
  2. Fund the wallet with confirmed UTXOs
  3. Create a consolidation transaction that spends wallet UTXOs to a wallet-owned external address
  (using raw transaction APIs like Bitcoin Core's `createrawtransaction`)
  4. Sync the wallet and check balance categorization
  5. Repeat the same consolidation but sending to an internal/change address

  **Expected behavior**
  Both consolidation transactions should result in `trusted_pending` UTXOs since they are generated by
  the wallet (spending wallet's own UTXOs to wallet's own addresses). The destination keychain (external
   vs internal) should not affect trust classification when the wallet created the transaction.

  **Actual behavior:**

  **Case 1: Consolidation to External/Receive Address**
  🔄 Consolidating 2 UTXOs ...
     💰 Total: 0.02000000 BTC → 0.01990000 BTC (0.0001 BTC fee)
     🎯 Consolidating to: bcrt1qghklxanlt9wnjxuqdgfuzcg66n28rw0m3wqjl4

  Address generated with: `bitcoin-cli getnewaddress` (external keychain)
  Address appears in BDK logs as: `External 33: bcrt1qghklxanlt9wnjxuqdgfuzcg66n28rw0m3wqjl4`

  BDK Balance Result:

  | Wallet | Before | After | Diff |
  | ------------- | ------------- | ------------- | ------------- |
  | Trusted pending |  |  |  |
  | Unconfirmed pending |  | 0.01990000 BTC | +0.01990000 BTC ❌ |
  | Confirmed | 0.99970000 BTC | 0.97970000 BTC | -0.02000000 BTC |
  | **Total** | **0.99970000 BTC** | **0.99960000 BTC** | **-0.00010000 BTC** |

  **Case 2: Consolidation to Internal/Change Address**
  🔄 Consolidating UTXOs...
     💰 Total: 0.02000000 BTC → 0.01990000 BTC (0.0001 BTC fee)
     🎯 Consolidating to: bcrt1q293wv26t0s2242vec2jr5jwx63fezwshu29agy

  Address generated with: `bitcoin-cli getrawchangeaddress` (internal keychain)
  Address appears in BDK logs as: `Internal 0: bcrt1q293wv26t0s2242vec2jr5jwx63fezwshu29agy`

  BDK Balance Result:

  | Wallet | Before | After | Diff |
  | ------------- | ------------- | ------------- | ------------- |
  | Trusted pending |  | 0.01990000 BTC | +0.01990000 BTC ✅ |
  | Unconfirmed pending |  |  |  |
  | Confirmed | 0.99980000 BTC | 0.97980000 BTC | -0.02000000 BTC |
  | **Total** | **0.99980000 BTC** | **0.99970000 BTC** | **-0.00010000 BTC** |

  **Build environment**
  - BDK tag/commit: bdk_wallet = "2"
  - OS+version: macOS (Darwin 24.5.0)
  - Rust/Cargo version: Latest stable
  - Rust/Cargo target: x86_64-apple-darwin

  **Additional context**
  This relates to the existing issue #16 about naive trust classification. The current implementation
  determines trust based on the destination keychain (internal vs external) rather than whether the
  wallet created the transaction by spending its own UTXOs.

  In both test cases above, the wallet is spending its own confirmed UTXOs to create new unconfirmed
  UTXOs at wallet-owned addresses. According to the BDK documentation, these should be `trusted_pending`
   since they are "unconfirmed UTXOs generated by a wallet tx." However, BDK only considers them trusted
   when sent to internal/change addresses, creating inconsistent behavior where identical
  wallet-generated transactions are categorized differently based solely on the destination address
  type.

  This inconsistency could lead to incorrect balance reporting and potentially unsafe spending decisions
   in applications that rely on the trust categorization to determine spendable balances.

Dmenec:
Fixes #273, #16. 

### Description

This PR originally fixed the balance categorization with a self-contained ancestry walk implemented in wallet.
Superseded by another walk implementation built on top of bdk_chain's `CanonicalView` (bitcoindevkit/bdk#2246) so the walk is delegated to chain and we only select the behaviour closure in wallet. 

Uses the new `does_taint` predicate of the bdk_chain `classify_outpoints` API. The wallet passes a closure that marks a `CanonicalTx` as tainted when any of its inputs spends an output we don't own (foreign or unknown). Transitivity is handled by `CanonicalView`, which applies the predicate across the unconfirmed ancestry, so an output ends up untrusted if it, or any of its unconfirmed ancestors, is tainted.

### Notes to reviewers

This PR does not use the bdk_chain `balance` API, it folds `classify_outpoints` directly in the wallet instead. `CanonicalView::balance` is just a thin fold over `classify_outpoints`, so the result is the same, but doing the fold here lets the wallet layer its own categories (e.g. locked or reserved coins) on top of the chain-level `Eligibility` without touching bdk_chain.

This follows @evanlinjin's idea of making `classify_outpoints` the primary balance API in bdk_chain (https://github.com/bitcoindevkit/bdk/pull/2235#issue-4724998049).

#### All Submissions:

#### Bugfixes:
* [x] I've added tests to reproduce the issue

### DIFF
--- a/src/wallet/mod.rs
+++ b/src/wallet/mod.rs
@@ -1114,14 +1114,107 @@ impl Wallet {
 
     /// Return the balance, separated into available, trusted-pending, untrusted-pending, and
     /// immature values.
+    ///
+    /// A pending UTXO is `trusted_pending` if every unconfirmed ancestor transaction  
+    /// was created by this wallet.
+    /// Otherwise it is `untrusted_pending`.
     pub fn balance(&self) -> Balance {
-        self.tx_graph.graph().balance(
+        let graph = self.tx_graph.graph();
+        let chain_tip = self.chain.tip().block_id();
+
+        // TODO: simplify with CanonicalView::txout() once released on bdk_chain
+        let canonical_txs: HashMap<Txid, (Arc<Transaction>, ChainPosition<ConfirmationBlockTime>)> =
+            graph
+                .list_canonical_txs(&self.chain, chain_tip, CanonicalizationParams::default())
+                .map(|ctx| {
+                    (
+                        ctx.tx_node.txid,
+                        (ctx.tx_node.tx.clone(), ctx.chain_position),
+                    )
+                })
+                .collect();
+
+        let mut immature = Amount::ZERO;
+        let mut trusted_pending = Amount::ZERO;
+        let mut untrusted_pending = Amount::ZERO;
+        let mut confirmed = Amount::ZERO;
+
+        for (_spk_i, txout) in graph.filter_chain_unspents(
             &self.chain,
-            self.chain.tip().block_id(),
+            chain_tip,
             CanonicalizationParams::default(),
             self.tx_graph.index.outpoints().iter().cloned(),
-            |&(k, _), _| k == KeychainKind::Internal,
-        )
+        ) {
+            match &txout.chain_position {
+                ChainPosition::Confirmed { .. } => {
+                    // TODO: Check with min_confirmation
+                    if txout.is_confirmed_and_spendable(chain_tip.height) {
+                        confirmed += txout.txout.value;
+                    } else if !txout.is_mature(chain_tip.height) {
+                        immature += txout.txout.value;
+                    }
+                }
+
+                ChainPosition::Unconfirmed { .. } => {
+                    let mut txids: Vec<Txid> = vec![txout.outpoint.txid];
+                    let mut visited_tx: HashSet<Txid> = HashSet::new();
+                    let mut trusted: bool = true;
+
+                    'trust_check: while let Some(current_txid) = txids.pop() {
+                        if !visited_tx.contains(&current_txid) {
+                            let Some((tx, _)) = canonical_txs.get(&current_txid) else {
+                                trusted = false;
+                                break;
+                            };
+
+                            for input in tx.input.iter() {
+                                if !&input.previous_output.is_null() {
+                                    let Some((parent_tx, parent_chain_pos)) =
+                                        canonical_txs.get(&input.previous_output.txid)
+                                    else {
+                                        trusted = false;
+                                        break 'trust_check;
+                                    };
+                                    // TODO: Check with min_confirmation
+                                    if !parent_chain_pos.is_confirmed() {
+                                        let vout = input.previous_output.vout as usize;
+                                        let is_ours = parent_tx
+                                            .output
+                                            .get(vout)
+                                            .map(|o| {
+                                                self.tx_graph
+                                                    .index
+                                                    .index_of_spk(o.script_pubkey.clone())
+                                                    .is_some()
+                                            })
+                                            .unwrap_or(false);
+                                        if is_ours {
+                                            txids.push(input.previous_output.txid);
+                                        } else {
+                                            trusted = false;
+                                            break 'trust_check;
+                                        }
+                                    }
+                                }
+                            }
+                            visited_tx.insert(current_txid);
+                        }
+                    }
+                    if trusted {
+                        trusted_pending += txout.txout.value;
+                    } else {
+                        untrusted_pending += txout.txout.value;
+                    }
+                }
+            }
+        }
+
+        Balance {
+            immature,
+            trusted_pending,
+            untrusted_pending,
+            confirmed,
+        }
     }
 
     /// Add an external signer
--- a/src/wallet/mod.rs
+++ b/src/wallet/mod.rs
@@ -1151,22 +1151,24 @@ impl Wallet {
         txs
     }
 
-    /// Return the balance, separated into available, trusted-pending, untrusted-pending, and
-    /// immature values.
+    /// Return the balance, separated into available, trusted-pending, untrusted-pending, and immature values.
     ///
     /// A pending output is trusted only when its entire unconfirmed ancestry spends coins we own.
     /// If any unconfirmed ancestor pulls in a foreign or unknown output, the output is untrusted.
     ///
-    // NOTE: depends on `CanonicalView` (bitcoindevkit/bdk#2246), not yet in a published
-    // `bdk_chain` release.
-    pub fn balance(&self) -> Balance {
+    /// # Arguments
+    /// 
+    /// * `min_confirmations` - How many confirmations an output needs to count as settled. `0` and `1` behave identically. It defines the `is_settled` predicate that bdk_chain's `classify_outpoints` uses to draw the confirmed/pending boundary.
+    /// 
+    // NOTE: depends on `CanonicalView` (bitcoindevkit/bdk#2246), not yet in a published `bdk_chain` release.
+    pub fn balance(&self, min_confirmations: u32) -> Balance {
         let graph = self.tx_graph.graph();
         let index = &self.tx_graph.index;
         let chain_tip = self.chain.tip().block_id();
+        let tip_height = chain_tip.height;
 
-        // A tx pulls in untrusted funds if any of its inputs spends an output we don't own
-        // (foreign spk, or unknown to our graph). Transitive taint through unconfirmed ancestry
-        // is handled by `classify_outpoints`, which calls this on every unsettled ancestor.
+        // A tx pulls in untrusted funds if any of its inputs spends an output we don't own (foreign spk, or unknown to our graph). 
+        // Transitive taint through unconfirmed ancestry is handled by `classify_outpoints`, which calls this on every unsettled ancestor.
         let does_taint = |ctx: &CanonicalTx<ChainPosition<ConfirmationBlockTime>>| {
             ctx.tx.input.iter().any(|txin| {
                 let op = txin.previous_output;
@@ -1178,6 +1180,12 @@ impl Wallet {
             })
         };
 
+        let min_confirmations = min_confirmations.max(1);
+        let is_settled = move |pos: &ChainPosition<ConfirmationBlockTime>| {
+            pos.confirmation_height_upper_bound()
+                .is_some_and(|h| tip_height - h + 1 >= min_confirmations)
+        };
+
         let view = self
             .chain
             .canonical_view(graph, chain_tip, CanonicalParams::default());
@@ -1186,7 +1194,7 @@ impl Wallet {
         for (txout, eligibility) in view.classify_outpoints(
             index.outpoints().iter().map(|(_, op)| *op),
             does_taint,
-            |pos| pos.is_confirmed(),
+            is_settled,
         ) {
             let bucket = match eligibility {
                 Eligibility::Settled => &mut balance.confirmed,
--- a/src/wallet/mod.rs
+++ b/src/wallet/mod.rs
@@ -32,7 +32,7 @@ use bdk_chain::{
     },
     tx_graph::{CalculateFeeError, CanonicalTx, TxGraph, TxUpdate},
     BlockId, CanonicalizationParams, ChainPosition, ConfirmationBlockTime, DescriptorExt,
-    FullTxOut, Indexed, IndexedTxGraph, Indexer, Merge,
+    Eligibility, FullTxOut, Indexed, IndexedTxGraph, Indexer, Merge, Trust,
 };
 use bitcoin::{
     absolute,
@@ -1153,14 +1153,52 @@ impl Wallet {
 
     /// Return the balance, separated into available, trusted-pending, untrusted-pending, and
     /// immature values.
+    ///
+    /// A pending output is trusted only when its entire unconfirmed ancestry spends coins we own.
+    /// If any unconfirmed ancestor pulls in a foreign or unknown output, the output is untrusted.
+    ///
+    // NOTE: depends on `CanonicalView` (bitcoindevkit/bdk#2246), not yet in a published
+    // `bdk_chain` release.
     pub fn balance(&self) -> Balance {
-        self.tx_graph.graph().balance(
-            &self.chain,
-            self.chain.tip().block_id(),
-            CanonicalizationParams::default(),
-            self.tx_graph.index.outpoints().iter().cloned(),
-            |&(k, _), _| k == KeychainKind::Internal,
-        )
+        let graph = self.tx_graph.graph();
+        let index = &self.tx_graph.index;
+        let chain_tip = self.chain.tip().block_id();
+
+        // A tx pulls in untrusted funds if any of its inputs spends an output we don't own
+        // (foreign spk, or unknown to our graph). Transitive taint through unconfirmed ancestry
+        // is handled by `classify_outpoints`, which calls this on every unsettled ancestor.
+        let does_taint = |ctx: &CanonicalTx<ChainPosition<ConfirmationBlockTime>>| {
+            ctx.tx.input.iter().any(|txin| {
+                let op = txin.previous_output;
+                !op.is_null()
+                    && graph
+                        .get_txout(op)
+                        .map(|txo| index.index_of_spk(txo.script_pubkey.clone()).is_none())
+                        .unwrap_or(true)
+            })
+        };
+
+        let view = self
+            .chain
+            .canonical_view(graph, chain_tip, CanonicalParams::default());
+
+        let mut balance = Balance::default();
+        for (txout, eligibility) in view.classify_outpoints(
+            index.outpoints().iter().map(|(_, op)| *op),
+            does_taint,
+            |pos| pos.is_confirmed(),
+        ) {
+            let bucket = match eligibility {
+                Eligibility::Settled => &mut balance.confirmed,
+                Eligibility::Immature => &mut balance.immature,
+                Eligibility::Unsettled(Trust::Trusted) => &mut balance.trusted_pending,
+                Eligibility::Unsettled(Trust::Untrusted | Trust::Unknown) => {
+                    &mut balance.untrusted_pending
+                }
+            };
+            *bucket += txout.txout.value;
+        }
+        balance
     }
 
     /// Add an external signer
--- a/tests/wallet.rs
+++ b/tests/wallet.rs
@@ -3007,3 +3007,260 @@ fn test_tx_ordering_untouched_preserves_insertion_ordering_bnb_success() {
         "UTXOs should be ordered with required first, then selected"
     );
 }
+
+#[test]
+fn test_trusted_pending_balance_from_owned_outpoints() {
+    let (mut wallet, txid) = get_funded_wallet_wpkh();
+    let tx = Transaction {
+        input: vec![TxIn {
+            previous_output: OutPoint { txid, vout: 0 },
+            ..Default::default()
+        }],
+        output: vec![TxOut {
+            value: Amount::from_sat(500),
+            script_pubkey: wallet
+                .next_unused_address(KeychainKind::Internal)
+                .address
+                .script_pubkey(),
+        }],
+        version: transaction::Version::ONE,
+        lock_time: absolute::LockTime::ZERO,
+    };
+
+    insert_tx(&mut wallet, tx.clone());
+
+    let balance = wallet.balance();
+
+    assert!(balance.trusted_pending > Amount::ZERO);
+    assert_eq!(balance.untrusted_pending, Amount::ZERO);
+}
+
+#[test]
+fn test_untrusted_pending_balance_from_external_inputs() {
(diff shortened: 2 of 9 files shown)


src/wallet/mod.rs
```diff
@@ -1115,12 +1115,48 @@ impl Wallet {
...
+                    changed = true;
+                }
+            }
+        }
+
+        let outpoints_with_txid = index
+            .outpoints()
+            .iter()
+            .cloned()
+            .map(|((k, i), op)| ((k, i, op.txid), op));
```
ValuedMammal: 💯 This looks like an easier solution (including the txid with the output index) than rewriting the entire balance function on TxGraph.

src/wallet/mod.rs
```diff
@@ -1115,12 +1115,48 @@ impl Wallet {
...
+
+        graph.balance(
             &self.chain,
             self.chain.tip().block_id(),
             CanonicalizationParams::default(),
-            self.tx_graph.index.outpoints().iter().cloned(),
-            |&(k, _), _| k == KeychainKind::Internal,
+            outpoints_with_txid,
+            |&(k, _, txid), _| k == KeychainKind::Internal || trusted_txids.contains(&txid),
```
ValuedMammal: I think we should drop the `KeychainKind::Internal` check here entirely as there's no guarantee that foreign wallets don't pay to our change address by discovering it on chain, see https://github.com/bitcoindevkit/bdk_wallet/issues/16#issue-2978011098.

ValuedMammal (commented):
Interesting approach. I'm not sure that I follow the transitive trust logic though - if Alice sends money to Bob, she might trust her own transaction but not Bob's transaction spending the output.

qatkk:
I went through the code and ran the tests, everything looks good and all the tests pass aea97d46eb0e04cbb43a530159a103b48767a27f
I just wanted to mention one small point. The current implementation might process a transaction multiple times (depending on transaction dependencies and the graph structure), since it keeps going through the graph until no new transactions are added to the trusted transaction list. 
For normal wallets with small transaction graphs, this shouldn’t be an issue, but it’s something to keep in mind for larger graphs or if performance ever becomes a concern. I also came across a discussion benchmarking transaction graph reconciliation and touching on transaction graph size [Issue #1687](https://github.com/bitcoindevkit/bdk/issues/1687?utm_source=chatgpt.com). 
Even so, I think it’s probably not worth adding extra complexity (like Kahn’s algorithm) just for efficiency.

src/wallet/mod.rs
qatkk: I agree with dropping the KeychainKind::Internal check. Another example (if I’m not mistaken) could be that, in a multisig wallet, outputs that belong to a shared descriptor can be marked trusted even if the wallet didn’t actually fund them. Dropping the check makes sure this doesn't happen.

Dmenec:
> Interesting approach. I'm not sure that I follow the transitive trust logic though - if Alice sends money to Bob, she might trust her own transaction but not Bob's transaction spending the output.

Yup, I think just verifying that all inputs of a transaction are owned by the wallet and the output belongs to it should be a better approach.

_[edit]_

Discussing with @qatkk we found there was still a problem with dependencies (addressed in 3b2af6bef81cd7943ab3410ce11fdbc61d3e3b63)

It now propagates untrust by iterating over all transactions and marks owned outputs as untrusted if any of their inputs are external or inherit untrust from a previous transaction, then excludes them from the trusted set. This should solve the untrusted problem if there's some untrusted transacition in the middle.

src/wallet/mod.rs
```diff
@@ -1115,12 +1115,54 @@ impl Wallet {
...
+        let mut changed = true;
+        while changed {
+            changed = false;
+            for tx_node in graph.full_txs() {
+                if tx_node.tx.is_coinbase() {
+                    continue;
+                }
+                let is_untrusted = tx_node.tx.input.iter().any(|txin| {
+                    !owned_outpoints.contains(&txin.previous_output)
+                        || untrusted_outpoints.contains(&txin.previous_output)
```
nymius: I don't know if I'm missing an assumption over graph here, but if `txin.previous_output` is confirmed, wouldn't it mean that that output should be considered trusted?

This case:
```rust
#[test]
fn test_pay_to_internal_from_not_trusted_but_confirmed() {
    let (mut wallet, _) = get_funded_wallet_wpkh();

    // Build a tx whose input comes from an unknown (external) outpoint,
    // but whose output goes to our change (internal) keychain address.
    let external_txid = Txid::from_raw_hash(Hash::all_zeros());
    let tx = Transaction {
        input: vec![TxIn {
            previous_output: OutPoint {
                txid: external_txid,
                vout: 0,
            },
            ..Default::default()
        }],
        output: vec![TxOut {
            value: Amount::from_sat(500),
            script_pubkey: wallet
                .next_unused_address(KeychainKind::Internal)
                .address
                .script_pubkey(),
        }],
        version: transaction::Version::ONE,
    };

    insert_tx(&mut wallet, tx);

    let anchor = ConfirmationBlockTime {
        block_id: wallet.latest_checkpoint().get(2000).unwrap().block_id(),
        confirmation_time: 0,
    };
    
    // The transaction is confirmed
    insert_anchor(&mut wallet, external_txid, anchor);

    let balance = wallet.balance();

    // The output is ours but the input is not owned, so it must be untrusted_pending.
    assert!(balance.trusted_pending > Amount::ZERO);
    assert_eq!(balance.untrusted_pending, Amount::ZERO);
}
```

src/wallet/mod.rs
```diff
@@ -1115,12 +1115,54 @@ impl Wallet {
...
+        let graph = self.tx_graph.graph();
+        let index = &self.tx_graph.index;
+
+        let owned_outpoints: HashSet<OutPoint> =
+            index.outpoints().iter().map(|(_, op)| *op).collect();
+
+        let mut untrusted_outpoints: HashSet<OutPoint> = HashSet::new();
+
+        let mut changed = true;
+        while changed {
```
nymius: I would change this for `loop { ... }` and break if conditions are met. `changed` doesn't hold any meaning by itself outside of this particular context.

nymius (commented):
Although the code fixes the referenced issue, Isn't the code too much constrained by the `TxGraph::balance` signature?

The `trust_predicate` is missing key information that's available in the context of execution. What about bypassing the graph balance method and re implementing the logic directly in wallet?

nymius:
> For normal wallets with small transaction graphs, this shouldn’t be an issue, but it’s something to keep in mind for larger graphs or if performance ever becomes a concern. I also came across a discussion benchmarking transaction graph reconciliation and touching on transaction graph size [Issue #1687](https://github.com/bitcoindevkit/bdk/issues/1687).

We are still behind on this, are you willing to add bench marking test for this PR?

Dmenec:
> Although the code fixes the referenced issue, Isn't the code too much constrained by the `TxGraph::balance` signature?

Yes! 

> The `trust_predicate` is missing key information that's available in the context of execution. What about bypassing the graph balance method and re implementing the logic directly in wallet?

Probably it's the best option, maybe tempting to deprecate it in the future on `bdk_chain` (if I'm not missing something on it's usages). Another option could be modify the signature to receive more context, but that would also be a breaking change, so bypassing it might be the easier path for incremental integration.

src/wallet/mod.rs
Dmenec: Solved in 3b2af6bef81cd7943ab3410ce11fdbc61d3e3b63

src/wallet/mod.rs
Dmenec: Yup, I think there is another bug in there. Maybe another hashset of confirmed outpoints and checking if the hashet contains it on: 

```
let is_untrusted = tx_node.tx.input.iter().any(|txin| {
    !owned_outpoints.contains(&txin.previous_output)
        || untrusted_outpoints.contains(&txin.previous_output)
});
```
Would solve that?

qatkk:
> We are still behind on this, are you willing to add bench marking test for this PR?

Sure just to make it a bit more clear, the benchmarking would be only to report the overhead of the balance computation regarding the graph size and its topology, true?

nymius:
> Sure just to make it a bit more clear, the benchmarking would be only to report the overhead of the balance computation regarding the graph size and its topology, true?

Yes, just benchmark the overhead of the balance computation in relation to the graph size. Adding topology without at least a simpler benchmark is an overkill for now.

src/wallet/mod.rs
nymius: Sorry, did you intend to show something in the code? I'm not getting it.
Anyways, I wouldn't be using hash sets lightly here, not without testing against large cases.

Dmenec: Yes, I think I lost context just with this snippet. The point was to add another condition there to check if it's confirmed or not, maybe by storing all the confirmed transactions in a separate hash set. 

Agreeing with benchmarking it first.

nymius: After discussing with @ValuedMammal, we conclude the definition of trust is: `all inputs must be controlled by this wallet`.
I don't know if it's already in the docs, but it would be worth adding it here.
In that case, my original comment is invalid, and you shouldn't make changes based on it.

qatkk:
> Yes, just benchmark the overhead of the balance computation in relation to the graph size. Adding topology without at least a simpler benchmark is an overkill for now.

I'm on it.

Meanwhile I wanted to propose a new approach to address the balance computation. Since the concept of `Balance` for a wallet only concerns the unspent outputs, I think it would be better to only process the transactions or the ancestors of a transaction from which the wallet owns unspent outputs. This approach effectively removes the need for traversing the transaction graph (which only grows in size with more usage of the wallet) multiple times. Can you please let me know if this approach makes sense to you?

src/wallet/mod.rs
qatkk: Regarding the definition of trust, are we deliberately not taking into account if a transaction is confirmed or not?

nymius:
> Meanwhile I wanted to propose a new approach to address the balance computation. Since the concept of `Balance` for a wallet only concerns the unspent outputs, I think it would be better to only process the transactions or the ancestors of a transaction from which the wallet owns unspent outputs. This approach effectively removes the need for traversing the transaction graph (which only grows in size with more usage of the wallet) multiple times.

Makes sense. Could you show me where the traversing is done multiple times (in the code)? And how you would do it differently.

qatkk:
> Makes sense. Could you show me where the traversing is done multiple times (in the code)? And how you would do it differently.
 

Sure. Basically each time the following "if" (from 3674263bc0bd15b7473da4a2d43af569df1ccf52) is executed, the code repeats processing all the nodes in the transaction graph.

```rust
 if owned_outpoints.contains(&op) && untrusted_outpoints.insert(op) {
changed = true;
}
```

We can process the chain of the transactions corresponding to each UTXO from the UTXO set of the wallet. The processing starts from each UTXO and followed by the parents of its transaction till we encounter a transaction which is with a high probability trusted/untrusted.
In the following image (from [Bitcoin Developer](https://developer.bitcoin.org/devguide/block_chain.html)), imagine the 20k unspent output belonging to our wallet. We can then start processing tx3 then tx1, and tx0. The process will finish once we encounter a transaction that has external inputs and is unconfirmed, making the UTXO untrusted, or a confirmed/trusted transaction, resulting in the UTXO being trusted.
<p align="center">
  
</p>

ValuedMammal:
> In the following image (from [Bitcoin Developer](https://developer.bitcoin.org/devguide/block_chain.html)), imagine the 20k unspent output belonging to our wallet. We can then start processing tx3 then tx1, and tx0. The process will finish once we encounter a transaction that has external inputs and is unconfirmed, making the UTXO untrusted, or a confirmed/trusted transaction, resulting in the UTXO being trusted.

That's a good point @qatkk, the untrusted quality should propagate down through the descendants.

src/wallet/mod.rs
nymius: Yes, because the notion of trust is only relevant for outputs from unconfirmed transactions.
**Update**: I didn't consider the min confirmation parameter here. So yes, you've to check confirmed transactions if you are using that parameter.

nymius:
Thanks for the explanation @qatkk 

> We can process the chain of the transactions corresponding to each UTXO from the UTXO set of the wallet. The processing starts from each UTXO and followed by the parents of its transaction till we encounter a transaction which is with a high probability trusted/untrusted.

It's a good idea. Is down propagation, but we walk from the leaves of the graph, so we still have to keep a backlog of "unmarked" (as trusted or unstrusted) outputs.
The criteria for trusted is still the one stated above: all inputs are controlled by the wallet. If there are external inputs, or one of the inputs was marked as untrusted, all outputs should be untrusted.

I've to think more on the `min_confirmations` handling.

cc: @Dmenec

Dmenec:
Thinking a bit about it, @qatkk I think you're right with the suggestion. 

The current algorithm has a complexity of **O(O x T x (I + V))**, being **O** the number of indexed outpoints, **T** the total number of tx in the graph, **I** the inputs and **V** the outputs. The new approach would be **O(U x D x I)** where **U** is the number of unconfirmed UTXOs and **D** the depth of the ancestor chain. 

While it is possible that some outputs have a lot of ancestors, the main reason I see to change the algorithm is that this new approach does not grow with wallet history. 

That said, I'm considering using `list_unspent()` filtering by `ChainPosition::Unconfirmed`. Note that UTXOs confirmed but below the `min_confirmations` threshold would not be captured by this filter.

So, regarding `min_confirmations` @nymius mentions, and if my assumptions are not wrong. Trusted and untrusted exist only to avoid a potential double-spend risk when the tx is unconfirmed. Once the transaction is confirmed, that risk no longer exists, aside from the possibility of a reorg, which in my opinion is a fundamentally different concept. Threfore, applying those trust tags would be semantically incorrect. 

It might be worth considering an intermediate state that expresses this "confirmed but not yet considered final" and handle them aside trust logic? Would this make sense?

nymius:
> is that this new approach does not grow with wallet history.

Just to make it clear, It may grow, but it is bounded by the amount of unconfirmed UTXOs and their ancestors. 

> That said, I'm considering using `list_unspent()` filtering by `ChainPosition::Unconfirmed`

I'm thinking the same. I won't care if you just copy the chain balance body and use it here. It would be a breaking change, but at least we are not breaking bdk-chain upstream, and we can roll the change there in the future.

> Trusted and untrusted exist only to avoid a potential double-spend risk when the tx is unconfirmed. Once the transaction is confirmed, that risk no longer exists, aside from the possibility of a reorg, which in my opinion is a fundamentally different concept. Therefore, applying those trust tags would be semantically incorrect.

How different it is (the reorg)? As long as reorgs change your trust assumptions, I think they are related; and must be considered. The min confirmations is like a "degree of distrust" parameter. So applying the trust tags still applies.

> It might be worth considering an intermediate state that expresses this "confirmed but not yet considered final" and handle them aside trust logic? Would this make sense?

It makes sense. The actual balance in chain works as a reduce operation, erasing details in the process. But this may also be fixed by adding the number of confirmations to the `Confirmed` outpoints, then applying the trust distinction further in the pipeline. Which is what the code in chain is currently doing, but defining trust based on spks. Which leads us back to trying a solution by copying that code and fixing it here.

Dmenec:
Discussing yesterday with @nymius we went through a few things worth noting here.

First, the current DFS could be replaced by `TxAncestors`, which internally uses BFS. The advantage is that with BFS you process ancestors level by level, so if an untrusted node exists at a shallow depth you find it before going deeper into other branches. A closure could be used to label each ancestor as trusted or untrusted, as done [here](https://github.com/bitcoindevkit/bdk/pull/2038/changes#diff-e8ad39c06ed87b600836f1f6414ceedebb47ae82c4c136bf99a82bbbdfba692fR273).

We also discussed how `min_confirmations` should interact with trust classification. Right now the traversal stops at any confirmed ancestor. With `min_confirmations` you'd want to stop only at ancestors with at least N confirmations, which means canonicalization itself would need to be run from `tip - min_confirmations` rather than the current tip.

Related with the trust concept... a UTXO is trusted pending if you control all inputs up the ancestor chain. External inputs mean a third party could RBF or double-spend before confirmation, so we mark it untrusted. As @qatkk noted, the traversal terminates either when it finds an unconfirmed ancestor spending an output that doesn't belong to our wallet, making the UTXO untrusted, or when it reaches a confirmed ancestor, making it trusted. When I have time I'll write an explicit breakdown of each trusted/untrusted case, including how `min_confirmations` fits in, and add it to the `Balance` docs.

src/wallet/mod.rs
Dmenec: Resolving as does not hold anymore

tests/wallet.rs
```diff
@@ -3007,3 +3007,260 @@ fn test_tx_ordering_untouched_preserves_insertion_ordering_bnb_success() {
...
+                .script_pubkey(),
+        }],
+        version: transaction::Version::ONE,
+        lock_time: absolute::LockTime::ZERO,
+    };
+    insert_tx(&mut wallet, tx_b);
+
+    let balance = wallet.balance();
+
+    assert!(balance.trusted_pending > Amount::ZERO);
```
nymius: Here we want to verify both balances are considered, so I would be more precise with the threshold to surpass or directly use equality with the expected number.

tests/wallet.rs
```diff
@@ -3007,3 +3007,260 @@ fn test_tx_ordering_untouched_preserves_insertion_ordering_bnb_success() {
...
```
nymius: I would check exact amounts here, this could be accounting for the second output of `tx_a` **and** the only output of `tx_b`, which is wrong.

nymius (commented):
I'm exploring other alternative approaches. Will share once I'm happy with them.

Dmenec:
Made some changes to align with bitcoindevkit/bdk#2246. Keeping this as a draft to move it forward once everything is in place.

tests/wallet.rs
Dmenec: Solved in ff1dc99a017265fd94a489d48b0ffdef8a585d67

tests/wallet.rs
Dmenec: Solved in ff1dc99a017265fd94a489d48b0ffdef8a585d67

src/wallet/mod.rs
```diff
@@ -1154,14 +1153,51 @@ impl Wallet {
 
     /// Return the balance, separated into available, trusted-pending, untrusted-pending, and
     /// immature values.
+    ///
+    /// A pending output is trusted only when its entire unconfirmed ancestry spends coins we own.
+    /// If any unconfirmed ancestor pulls in a foreign or unknown output, the output is untrusted.
+    ///
+    /// NOTE: depends on `CanonicalView` (bitcoindevkit/bdk#2246), not yet in a published
```
nymius: Make this a comment (i.e., `// NOTE:`) so IDEs highlight it.

src/wallet/mod.rs
```diff
@@ -1154,14 +1153,51 @@ impl Wallet {
...
+
+        let view = self
+            .chain
+            .canonical_view(graph, chain_tip, CanonicalParams::default());
+
+        let mut balance = Balance::default();
+        for (txout, eligibility) in view.classify_outpoints(
+            index.outpoints().iter().map(|(_, op)| *op),
+            does_taint,
+            // TODO: take `min_confirmations` and build `is_settled` from it.
```
nymius: Why `min_confirmations` isn't here yet?

src/wallet/mod.rs
Dmenec: Solved in 3c5ad5fb7c523879998242769c7349f942e046e4

src/wallet/mod.rs
Dmenec: Solved in 56c1299a75789e976a8ef6767815ba1d82b911ad

PR #433, opened 2026-04-07
feat: expose detailed try_finalize_psbt outcomes

### LINKED ISSUE #73: additional information for `sign` when returning false
**Describe the enhancement**  
I'm wondering if anyone has potentially thought of any alternatives to the return type than what currently exists for [sign](https://docs.rs/bdk_wallet/latest/bdk_wallet/struct.Wallet.html#method.sign) `Result<bool, SignerError>`

**Use case**  
I'm specifically thinking about a client app signing a transaction and getting `bool` of `false` back without additional information of why `sign` returned `false`. 

Maybe they were expecting it to return `false` but maybe not; that client app may expect successful `sign`ing, not receive `true` or `SignerError` back, but get back `false` instead without extra context, on what could be considered an imporant operation for their app.

**Additional context**
Just looking for feedback/thoughts if there has been conversation about this historically/presently, and any possible thoughts.

reez:
### Description
Close #73

This PR keeps the scope to PSBT finalization only:

- add `try_finalize_psbt` which returns `FinalizePsbtOutcome`
- report per input outcomes via `FinalizeInputOutcome`

### Notes to the reviewers

This is now additive rather than breaking: `Wallet::finalize_psbt` still returns `bool`

Intentionally not included here:
- changing `Wallet::sign` to return a richer result type
- any broader redesign beyond the PSBT finalization API surface

### Changelog notice

- Add `Wallet::try_finalize_psbt`, which returns `FinalizePsbtOutcome` and exposes per-input finalization outcomes while preserving the existing `Wallet::finalize_psbt` API.

#### All Submissions:

#### New Features:

### DIFF
--- a/src/types.rs
+++ b/src/types.rs
@@ -14,6 +14,8 @@ use chain::{ChainPosition, ConfirmationBlockTime};
 use core::convert::AsRef;
 use core::fmt;
 
+use crate::collections::BTreeMap;
+
 use bitcoin::transaction::{OutPoint, Sequence, TxOut};
 use bitcoin::{psbt, Weight};
 
@@ -138,6 +140,55 @@ impl Utxo {
             Utxo::Local(_) => None,
             Utxo::Foreign { sequence, .. } => Some(*sequence),
         }
+    }
+}
+
+/// The finalization status for a single PSBT input.
+#[derive(Debug, PartialEq)]
+pub enum FinalizeInputOutcome {
+    /// The input was already finalized before this call.
+    AlreadyFinalized,
+    /// The input was successfully finalized during this call.
+    Finalized,
+    /// The wallet could not derive a descriptor for the input.
+    MissingDescriptor,
+    /// The wallet found the descriptor but could not construct the input satisfaction.
+    CouldNotSatisfy(miniscript::Error),
+}
+
+impl FinalizeInputOutcome {
+    /// Whether the input is finalized after this call.
+    pub fn is_finalized(&self) -> bool {
+        matches!(self, Self::AlreadyFinalized | Self::Finalized)
+    }
+}
+
+/// The outcome of a PSBT finalization attempt.
+#[derive(Debug, PartialEq)]
+pub struct FinalizePsbtOutcome {
+    outcomes: BTreeMap<usize, FinalizeInputOutcome>,
+}
+
+impl FinalizePsbtOutcome {
+    pub(crate) fn new(outcomes: BTreeMap<usize, FinalizeInputOutcome>) -> Self {
+        Self { outcomes }
+    }
+
+    /// Whether all inputs are finalized after this call.
+    pub fn is_finalized(&self) -> bool {
+        self.outcomes
+            .values()
+            .all(FinalizeInputOutcome::is_finalized)
+    }
+
+    /// Borrow the per-input finalization outcomes.
+    pub fn outcomes(&self) -> &BTreeMap<usize, FinalizeInputOutcome> {
+        &self.outcomes
+    }
+
+    /// Consume the collection and return the per-input finalization outcomes.
+    pub fn into_outcomes(self) -> BTreeMap<usize, FinalizeInputOutcome> {
+        self.outcomes
     }
 }
 
--- a/src/wallet/mod.rs
+++ b/src/wallet/mod.rs
@@ -184,6 +184,55 @@ impl fmt::Display for AddressInfo {
 /// A `CanonicalTx` managed by a `Wallet`.
 pub type WalletTx<'a> = CanonicalTx<'a, Arc<Transaction>, ConfirmationBlockTime>;
 
+/// The finalization status for a single PSBT input.
+#[derive(Debug, PartialEq)]
+pub enum FinalizeInputOutcome {
+    /// The input was already finalized before this call.
+    AlreadyFinalized,
+    /// The input was successfully finalized during this call.
+    Finalized,
+    /// The wallet could not derive a descriptor for the input.
+    MissingDescriptor,
+    /// The wallet found the descriptor but could not construct the input satisfaction.
+    CouldNotSatisfy(miniscript::Error),
+}
+
+impl FinalizeInputOutcome {
+    /// Whether the input is finalized after this call.
+    pub fn is_finalized(&self) -> bool {
+        matches!(self, Self::AlreadyFinalized | Self::Finalized)
+    }
+}
+
+/// Holds per-input PSBT finalization outcomes.
+#[derive(Debug, PartialEq)]
+pub struct FinalizedInputs {
+    outcomes: BTreeMap<usize, FinalizeInputOutcome>,
+}
+
+impl FinalizedInputs {
+    fn new(outcomes: BTreeMap<usize, FinalizeInputOutcome>) -> Self {
+        Self { outcomes }
+    }
+
+    /// Whether all inputs are finalized after this call.
+    pub fn is_finalized(&self) -> bool {
+        self.outcomes
+            .values()
+            .all(FinalizeInputOutcome::is_finalized)
+    }
+
+    /// Borrow the per-input finalization outcomes.
+    pub fn outcomes(&self) -> &BTreeMap<usize, FinalizeInputOutcome> {
+        &self.outcomes
+    }
+
+    /// Consume the collection and return the per-input finalization outcomes.
+    pub fn into_outcomes(self) -> BTreeMap<usize, FinalizeInputOutcome> {
+        self.outcomes
+    }
+}
+
 impl Wallet {
     /// Build a new single descriptor [`Wallet`].
     ///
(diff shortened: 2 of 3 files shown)


reez:
Updated per some good feedback from mammal

ValuedMammal (changes requested):
I don't think After and Older are doing much good, and would prefer to leave them out of the new implementation if we plan to support `try_finalize_psbt` into the future.

notmandatory:
Concept ACK, glad to see this as a non-breaking change.

reez:
Added four commits as small follow-ups from review feedback from mammal: make finalization cleanup more explicit, preserve opaque PSBT/input metadata, stop the new `try_finalize_psbt` path from redacting output metadata, and clarify the internal timelock inputs without changing legacy `finalize_psbt` behavior.

src/wallet/mod.rs
```diff
@@ -184,6 +184,55 @@ impl fmt::Display for AddressInfo {
...
+    /// Borrow the per-input finalization outcomes.
+    pub fn outcomes(&self) -> &BTreeMap<usize, FinalizeInputOutcome> {
+        &self.outcomes
+    }
+
+    /// Consume the collection and return the per-input finalization outcomes.
+    pub fn into_outcomes(self) -> BTreeMap<usize, FinalizeInputOutcome> {
+        self.outcomes
+    }
+}
```
oleonardolima: nit: these should are better suited either in the psbt.rs or types.rs

src/wallet/mod.rs
```diff
@@ -184,6 +184,55 @@ impl fmt::Display for AddressInfo {
 /// A `CanonicalTx` managed by a `Wallet`.
 pub type WalletTx<'a> = CanonicalTx<'a, Arc<Transaction>, ConfirmationBlockTime>;
 
+/// The finalization status for a single PSBT input.
+#[derive(Debug, PartialEq)]
+pub enum FinalizeInputOutcome {
+    /// The input was already finalized before this call.
+    AlreadyFinalized,
+    /// The input was successfully finalized during this call.
+    Finalized,
```
oleonardolima: question: are we using this differently ? otherwise, a single `Finalized` would be enough imho.

src/wallet/mod.rs
```diff
@@ -184,6 +184,55 @@ impl fmt::Display for AddressInfo {
...
+    pub fn is_finalized(&self) -> bool {
+        matches!(self, Self::AlreadyFinalized | Self::Finalized)
+    }
+}
+
+/// Holds per-input PSBT finalization outcomes.
+#[derive(Debug, PartialEq)]
+pub struct FinalizedInputs {
+    outcomes: BTreeMap<usize, FinalizeInputOutcome>,
+}
```
oleonardolima: nit: i think the name is a bit odd, not sure if `FinalizePsbtOutcome` would be better wdyt ?

src/wallet/mod.rs
```diff
@@ -184,6 +184,55 @@ impl fmt::Display for AddressInfo {
...
+
+    /// Borrow the per-input finalization outcomes.
+    pub fn outcomes(&self) -> &BTreeMap<usize, FinalizeInputOutcome> {
+        &self.outcomes
+    }
+
+    /// Consume the collection and return the per-input finalization outcomes.
+    pub fn into_outcomes(self) -> BTreeMap<usize, FinalizeInputOutcome> {
+        self.outcomes
+    }
```
oleonardolima: question: is this being used ?

src/wallet/mod.rs
```diff
@@ -1917,23 +1966,65 @@ impl Wallet {
...
+        psbt: &mut Psbt,
+    ) -> Result<FinalizedInputs, IndexOutOfBoundsError> {
+        self.try_finalize_psbt_with(psbt, None, |_, _| None, false)
+    }
+
+    fn try_finalize_psbt_with<F>(
+        &self,
+        psbt: &mut Psbt,
+        current_height: Option<u32>,
+        mut confirmation_height_for_input: F,
```
oleonardolima, line 1960: question: if i got it right, you added this to not depend on `SignOptions` ?

reez: Yeah that's part of it for sure. `try_finalize_psbt` is meant to just finalize the PSBT, so I didn’t want callers to have to pass `SignOptions` only for `assume_height`.

src/wallet/mod.rs
reez: Not internally no. I added it as a convenience for external callers that want to consume `FinalizedInputs` and take ownership of the outcomes map, but `outcomes()` covers the current usage. 

(Since `FinalizedInputs` owns the outcomes map but keeps the field private, `outcomes()` only supports borrowed inspection, `into_outcomes()` lets callers consume the result and take ownership of the map without cloning.)

Totally happy to remove it tho, its a good question.

src/wallet/mod.rs
reez: Yeah I think you're totally right. `FinalizedInputs` can read like it only contains finalized inputs, but it actually represents the whole PSBT finalization attempt, including failures. `FinalizePsbtOutcome` pairs better with `FinalizeInputOutcome`, so I’ll rename it that way right now. Great call!

reez: ok updated code to reflect this change now, thanks a bunch @oleonardolima

src/wallet/mod.rs
reez: I see where you're coming from for sure but I do think the distinctions intentional. `Finalized` would be this call was able to finalize the input, `AlreadyFinalized` would be the input already had final script/witness data before this call and we skipped it. `is_finalized()` treats both as success, but I thought preserving that difference was useful since this API is meant to explain what happened per input.

src/wallet/mod.rs
reez: Agree. I'm thinking `types.rs` is probably a better home so I’ll move them there.

src/wallet/mod.rs
oleonardolima: > Yeah that's part of it for sure. `try_finalize_psbt` is meant to just finalize the PSBT, so I didn’t want callers to have to pass `SignOptions` only for `assume_height`.

Great, sounds good!

src/wallet/mod.rs
oleonardolima: Good, I think I see where you're going. Agree, it's probably useful to have that distinction if it'll be used for upstream users (e.g UX/UI).

src/wallet/mod.rs
```diff
@@ -1917,23 +1917,65 @@ impl Wallet {
...
+    ) -> Result<FinalizePsbtOutcome, IndexOutOfBoundsError> {
+        self.try_finalize_psbt_with(psbt, None, |_, _| None, false)
+    }
+
+    fn try_finalize_psbt_with<F>(
+        &self,
+        psbt: &mut Psbt,
+        current_height: Option<u32>,
+        mut confirmation_height_for_input: F,
+        clear_output_derivations: bool,
```
oleonardolima, line 1961: question: what scenarios you have in mind for this new clearing output behavior ?

oleonardolima (commented):
cACK 09d2cd04a1dbfc028c2d00900f87baa5d11f65c9

src/wallet/mod.rs
reez: Inspiration/suggestion from mammal pointing me at https://github.com/bitcoindevkit/bdk-tx/issues/77#issue-4562924722 + `Make finalize pure BIP-174 (inputs only; outputs untouched). Redaction becomes explicit opt-in.`

Existing `finalize_psbt` passes `true` to preserve its current output derivation clearing behavior, new `try_finalize_psbt` passes `false` so it doesn't clear output metadata.

oleonardolima: Sounds good! Yes, if I'm not mistaken that's the right approach as per BIP.

Musab1258:
tACK 09d2cd04a1dbfc028c2d00900f87baa5d11f65c9

The concept and approach make sense. I fetched the branch and ran the tests locally, and they all pass. The documentation also clearly specifies when to use the finalize_psbt and try_finalize_psbt methods.

src/types.rs
```diff
@@ -141,6 +143,55 @@ impl Utxo {
...
+    outcomes: BTreeMap<usize, FinalizeInputOutcome>,
+}
+
+impl FinalizePsbtOutcome {
+    pub(crate) fn new(outcomes: BTreeMap<usize, FinalizeInputOutcome>) -> Self {
+        Self { outcomes }
+    }
+
+    /// Whether all inputs are finalized after this call.
+    pub fn is_finalized(&self) -> bool {
```
Musab1258, line 178: If outcomes is empty (e.g. a PSBT with no inputs), is_finalized() returns true, since .all() on an empty iterator is true. Is this the intended behavior or worth documenting/guarding against explicitly?

src/wallet/mod.rs
```diff
@@ -1917,23 +1917,65 @@ impl Wallet {
...
+    ) -> Result<FinalizePsbtOutcome, IndexOutOfBoundsError>
+    where
+        F: FnMut(usize, &bitcoin::TxIn) -> Option<u32>,
+    {
+        let tx = &psbt.unsigned_tx;
 
-        let mut finished = true;
+        let mut outcomes = BTreeMap::new();
 
         for (n, input) in tx.input.iter().enumerate() {
```
Musab1258, line 1976: If psbt.inputs.len() < tx.input.len() but not 0, this loop will partially mutate earlier inputs before hitting the IndexOutOfBoundsError on a later one. Is it expected to have a partially-mutated PSBT on Err , or should length be validated up front before any mutation starts?

src/types.rs
```diff
@@ -141,6 +143,55 @@ impl Utxo {
     }
 }
 
+/// The finalization status for a single PSBT input.
+#[derive(Debug, PartialEq)]
+pub enum FinalizeInputOutcome {
```
Musab1258, line 148: Would it make sense to add a Display impl for FinalizeInputOutcome so callers don't have to match on the variants themselves to produce an outcome that can be logged in user interfaces?

src/types.rs
reez: I think this is intended since `is_finalized()` is only answering whether any input failed finalization, so if there are no inputs there is nothing to fail. This also matches the old `finalize_psbt` behavior where `finished` started as `true` and only changed inside the input loop. But that's my take, open to further thoughts

src/wallet/mod.rs
reez: Agree I think validating up front is cleaner here. 6afd006737e3cc689db177617f9626b0055436d1 adds a guard before the loop so a PSBT with missing input metadata returns `IndexOutOfBoundsError` before any inputs are mutated, plus a regression test for the partial mutation case.

src/types.rs
reez: Maybe, but I lean toward leaving this out for now. I was thinking of `FinalizeInputOutcome` as structured data that callers can match on and render however they want, especially for UI wording, while `Debug` is still available for basic logging. Happy to add `Display` if others think that belongs in the initial API though.

ValuedMammal:
ACK ec3a683f0d9247c6e533914e84d83b68ad09d3ce

ValuedMammal:
`cargo +nightly fmt` produced a diff, and it wasn't caught by CI

```sh
                             let final_script_witness =
                                 (!tmp_input.witness.is_empty()).then_some(tmp_input.witness);

-                            // BIP174 finalization clears input metadata except UTXOs, final scripts,
-                            // and opaque fields the finalizer does not understand.
+                            // BIP174 finalization clears input metadata except UTXOs, final
+                            // scripts, and opaque fields the finalizer
+                            // does not understand.
                             *psbt_input = bitcoin::psbt::Input {
                                 non_witness_utxo: original.non_witness_utxo,
                                 witness_utxo: original.witness_utxo,
```

ValuedMammal (approved):
ACK 72261ced00fd0ce71a3377395de6d72b9d241e43

PR #436, opened 2026-04-08
test(locked_outpoints): add unit tests for ChangeSet merge and is_empty

KrishnaShuk:
### Description

This PR adds unit test coverage for `locked_outpoints::ChangeSet`, which previously had zero inline tests. It tests the `Merge` trait implementations, ensuring that [merge](https://github.com/bitcoindevkit/bdk_wallet/blob/5df32ca707e859506e8198f1f102aed87a837bf0/src/wallet/locked_outpoints.rs#L17) and [is_empty](https://github.com/bitcoindevkit/bdk_wallet/blob/5df32ca707e859506e8198f1f102aed87a837bf0/src/wallet/locked_outpoints.rs#L23) behave correctly under various conditions.

### Notes to the reviewers

- Added 6 deterministic tests verifying:
  - Default initialization is empty
  - Merging with empty datasets are clean no-ops
  - Disjoint changeset merges successfully combine outpoints
  - Overwriting logic accurately fulfills the `extend()` semantics where the incoming `other` duplicate outpoints overwrite `self`
- Uses idiomatic boolean assertions (`assert!(...)`) verified cleanly by `cargo clippy`.
- Tests use `Txid::from_byte_array` internally to maintain deterministic speed without relying on `rand`.

#### All Submissions:

#### New Features:

#### Bugfixes:

### DIFF
--- a/src/wallet/changeset.rs
+++ b/src/wallet/changeset.rs
@@ -417,3 +417,116 @@ impl From<locked_outpoints::ChangeSet> for ChangeSet {
         }
     }
 }
+
+#[cfg(test)]
+mod test {
+    use super::*;
+    use alloc::string::ToString;
+    use bdk_chain::Merge;
+    use bitcoin::Network;
+
+    /// Helper to get public descriptors via a temporary wallet.
+    fn test_descriptors() -> (
+        Descriptor<DescriptorPublicKey>,
+        Descriptor<DescriptorPublicKey>,
+    ) {
+        let (desc, change_desc) = crate::test_utils::get_test_wpkh_and_change_desc();
+        let wallet = crate::Wallet::create(desc.to_string(), change_desc.to_string())
+            .network(Network::Regtest)
+            .create_wallet_no_persist()
+            .unwrap();
+        (
+            wallet
+                .public_descriptor(crate::KeychainKind::External)
+                .clone(),
+            wallet
+                .public_descriptor(crate::KeychainKind::Internal)
+                .clone(),
+        )
+    }
+
+    #[test]
+    fn test_default_is_empty() {
+        let cs = ChangeSet::default();
+        assert!(cs.is_empty());
+    }
+
+    #[test]
+    fn test_is_empty_with_descriptor() {
+        let (desc, _) = test_descriptors();
+        let cs = ChangeSet {
+            descriptor: Some(desc),
+            ..Default::default()
+        };
+        assert!(!cs.is_empty());
+    }
+
+    #[test]
+    fn test_is_empty_with_network() {
+        let cs = ChangeSet {
+            network: Some(Network::Regtest),
+            ..Default::default()
+        };
+        assert!(!cs.is_empty());
+    }
+
+    #[test]
+    fn test_merge_sets_descriptor() {
+        let (desc, _) = test_descriptors();
+        let mut cs = ChangeSet::default();
+        let other = ChangeSet {
+            descriptor: Some(desc.clone()),
+            ..Default::default()
+        };
+        cs.merge(other);
+        assert_eq!(cs.descriptor, Some(desc));
+    }
+
+    #[test]
+    fn test_merge_sets_network() {
+        let mut cs = ChangeSet::default();
+        let other = ChangeSet {
+            network: Some(Network::Regtest),
+            ..Default::default()
+        };
+        cs.merge(other);
+        assert_eq!(cs.network, Some(Network::Regtest));
+    }
+
+    #[test]
+    fn test_merge_same_descriptor_is_ok() {
+        let (desc, _) = test_descriptors();
+        let mut cs = ChangeSet {
+            descriptor: Some(desc.clone()),
+            ..Default::default()
+        };
+        let other = ChangeSet {
+            descriptor: Some(desc.clone()),
+            ..Default::default()
+        };
+        cs.merge(other);
+        assert_eq!(cs.descriptor, Some(desc));
+    }
+
+    #[test]
+    fn test_merge_empty_into_non_empty() {
+        let (desc, change_desc) = test_descriptors();
+        let mut cs = ChangeSet {
+            descriptor: Some(desc),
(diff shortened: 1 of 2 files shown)


lambaaryan011 (commented):
## Issues Found 

1. **Missing Test for Multiple Merges**
   - The code only tests 2 merges (true → false)
   - Should test 3+ merges (true → false → true → false)
   - Example: What if merge happens 10 times?

2. **Missing Implementation Code**
   - Tests check merge() but we can't see merge() code
   - Need to see the actual implementation
   - Can't verify tests are correct without it

3. **Unclear Checklist**
   - "breaks the existing API" is confusing
   - Does it break or not? Need clarity

## Good Things 
- Tests are written correctly
- Uses `assert!()` properly
- Deterministic (no random data)

luisschwab:
@lambaaryan011 you only waste everyone's time with this LLM review.

lambaaryan011:
@luisschwab That’s fair — my earlier comment wasn’t very useful.

I’ll take a closer look at the implementation and follow up with more concrete feedback.

ValuedMammal:
I would like to see more Merge tests like this for the `wallet::ChangeSet` type.

ValuedMammal:
@KrishnaShuk GitHub is complaining that your signing key is not verified.

KrishnaShuk:
Sure, I'll add wallet::ChangeSet merge tests as a follow-up commit in this PR.

> @KrishnaShuk GitHub is complaining that your signing key is not verified.

I've_ added my GPG key to GitHub. The commit is now showing as verified now.

ValuedMammal:
While we're testing ChangeSet invariants, I think we should look into #470 for this PR as well.

src/wallet/changeset.rs
```diff
@@ -417,3 +417,116 @@ impl From<locked_outpoints::ChangeSet> for ChangeSet {
...
+    }
+
+    #[test]
+    fn test_default_is_empty() {
+        let cs = ChangeSet::default();
+        assert!(cs.is_empty());
+    }
+
+    #[test]
+    fn test_is_empty_with_descriptor() {
```
ValuedMammal, line 455: It would be more clear to the reader if the test name reflected the expectation.
```suggestion
    fn test_is_non_empty_with_descriptor() {
```

src/wallet/changeset.rs
```diff
@@ -417,3 +417,116 @@ impl From<locked_outpoints::ChangeSet> for ChangeSet {
...
+        let (desc, _) = test_descriptors();
+        let cs = ChangeSet {
+            descriptor: Some(desc),
+            ..Default::default()
+        };
+        assert!(!cs.is_empty());
+    }
+
+    #[test]
+    fn test_is_empty_with_network() {
```
ValuedMammal, line 465:
```suggestion
    fn test_is_non_empty_with_network() {
```

ValuedMammal (changes requested):
The `debug_assert` in `ChangeSet::merge` only fires in debug mode, meaning reassignment is actually permitted in release builds, which is surprising.

I think the correct merge semantics should be: "If self is_none and other is_some, then proceed with the assignment, else debug_assert other is none or self equals other (to catch bugs during development)."

`cargo test -r --lib -- wallet::changeset::test`

```rs
// Merging other changeset should succeed and ignore other descriptor
#[cfg(not(debug_assertions))]
#[test]
fn test_merge_must_not_change_descriptor() {
    use crate::test_utils::*;
    use bitcoin::secp256k1::Secp256k1;
    let secp = Secp256k1::new();
    let (desc_1, _) = Descriptor::parse_descriptor(&secp, get_test_tr_single_sig()).unwrap();
    let (desc_2, _) = Descriptor::parse_descriptor(&secp, get_test_wpkh()).unwrap();

    let mut changeset = ChangeSet {
        descriptor: Some(desc_1.clone()),
        ..Default::default()
    };
    let other_descriptor_changeset = ChangeSet {
        descriptor: Some(desc_2),
        ..Default::default()
    };
    Merge::merge(&mut changeset, other_descriptor_changeset);
    assert_eq!(
        changeset.descriptor,
        Some(desc_1),
        "descriptor must never change"
    );
}
```

src/wallet/locked_outpoints.rs
```diff
@@ -24,3 +24,81 @@ impl Merge for ChangeSet {
...
+    }
+
+    #[test]
+    fn test_is_empty_default() {
+        let cs = ChangeSet::default();
+        assert!(cs.is_empty());
+    }
+
+    #[test]
+    fn test_is_empty_with_entries() {
```
Musab1258, line 49:
```suggestion
    fn test_is_non_empty_with_entries() {
```

Musab1258: > 

When I saw the test name and read through what the test tests, I was confused. @ValuedMammal made a suggestion in src/wallet/changeset.rs to update the test names so they reflect the expected outcome rather than just the method being called. I think that applies to this too.

Musab1258:
@KrishnaShuk, I was looking at how the `merge` logic can be restructured to enforce the write-once rule in release builds while keeping the helpful debug warnings. Based on @ValuedMammal's note about the `debug_assert!` vulnerability in release builds.

We could update the assignment logic for the three write-once fields (`descriptor`, `change_descriptor`, and `network`) to check if `self` is `None` before assigning. Here is a snippet of what it could look like for the `descriptor` field:

```rust
if other.descriptor.is_some() { 
    if self.descriptor.is_none() {
        self.descriptor = other.descriptor;
    } else {
        debug_assert!(
            self.descriptor == other.descriptor,
            "descriptor must never change"
        );
    }
}
```

We can fully resolve issue #470 if we apply this pattern to `descriptor`, `change_descriptor`, and `network`.

Your current tests will still pass perfectly with this change. However, to fully close this out, it would be great to add the #[cfg(not(debug_assertions))] test that @ValuedMammal suggested. This will explicitly verify that a conflicting overwrite is blocked in release builds.

PR #334, opened 2025-10-28
feat(bdk_wallet): add get_keymap() helper method to get the KeyMap for Psbt.sign

GideonBature:
### Description

This PR adds a new helper method get_keymap() to the Wallet struct that returns a combined KeyMap containing all signing keys from both external and internal (change) signers.

### Changelog notice

Added Wallet::get_keymap() helper method to retrieve a combined KeyMap from all wallet signers for convenient PSBT signing

#### All Submissions:

* [x] I've signed all my commits
* [x] I followed the [contribution guidelines](https://github.com/bitcoindevkit/bdk/blob/master/CONTRIBUTING.md)
* [x] I ran `just p` before pushing

#### New Features:

* [x] I've added tests for the new feature
* [x] I've added docs for the new feature

#### Bugfixes:

* [ ] This pull request breaks the existing API
* [ ] I've added tests to reproduce the issue which are now passing
* [ ] I'm linking the issue being fixed by this PR

Closes #331

Issue #331, opened 2025-10-18
feat(bdk_wallet): add helper method to get the KeyMap for Psbt.sign


oleonardolima:
**Describe the enhancement**  

It might be useful to use the helper method in order to get the Keymap for the existing Wallet, to be used with the new way of signing with Psbt.sign.

**Use case**  

**Impact**
- [ ] Blocking production usage
- [ ] Nice-to-have / UX improvement
- [ ] Developer experience / maintainability

**Are you using BDK in a production project?**
- [ ] Yes
- [ ] No
- [ ] Not yet, but planning to

**Which backend(s) are relevant (if any)?**
- [ ] Electrum
- [ ] Esplora
- [ ] Bitcoin Core RPC
- [ ] None / not backend-related (e.g. `bdk_chain`, `bdk_core`)
- [ ] Other (please specify): `____`

**Project or organization (optional)**

**Additional context**

GideonBature:
> How about something like this? (noting that it's still blocked by [rust-bitcoin/rust-miniscript#867](https://github.com/rust-bitcoin/rust-miniscript/pull/867)).
> 
> impl Wallet {
>     /// Sign with descriptor.
>     pub fn sign_psbt_with_descriptor<D>(
>         &self,
>         psbt: &mut Psbt,
>         descriptor: D,
>     ) -> Result<SigningKeysMap, (SigningKeysMap, SigningErrors)>
>     where
>         D: IntoWalletDescriptor,
>     {
>         let (_, keymap) = descriptor
>             .into_wallet_descriptor(&self.secp, self.network)
>             .unwrap();
>         let keymap = KeyMapWrapper::from(keymap);
>         psbt.sign(&keymap, &self.secp)
>     }
> 
>     /// Sign with key.
>     pub fn sign_psbt<K>(
>         &self,
>         psbt: &mut Psbt,
>         key: &K,
>     ) -> Result<SigningKeysMap, (SigningKeysMap, SigningErrors)>
>     where
>         K: GetKey,
>     {
>         psbt.sign(key, &self.secp)
>     }
> }

@notmandatory Seems I will have to wait for this: [rust-bitcoin/rust-miniscript#867](https://github.com/rust-bitcoin/rust-miniscript/pull/867) to be merged before I will be able to rework the PR I sent initially [#334](https://github.com/bitcoindevkit/bdk_wallet/pull/334) because the current miniscript doesn't have the KeyMapWrapper etc.

[...]

PR #438, opened 2026-04-10
feat(wallet): add Wallet::sign_psbt<K> wrapping bitcoin::Psbt::sign

EliteCoder18:
### Description
Related to #331
Incorporates get_keymap() from #334

Adds `Wallet::sign_psbt<K>`, a thin wrapper around `bitcoin::Psbt::sign` that passes the wallet's internal secp256k1 context to the caller's `GetKey` implementor (e.g. `Xpriv`, a hardware wallet adapter).

Without this, callers driving signing themselves via the bitcoin crate must construct their own `Secp256k1<All>` — setup the wallet already owns.

### Notes to the reviewers

- **sign vs sign_psbt**: `Wallet::sign` uses BDK's internal signer infrastructure (`InputSigner`/`TransactionSigner`). `sign_psbt` delegates entirely to `bitcoin::Psbt::sign`; callers supply their own `GetKey`. Both methods coexist without conflict.

### Changelog notice

- Added `Wallet::sign_psbt<K>` wrapper around `bitcoin::Psbt::sign` using the wallet's internal secp256k1 context.
- Added `Wallet::get_keymap()` to retrieve a combined KeyMap from all wallet signers for use with `KeyMapWrapper` and `Wallet::sign_psbt`.

#### All Submissions:

### DIFF
--- a/src/wallet/mod.rs
+++ b/src/wallet/mod.rs
@@ -1802,6 +1802,27 @@ impl Wallet {
         }
     }
 
+    /// Sign a PSBT using an external key provider via [`bitcoin::Psbt::sign`].
+    ///
+    /// This is a thin wrapper around [`bitcoin::Psbt::sign`] that supplies the wallet's
+    /// internal [`secp256k1`] context. It lets callers sign with any type that implements
+    /// [`psbt::GetKey`], such as [`bitcoin::bip32::Xpriv`], without having to manage their
+    /// own secp256k1 context.
+    ///
+    /// # BIP32 derivation metadata
+    ///
+    /// [`bitcoin::Psbt::sign`] uses the `bip32_derivation` fields (for legacy and segwit inputs)
+    /// or the `tap_key_origins` fields (for taproot inputs) in each PSBT input to locate
+    /// the correct child keys. PSBTs received from external coordinator tools or
+    /// hardware wallet flows typically carry this metadata already.
+    ///
+    /// # Returns
+    ///
+    /// On success, a [`psbt::SigningKeysMap`] mapping each signed input index to the keys
+    /// that were used. On failure, a tuple of the partial success map and a
+    /// [`psbt::SigningErrors`] map of per-input errors.
+    ///
+    /// [`secp256k1`]: bitcoin::secp256k1
     pub fn sign_psbt<K>(
         &self,
         psbt: &mut Psbt,
--- a/src/wallet/mod.rs
+++ b/src/wallet/mod.rs
@@ -1802,6 +1802,17 @@ impl Wallet {
         }
     }
 
+    pub fn sign_psbt<K>(
+        &self,
+        psbt: &mut Psbt,
+        key: &K,
+    ) -> Result<psbt::SigningKeysMap, (psbt::SigningKeysMap, psbt::SigningErrors)>
+    where
+        K: psbt::GetKey,
+    {
+        psbt.sign(key, &self.secp)
+    }
+
     /// Return the spending policies for the wallet's descriptor.
     pub fn policies(&self, keychain: KeychainKind) -> Result<Option<Policy>, DescriptorError> {
         let signers = match keychain {
--- a/src/wallet/mod.rs
+++ b/src/wallet/mod.rs
@@ -1853,6 +1853,38 @@ impl Wallet {
         }
     }
 
+    /// Sign a PSBT using an external key provider via [`bitcoin::Psbt::sign`].
+    ///
+    /// This is a thin wrapper around [`bitcoin::Psbt::sign`] that supplies the wallet's
+    /// internal [`secp256k1`] context. It lets callers sign with any type that implements
+    /// [`psbt::GetKey`], such as [`bitcoin::bip32::Xpriv`], without having to manage their
+    /// own secp256k1 context.
+    ///
+    /// # BIP32 derivation metadata
+    ///
+    /// [`bitcoin::Psbt::sign`] uses the `bip32_derivation` fields (for legacy and segwit inputs)
+    /// or the `tap_key_origins` fields (for taproot inputs) in each PSBT input to locate
+    /// the correct child keys. PSBTs received from external coordinator tools or
+    /// hardware wallet flows typically carry this metadata already.
+    ///
+    /// # Returns
+    ///
+    /// On success, a [`psbt::SigningKeysMap`] mapping each signed input index to the keys
+    /// that were used. On failure, a tuple of the partial success map and a
+    /// [`psbt::SigningErrors`] map of per-input errors.
+    ///
+    /// [`secp256k1`]: bitcoin::secp256k1
+    pub fn sign_psbt<K>(
+        &self,
+        psbt: &mut Psbt,
+        key: &K,
+    ) -> Result<psbt::SigningKeysMap, (psbt::SigningKeysMap, psbt::SigningErrors)>
+    where
+        K: psbt::GetKey,
+    {
+        psbt.sign(key, &self.secp)
+    }
+
     /// Return the spending policies for the wallet's descriptor.
     pub fn policies(&self, keychain: KeychainKind) -> Result<Option<Policy>, DescriptorError> {
         let signers = match keychain {
--- a/src/wallet/mod.rs
+++ b/src/wallet/mod.rs
@@ -1853,6 +1853,17 @@ impl Wallet {
         }
     }
 
+    /// Returns a combined [`KeyMap`] from all signing keys loaded in the wallet.
+    ///
+    /// The map merges keys from both the external (receive) and internal (change) keychains.
+    /// It can be used with [`miniscript::descriptor::KeyMapWrapper`] to sign a PSBT via
+    /// [`Wallet::sign_psbt`] using the wallet's own keys.
+    pub fn get_keymap(&self) -> KeyMap {
+        let mut keymap = self.signers.as_key_map(&self.secp);
+        keymap.extend(self.change_signers.as_key_map(&self.secp));
+        keymap
+    }
+
     /// Sign a PSBT using an external key provider via [`bitcoin::Psbt::sign`].
     ///
     /// This is a thin wrapper around [`bitcoin::Psbt::sign`] that supplies the wallet's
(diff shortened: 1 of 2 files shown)


ValuedMammal:
Concept ACK. I've not had an extensive look at the implementation.

> Callers must first run wallet.sign()

`sign_psbt` should be written in a way that assumes the PSBT has been updated and totally independent of `wallet.sign()`. For the tests you can create the PSBT to sign by parsing it from a string in base64.

EliteCoder18:
@ValuedMammal Thanks for the feedback!

I've updated the tests to use a base64 PSBT with pre-populated `bip32_derivation`, so `sign_psbt` is now fully independent of `wallet.sign`.

Would appreciate a review when you get a chance.

src/wallet/mod.rs
```diff
@@ -1802,6 +1802,37 @@ impl Wallet {
...
+    ///
+    /// This is a thin wrapper around [`bitcoin::Psbt::sign`] that supplies the wallet's
+    /// internal [`secp256k1`] context. It lets callers sign with any type that implements
+    /// [`psbt::GetKey`], such as [`bitcoin::bip32::Xpriv`], without having to manage their
+    /// own secp256k1 context.
+    ///
+    /// # BIP32 derivation metadata
+    ///
+    /// [`bitcoin::Psbt::sign`] uses the `bip32_derivation` fields in each PSBT input to
+    /// locate the correct child keys. PSBTs received from external coordinator tools or
```
ValuedMammal: > the `bip32_derivation` fields in each PSBT input to locate the correct child keys.

Or `tap_key_origins` in the case of taproot.

ValuedMammal (commented):
Concept ACK

oleonardolima (commented):
cACK ce3d1894a7eca014409d19e11008ef603576e78c

src/wallet/mod.rs
EliteCoder18: Updated the doc comment to mention both `bip32_derivation` (legacy/segwit) and `tap_key_origins` (taproot). Thanks for catching that!

src/wallet/mod.rs
```diff
@@ -1802,6 +1802,38 @@ impl Wallet {
...
+    /// the correct child keys. PSBTs received from external coordinator tools or
+    /// hardware wallet flows typically carry this metadata already.
+    ///
+    /// # Returns
+    ///
+    /// On success, a [`psbt::SigningKeysMap`] mapping each signed input index to the keys
+    /// that were used. On failure, a tuple of the partial success map and a
+    /// [`psbt::SigningErrors`] map of per-input errors.
+    ///
+    /// [`secp256k1`]: bitcoin::secp256k1
```
noahjoeris, line 1904: nit: What do you think of those docs changes? I added that it's used for software signing and pointed to the correct approach for external signing.
```suggestion
    /// Sign a PSBT using an external key provider via [`bitcoin::Psbt::sign`].
    ///
    /// This is a thin wrapper around [`bitcoin::Psbt::sign`] that supplies the wallet's
    /// internal [`secp256k1`] context. It lets callers sign with any type that implements
    /// [`psbt::GetKey`], such as [`bitcoin::bip32::Xpriv`], without having to manage their
    /// own secp256k1 context. Use it for software signing.
    ///
    /// # BIP32 derivation metadata
    ///
    /// [`bitcoin::Psbt::sign`] uses the `bip32_derivation` fields (for legacy and segwit inputs)
    /// or the `tap_key_origins` fields (for taproot inputs) in each PSBT input to locate
    /// the correct child keys. PSBTs constructed from within the wallet, or PSBTs received from
    /// external coordinator tools or hardware wallet flows, typically carry this metadata already.
    ///
    /// # Returns
    ///
    /// On success, a [`psbt::SigningKeysMap`] mapping each signed input index to the keys
    /// that were used. On failure, a tuple of the partial success map and a
    /// [`psbt::SigningErrors`] map of per-input errors.
    ///
    /// [`secp256k1`]: bitcoin::secp256k1
    /// [`TransactionSigner`]: crate::signer::TransactionSigner
```

noahjoeris (commented):
cACK 0884a8d
Thanks for adding this. It will be useful after this is merged: rust-bitcoin/rust-miniscript#872

src/wallet/mod.rs
oleonardolima: I'm not sure if we should refer to `TransactionSigner`, as it'll be later removed with signers module.

The hardware signers should handle the PSBT and signing keys on their own, considering that we had archived https://github.com/bitcoindevkit/rust-hwi and it's considered unmaintained.

oleonardolima (commented):
cACK 0884a8d761b2091b270d63bc6063454f55c0945d

I suggest squashing the doc commit into the first one, and also updating the PR description to reflect the current state.

src/wallet/mod.rs
oleonardolima: > I'm not sure if we should refer to `TransactionSigner`, as it'll be later removed with signers module.

Anyway, we can add it now and remove it when removing the module.

oleonardolima:
nit: I think it's a good idea to also add the helper method `get_keymap` here, you can cherry-pick it from bitcoindevkit/bdk_wallet#334.

As you add the helper method here, you can add an example of how to use in the documentation of `sign_psbt`.

NOTE: that you'd need to use the `KeyMapWrapper`, as we're still in miniscript 12.x

src/wallet/mod.rs
noahjoeris: Yep thanks, was thinking about it too. 
From my understanding we added `sign_with_signers` as a transition step to keep supporting the bdk signer for some more time while moving to psbt.sign. 
Which means yeah agree we shouldn't refer to it.

EliteCoder18:
Happy to add `get_keymap`, but since `KeyMapWrapper` is a temporary shim until rust-miniscript#872 lands, I'd rather not bake it into the docs here and then need to update it again in 4.0. planning to add `get_keymap` as a proper follow-up once the clean descriptor-aware path is available. Does that work for you?

> nit: I think it's a good idea to also add the helper method `get_keymap` here, you can cherry-pick it from bitcoindevkit/bdk_wallet#334.
> 
> As you add the helper method here, you can add an example of how to use in the documentation of `sign_psbt`.
> 
> NOTE: that you'd need to use the `KeyMapWrapper`, as we're still in miniscript 12.x

oleonardolima:
> Happy to add `get_keymap`, but since `KeyMapWrapper` is a temporary shim until rust-miniscript#872 lands, I'd rather not bake it into the docs here and then need to update it again in 4.0. planning to add `get_keymap` as a proper follow-up once the clean descriptor-aware path is available. Does that work for you?
> 

This PR is aimed to `3.x.x` release, so it'll require the `KeyMapWrapper` anyway, either here or in the user side. It's best if we have all the ergonomic and convenience methods to guide the users onto how to use the new API, as per bitcoindevkit/book-of-bdk#159

EliteCoder18:
Thanks for the suggestion. I've added `get_keymap()` along with a test, and updated the `sign_psbt` docs with an example showing the `get_keymap` → `KeyMapWrapper` flow.

I left out a `KeyMapWrapper` re-export for now to keep the public API surface minimal, since it can already be imported directly from miniscript.

> This PR is aimed to `3.x.x` release, so it'll require the `KeyMapWrapper` anyway, either here or in the user side. It's best if we have all the ergonomic and convenience methods to guide the users onto how to use the new API, as per bitcoindevkit/book-of-bdk#159

tests/psbt.rs
```diff
@@ -221,3 +223,62 @@ fn test_psbt_multiple_internalkey_signers() {
...
+        keys_used, 0,
+        "expected zero keys used when signing with an unrelated xpriv"
+    );
+    for input in &psbt.inputs {
+        assert!(
+            input.partial_sigs.is_empty(),
+            "expected no partial signatures when signing with an unrelated key"
+        );
+    }
+}
```
oleonardolima, line 284: nit: you could also add a test that uses the `get_keymap` and sign with `Wallet` keys here.

oleonardolima (approved):
ACK 40a8cac4fa93540a81d827c35ebbe75738945816

oleonardolima:
cc'in @thunderbiscuit as you're interested in the new signing flow.

noahjoeris:
Why do we want `get_keymap` if we plan to remove wallet-owned signers? 
This is adding functionality that is already planned to be deprecated soon.

notmandatory:
> Why do we want `get_keymap` if we plan to remove wallet-owned signers? This is adding functionality that is already planned to be deprecated soon.

This is a good point. If this is something we can do without and will have to later remove then we shouldn't do it at all. Based on the docs in bitcoindevkit/book-of-bdk#155 it looks like we don't need this helper.

oleonardolima:
> Why do we want `get_keymap` if we plan to remove wallet-owned signers? This is adding functionality that is already planned to be deprecated soon.

Yes, it'll be removed in the next breaking change when removing the signers, though it's a good helper (not a MUST) method for current users to use the `Wallet::sign_psbt<K>` at current state, otherwise they would need to handle that themselves.

oleonardolima:
> > Why do we want `get_keymap` if we plan to remove wallet-owned signers? This is adding functionality that is already planned to be deprecated soon.
> 
> This is a good point. If this is something we can do without and will have to later remove then we shouldn't do it at all. Based on the docs in [bitcoindevkit/book-of-bdk#155](https://github.com/bitcoindevkit/book-of-bdk/pull/155) it looks like we don't need this helper.

Won't the `SignersContainers` in the example there also be removed in the future ?

EliteCoder18:
Thanks for the discussion everyone. I am happy to follow whatever direction the maintainers prefer here.

My understanding: @oleonardolima originally requested `get_keymap` as a convenience for current 3.x.x users who want to use `sign_psbt` with wallet-owned keys, while @noahjoeris and @notmandatory are concerned about adding surface that'll need removing once PR bitcoindevkit/bdk_wallet#505 lands.

If the consensus is to drop `get_keymap` and ship just `sign_psbt`...I can do that. The core API isn't affected and the two existing tests still cover the main cases. Let me know which path we'd prefer for the 3.x.x release, and I'll update the branch accordingly

noahjoeris:
Thanks so much for your effort with this!

In our last bdk call, we discussed that it might be better to directly use `psbt.sign` and avoid adding this wrapper function.

notmandatory:
I had a chat with @EliteCoder18 and after looking at bitcoindevkit/book-of-bdk#155 and given that we're trying to move users away from keeping private key descriptors in their wallet I think we should close this PR. We can discuss on the call tomorrow.

ValuedMammal:
Closing per team discussions

PR #442, opened 2026-04-13
fix(wallet): exclude unconfirmed v3 outputs from non-v3 coin selection

### LINKED ISSUE #419: Coin selection includes unconfirmed v3/TRUC outputs in v2 transactions, causing mempool rejection
**Describe the bug**  

When a v3 (TRUC) transaction produces a change output back into the BDK wallet, that unconfirmed output enters BDK's UTXO pool like any other coin. Bitcoin Core's TRUC policy (BIP-431) requires that any transaction spending an unconfirmed v3 output must itself be v3. BDK's coin selection is unaware of this constraint and freely includes unconfirmed v3 change outputs when building standard v2 transactions, which bitcoind then rejects with `TRUC-violation, non-version=3 tx cannot spend from version=3 tx`.

Once the v3 parent confirms, its outputs become ordinary UTXOs spendable by any transaction version — so the issue only manifests while the v3 parent is still in the mempool.

**To Reproduce**  

  1. Create and broadcast a v3 (TRUC) transaction that returns change to the BDK wallet (e.g. a CPFP child for an Ark exit).
  2. Sync the wallet so BDK sees the unconfirmed v3 change output.
  3. Use TxBuilder to build a standard (v2) transaction — BDK's coin selection may pick the unconfirmed v3 output.
  4. Sign and broadcast → bitcoind rejects with:
  `sendrawtransaction RPC error -26: TRUC-violation, non-version=3 tx <v2_txid>
  cannot spend from version=3 tx <v3_txid>`

**Expected behavior**  

BDK should either:
  - Automatically exclude unconfirmed v3 outputs from coin selection when building non-v3 transactions, or
  - Expose enough metadata (parent transaction version) for callers to filter them out, or
  - Provide a TxBuilder option to exclude UTXOs whose unconfirmed parent is v3

  Note: `TxBuilder::exclude_unconfirmed` (added in 2.1.0) is a partial workaround but too broad — it excludes unconfirmed UTXOs, not just those subject to TRUC rules.

**Build environment**  
- BDK tag/commit:  bdk_wallet 2.1.0
- OS+version:   macOS 26.3.1 (aarch64)
- Rust/Cargo version:  1.90.0
- Rust/Cargo target:   aarch64-apple-darwin

- [ ] Electrum
- [X] Esplora
- [X] Bitcoin Core RPC
- [ ] None / not backend-related (e.g. `bdk_chain`, `bdk_core`)
- [ ] Other (please specify): `____`

- [X] Yes
- [ ] No

Ark (Second) — during Ark exits, a CPFP child (v3) is broadcast to bump the exit transaction. Its change output lands in the on-chain wallet, and any subsequent on-chain send fails until the CPFP confirms. Exit tx can claimed/secured once the CPFP confirms.

**Additional context**  

Reproduction evidence from signet:
```
  VTXO exited:  d4c0eb4e49a064c6d8224efe4e384aa9c8f746dc785ba7cfe07c5204560120a3:0
  CPFP child:   77274504eae4eeb0d82af9cee679cad918c67f9b64a40764450ec41caa0c354c
  Change UTXO:  77274504...354c:0 (9723 sat, unconfirmed, v3)
  Failed send:  a7d53ec248ecdd0d264589657ddfa054052e2e4d90e06b99e5d26d4eb0395b74 (non-v3)

  Error: "sendrawtransaction RPC error -26: TRUC-violation, non-version=3 tx a7d53ec2...
         cannot spend from version=3 tx 77274504..."
```

EliteCoder18:
### Description
BIP-431 (TRUC) requires that any transaction spending an unconfirmed v3 output must itself be v3. BDK's coin selection had no awareness of this and would freely include unconfirmed v3 parent outputs when building standard non-v3 transactions, causing bitcoind to reject the broadcast.

This PR threads `tx_version` into `filter_utxos` to exclude such UTXOs when the target transaction is not v3. Confirmed outputs are unaffected.

Fixes #419.

### Notes to the reviewers
The fix is intentionally minimal only the cross-version spending restriction from BIP-431 is addressed here. The broader TRUC topological constraints (cluster size, child vbyte limit) are out of scope for this PR and can be tracked separately.

Manually selected UTXOs via `TxBuilder::add_utxo` bypass `filter_utxos` entirely and are not covered by this fix. Callers are responsible for not adding unconfirmed v3 outputs to non-v3 transactions manually.

### Changelog notice
- Fixed: `filter_utxos` now excludes unconfirmed UTXOs from v3 parent transactions when constructing non-v3 transactions, preventing TRUC policy violations at broadcast.

#### All Submissions:

#### Bugfixes:

### DIFF
--- a/src/wallet/mod.rs
+++ b/src/wallet/mod.rs
@@ -1983,13 +1983,14 @@ impl Wallet {
             vec![]
         // Only process optional UTxOs if manually_selected_only is false.
         } else {
+            let tx_graph = self.tx_graph.graph();
+            let target_is_truc = params.version.unwrap_or(transaction::Version::TWO).0 == 3;
             let manually_selected_outpoints = params
                 .utxos
                 .iter()
                 .map(|wutxo| wutxo.utxo.outpoint())
                 .collect::<HashSet<OutPoint>>();
-            self.tx_graph
-                .graph()
+            tx_graph
                 // Get all unspent UTxOs from wallet.
                 // NOTE: the UTxOs returned by the following method already belong to wallet as the
                 // call chain uses get_tx_node infallibly.
@@ -2022,6 +2023,14 @@ impl Wallet {
                 // If bumping fees only add to optional UTxOs those confirmed.
                 .filter(|local_output| {
                     params.bumping_fee.is_none() || local_output.chain_position.is_confirmed()
+                })
+                // Bitcoin Core's TRUC policy requires a transaction and each unconfirmed parent it
+                // spends from to both be either TRUC (version 3) or both non-TRUC.
+                .filter(|local_output| {
+                    local_output.chain_position.is_confirmed()
+                        || tx_graph
+                            .get_tx(local_output.outpoint.txid)
+                            .is_some_and(|tx| (tx.version.0 == 3) == target_is_truc)
                 })
                 .map(|utxo| WeightedUtxo {
                     satisfaction_weight: self
--- a/src/wallet/coin_selection.rs
+++ b/src/wallet/coin_selection.rs
@@ -471,10 +471,10 @@ impl<Cs: CoinSelectionAlgorithm> CoinSelectionAlgorithm for BranchAndBoundCoinSe
             .iter()
             .fold(SignedAmount::ZERO, |acc, x| acc + x.effective_value);
 
-        let cost_of_change = (Weight::from_vb(self.size_of_change).expect("overflow occurred")
-            * fee_rate)
-            .to_signed()
-            .expect("signed amount");
+        let cost_of_change = (fee_rate
+            * Weight::from_vb(self.size_of_change).expect("overflow occurred"))
+        .to_signed()
+        .expect("signed amount");
 
         // `curr_value` and `curr_available_value` are both the sum of *effective_values* of
         // the UTXOs. For the optional UTXOs (curr_available_value) we filter out UTXOs with
@@ -940,7 +940,7 @@ mod test {
             .coin_select(
                 utxos,
                 vec![],
-                FeeRate::from_sat_per_vb_u32(1),
+                FeeRate::from_sat_per_vb(1).unwrap(),
                 target_amount,
                 &drain_script,
                 &mut thread_rng(),
@@ -962,7 +962,7 @@ mod test {
             .coin_select(
                 utxos,
                 vec![],
-                FeeRate::from_sat_per_vb_u32(1),
+                FeeRate::from_sat_per_vb(1).unwrap(),
                 target_amount,
                 &drain_script,
                 &mut thread_rng(),
@@ -984,7 +984,7 @@ mod test {
             .coin_select(
                 vec![],
                 utxos,
-                FeeRate::from_sat_per_vb_u32(1),
+                FeeRate::from_sat_per_vb(1).unwrap(),
                 target_amount,
                 &drain_script,
                 &mut thread_rng(),
@@ -1005,7 +1005,7 @@ mod test {
         let result = LargestFirstCoinSelection.coin_select(
             vec![],
             utxos,
-            FeeRate::from_sat_per_vb_u32(1),
+            FeeRate::from_sat_per_vb(1).unwrap(),
             target_amount,
             &drain_script,
             &mut thread_rng(),
@@ -1022,7 +1022,7 @@ mod test {
         let result = LargestFirstCoinSelection.coin_select(
             vec![],
             utxos,
-            FeeRate::from_sat_per_vb_u32(1000),
+            FeeRate::from_sat_per_vb(1000).unwrap(),
             target_amount,
             &drain_script,
             &mut thread_rng(),
@@ -1040,7 +1040,7 @@ mod test {
             .coin_select(
                 vec![],
                 utxos,
-                FeeRate::from_sat_per_vb_u32(1),
+                FeeRate::from_sat_per_vb(1).unwrap(),
                 target_amount,
                 &drain_script,
                 &mut thread_rng(),
@@ -1062,7 +1062,7 @@ mod test {
             .coin_select(
                 utxos,
                 vec![],
-                FeeRate::from_sat_per_vb_u32(1),
+                FeeRate::from_sat_per_vb(1).unwrap(),
                 target_amount,
                 &drain_script,
                 &mut thread_rng(),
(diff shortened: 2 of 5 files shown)


src/wallet/mod.rs
```diff
@@ -1978,18 +1978,24 @@ impl Wallet {
 
     /// Given the options returns the list of utxos that must be used to form the
     /// transaction and any further that may be used if needed.
-    fn filter_utxos(&self, params: &TxParams, current_height: u32) -> Vec<WeightedUtxo> {
+    fn filter_utxos(
+        &self,
+        params: &TxParams,
+        current_height: u32,
+        tx_version: transaction::Version,
```
oleonardolima: You don't need to add a new argument, it's already available in `TxParams`.

oleonardolima (commented):
Thanks for working on this one!

I'm still wondering if it's the best  way to address the issue. I'll give it a try on some other alternatives.

yan-pi (commented):
First I reproduced #419 against pre-fix bdk-cli and bitcoind rejects on broadcast as expected.

Validated this branch with a local path dep: 
`create_tx` returns `InsufficientFunds`, `available == confirmed` matched exactly.

Covered direction works as advertised.

That said, `Fixes #419` closes the issue on merge, but some gaps remain:

- v3 -> non-v3 unconfirmed (other half of BIP-431 Rule 2).
Reproducible on this branch.
- `add_utxo` / `add_foreign_utxo` bypass: PR body mentions, rustdoc doesn't.
- vB rules ([BIP-431](https://github.com/bitcoin/bips/blob/master/bip-0431.mediawiki#specification) Rules 3/4/5): out of scope in this PR. 
But oleonardolima asked for this in the issue.

I think you should either expand scope, or change body to "addresses part of #419" and link follow-ups before merge, dunno about how this last one is handled in BDK org.

src/wallet/mod.rs
```diff
@@ -1429,7 +1429,7 @@ impl Wallet {
         let (required_utxos, optional_utxos) = {
             // NOTE: manual selection overrides unspendable
             let mut required: Vec<WeightedUtxo> = params.utxos.clone();
-            let optional = self.filter_utxos(&params, current_height.to_consensus_u32());
+            let optional = self.filter_utxos(&params, current_height.to_consensus_u32(), version);
```
yan-pi: The bypass for `add_utxo` / `add_foreign_utxo` is mentioned in the PR body, but not in the rustdoc of those methods.

IMO worth a note.
An integrator using manual selection might assume "fixes #419" covers them and still hit TRUC-violation.

src/wallet/mod.rs
```diff
@@ -2023,6 +2029,15 @@ impl Wallet {
...
                 })
+                // Bitcoin Core's TRUC policy requires spending unconfirmed v3 outputs with a v3
+                // transaction.
+                .filter(|local_output| {
+                    !excludes_unconfirmed_v3
+                        || local_output.chain_position.is_confirmed()
+                        || tx_graph
+                            .get_tx(local_output.outpoint.txid)
+                            .is_none_or(|tx| tx.version != transaction::Version(3))
+                })
```
yan-pi, line 2034: BIP-431 Rule 2 has two clauses, this predicate only covers one:

> Any TRUC transaction's unconfirmed ancestors must all be TRUC. Any descendant of an unconfirmed TRUC transaction must also be TRUC.

When target is v3, the filter is off: coin selection can pick non-v3 unconfirmed UTXOs and bitcoind rejects on broadcast.

Same shape as #419, other direction.

Reproduced on this branch:

1.5 BTC confirmed + 0.3 BTC v2 unconfirmed, target v3 - 1.6 BTC -> PSBT includes the v2 unconfirmed.

```
Attempting broadcast (expected: TRUC-violation)...
  bitcoind rejected: JSON-RPC error: RPC error response: RpcError { code: -26, message: "TRUC-violation, version=
3 tx 6d60a9caac7631812c5ed083c8c583d3cd9382485640aa7a543d3c5d19a00419 (wtxid=726b4a450e32084d5237c8a098f9e26c399b
e7f27ae3af7d328bb000ddb4aa19) cannot spend from non-version=3 tx 8ff29f49e540dbd32c9d1207628861504e0f981a1f5a2025
b6550bd380118724 (wtxid=38b52d8ad666a8296a4c4383e4572dc8b45cc796203e85d39033b772cabc1242)", data: None }

```

Not sure if you'd want to fix here or create a follow-up issue/pr?

src/wallet/mod.rs
EliteCoder18: Done.... removed the extra argument and reading from `TxParams.version` directly now.

src/wallet/mod.rs
EliteCoder18: Thanks for calling this out.... That's a great point. I’ve now added explicit rustdoc notes on both `TxBuilder::add_utxo and TxBuilder::add_foreign_utxo` clarifying that manually selected UTXOs bypass optional-UTXO filtering (including TRUC version-compatibility checks), so callers must validate unconfirmed manual inputs against the target tx version themselves.

This should make the behavior clear for integrators and avoid a false assumption that the #419 fix also covers manual selection paths.

src/wallet/mod.rs
EliteCoder18: The bidirectional filter is included in the latest revision both directions are covered by `(tx.version.0 == 3) == target_is_truc`. Could you re-test against the current HEAD? Happy to look into it further if the violation still reproduces.

src/wallet/mod.rs
ValuedMammal: Yes but we already validated the transaction version around line 1294, so not re-initializing it with `unwrap_or(Version::TWO)` would also make sense?

EliteCoder18: Good point, thanks!

You're right, we already validate and resolve the transaction version earlier in `create_tx`. Re-doing `unwrap_or(Version::TWO)` inside `filter_utxos` is redundant.

I'll update it to pass the already-resolved `version` directly into `filter_utxos` from the call site instead of re-deriving it inside the function.

EliteCoder18:
Thanks everyone for the feedback and discussion. After looking at the broader BIP-431 implementation, it's clear that #477 provides a more complete foundation than the approach taken here.

To avoid duplicating work, I'm closing this PR. The core fix has already been implemented in #478, and I'll be focusing my efforts on reviewing and contributing to #477 and #478 going forward.

Thanks again for the collaboration!

PR #443, opened 2026-04-14
test(utils): add boundary tests for Older::check_older and After::check_after

KrishnaShuk:
### Description

Addresses the `// TODO: test >= / >` on line 111 of [utils.rs](cci:7://file:///home/krishna/Contribution/bdk_wallet/src/test_utils.rs:0:0-0:0). Adds 11 boundary-condition tests for the `Satisfier` impls of [Older](cci:2://file:///home/krishna/Contribution/bdk_wallet/src/wallet/utils.rs:87:0-91:1) and [After](cci:2://file:///home/krishna/Contribution/bdk_wallet/src/wallet/utils.rs:43:0-46:1).

### Notes to the reviewers

Uses `Satisfier::<PublicKey>::check_older(...)` fully-qualified syntax because the trait is generic over `Pk`.

#### All Submissions:

#### New Features:

#### Bugfixes:

### DIFF
--- a/src/wallet/utils.rs
+++ b/src/wallet/utils.rs
@@ -278,4 +278,116 @@ mod test {
         shuffle_slice(&mut test, &mut rng);
         assert_eq!(test, &[0, 4, 1, 2, 5]);
     }
+
+    // --- Older::check_older boundary tests (addresses TODO on line 111) ---
+
+    use super::{After, Older};
+    use bitcoin::{absolute, relative, PublicKey};
+    use miniscript::Satisfier;
+
+    #[test]
+    fn test_check_older_at_boundary() {
+        // current_height == create_height + n → should pass (>=)
+        let older = Older::new(Some(110), Some(100), false);
+        assert!(Satisfier::<PublicKey>::check_older(
+            &older,
+            relative::LockTime::from_height(10)
+        ));
+    }
+
+    #[test]
+    fn test_check_older_above_boundary() {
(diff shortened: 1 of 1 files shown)


src/wallet/utils.rs
```diff
@@ -278,4 +278,116 @@ mod test {
...
+
+    // --- After::check_after boundary tests ---
+
+    #[test]
+    fn test_check_after_at_boundary() {
+        // current_height == locktime → should pass (>=)
+        let after = After::new(Some(100_000), false);
+        assert!(Satisfier::<PublicKey>::check_after(
+            &after,
+            absolute::LockTime::from_consensus(100_000)
```
busayo-OD, line 354: `absolute::LockTime` can also represent Unix timestamps (>= 500_000_000), but `After` only has `current_height` and no `current_time`. So a timestamp-based locktime would end up being compared against a block height. I also noticed the existing tests don't cover this. Is this intentionally limited to height-based locktimes for now?

busayo-OD: I traced the satisfaction flow further. Timestamp CLTV finalization works; existing wallet tests cover that path, and PsbtInputSatisfier::check_after() evaluates the transaction's lock_time via is_implied_by, which is locktime-type aware.

One thing I noticed: After::check_after() does current_height >= n.to_consensus_u32(), which doesn't distinguish between height- and timestamp-based locktimes. Existing timestamp CLTV tests still pass through the signing/finalization flow.

PR #445, opened 2026-04-15
fix(types): use bounds-checked access in `Utxo::txout()` for Foreign variant

tnull:
### Description

Replace unchecked array indexing of `prev_tx.output[outpoint.vout]` with `.get().expect()` to provide a descriptive panic message instead of an opaque index-out-of-bounds if the outpoint vout is invalid.

Co-Authored-By: HAL 9000

#### All Submissions:

#### Bugfixes:

### DIFF
--- a/src/types.rs
+++ b/src/types.rs
@@ -124,7 +124,10 @@ impl Utxo {
                 ..
             } => {
                 if let Some(prev_tx) = &psbt_input.non_witness_utxo {
-                    return &prev_tx.output[outpoint.vout as usize];
+                    return prev_tx
+                        .output
+                        .get(outpoint.vout as usize)
+                        .expect("outpoint vout must be in bounds");
                 }
 
                 if let Some(txout) = &psbt_input.witness_utxo {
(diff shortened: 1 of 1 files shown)


luisschwab:
cACK, can you fix the clippy warning?

```rs
let psbt_input = psbt::Input {
    non_witness_utxo: Some(prev_tx),
    ..Default::default()
};
```

tnull:
> cACK, can you fix the clippy warning?
> 
> ```rust
> let psbt_input = psbt::Input {
>     non_witness_utxo: Some(prev_tx),
>     ..Default::default()
> };
> ```

Done.

luisschwab (approved):
tACK 2bd536f1b752016688b9794bff22f722f2c620ab

ValuedMammal (commented):
I suppose this is marginally better than before. I'm not too concerned with asserting the content of the panic message string. If anything the method should state how it panics in the documentation. Noting that it generally works when combined with `add_foreign_utxo`, by throwing for MissingUtxo, since we can't control how the foreign Utxo is constructed in all cases, I guess the right solution long term is to handle a missing txout using Option or Result.

ValuedMammal:
Fixed in #487

PR #448, opened 2026-04-18
fix(wallet): retry coin selection on NoChange to avoid dust/zero drai…

### LINKED ISSUE #376: Spending P2A Output with a Non-Zero Value can Return CreateTxError::CoinSelection
**Describe the bug**  
We use BDK as our onchain wallet for the Bark project, an important part of that is using BDK to spend P2A outputs when users perform an emergency exit. Each "exit transaction" contains a P2A output, typically with a value of zero sats, but sometimes with a non-zero value due to how we manage fees for certain operations. When the P2A value is zero there's no problem, however when it's non-zero you can run into edge cases which result in unexpected errors.

Here's an example scenario:
- The wallet has 1 UTXO of 1,000,000 sats
- We try to create a new transaction to spend the P2A output of an exit transaction which isn't contained in the wallet itself
- We add a `drain_to` address which is the BDK wallet
- We set `absolute_fee` to 200 sats (enough to broadcast the exit transaction and the transaction we're constructing as a TRUC package)
- If the P2A output has a value of 0:
  - BDK will pull in the 1,000,000 sat UTXO and drain to a wallet address with a 999,800 sat output
  - This is as expected
- If the P2A output has a value of 1,000:
  - BDK won't use any extra UTXOs and will drain to a wallet address with a 800 sat output
  - This is as expected
- If the P2A output has a value of 500:
  - BDK won't use any extra UTXOs and will error because it can't create a 300 sat output as that is below the dust limit
  - This is not as expected
- If the P2A output has a value of 200:
  - BDK won't use any extra UTXOs and will error because the remaining value is 0 and it enforces the use of at least 1 TxOut even when draining
  - This is not as expected

**Expected behavior**  
The documentation for `fee_absolute` states the following:
> Note that this is really a minimum absolute fee -- it's possible to overshoot it slightly since adding a change output to drain the remaining excess might not be viable.

Under that assumption I would expect that when spending a P2A output which results in zero or dust change, the entire P2A value should be consumed and no output should be created. If BDK mandates there always being at least 1 TxOut, then I would expect the transaction builder to pull in the extra UTXO to drain to a non-dust output.

**Build environment**  
- BDK tag/commit: 2.3.0
- OS+version: macOS 26.2
- Rust/Cargo version: 1.90
- Rust/Cargo target: aarch64-apple-darwin

- [x] Esplora
- [x] Bitcoin Core RPC

- [x] No: We work around it by trying to detect this problem and increase the fee amount to force 1 sat to be used from another UTXO

[Bark](https://gitlab.com/ark-bitcoin/bark)

**Additional context**  
[This is the source code](https://gitlab.com/ark-bitcoin/bark/-/blob/master/bitcoin-ext/src/bdk.rs?ref_type=heads#L157) for where we encounter this issue.

alienx5499:
### Description

Fixes #376.

`Wallet::create_tx` could return `CreateTxError::CoinSelection` for drain-only transactions (`drain_to`, **no explicit recipients**) when spending a foreign input (e.g P2A) with `fee_absolute`.

In that situation, the initial `coin_select` pass may select only required inputs and yield `Excess::NoChange` (i.e. zero or dust remaining for the drain output). This results in a coin selection failure despite sufficient total funds, because the drain-only path cannot produce a non-dust output under `Excess::NoChange`.

This change adds a retry for that scenario : optional wallet UTXOs are promoted to required inputs one at a time and `coin_select` is run again until `Excess` is no longer `NoChange` or there are no optional UTXOs left. That yields a valid (non-dust) drain output when possible.

**Scope:** General coin selection behavior is unchanged. This only adds a retry for the `NoChange` + drain-only case. Optional UTXOs are taken in existing order (`pop()`), so no new selection policy is introduced.

### Notes to the reviewers

- **Guards:** Retry is gated on:
  - no explicit recipients (`params.recipients.is_empty()`)
  - `drain_to` is set
  - `(params.drain_wallet || !params.utxos.is_empty())` (preserves `NoRecipients` behavior vs the empty-output path in `mod.rs`)
  - not `manually_selected_only`
- **Tests:** `tests/drain_to_dust_pull_utxo.rs`
  - Foreign P2A-style UTXOs via `insert_txout`, `fee_absolute`, `drain_to`
  - Covers dust remainder (500 sat value, 400 sat fee) and zero remainder (200 sat value, 200 sat fee)
  - Asserts at least 2 inputs (foreign + local) non-empty outputs, foreign outpoint present in inputs

### Changelog notice

`fix(wallet): retry coin selection on NoChange to avoid dust/zero drain outputs (#376)`

#### All Submissions:

#### New Features:

#### Bugfixes:

### DIFF
--- a/src/wallet/mod.rs
+++ b/src/wallet/mod.rs
@@ -1473,31 +1473,40 @@ impl Wallet {
         let should_retry_for_dust_drain = params.recipients.is_empty()
             && params.drain_to.is_some()
             && (params.drain_wallet || !params.utxos.is_empty())
+            && !optional_utxos.is_empty()
             && !params.manually_selected_only;
 
         let selection_result = if should_retry_for_dust_drain {
             let mut required_for_attempt = required_utxos;
             let mut optional_remaining = optional_utxos;
+            let mut last_successful_result = None;
             loop {
-                let result = coin_selection
-                    .coin_select(
-                        required_for_attempt.clone(),
-                        optional_remaining.clone(),
-                        fee_rate,
-                        outgoing + fee_amount,
-                        &drain_script,
-                        rng,
-                    )
-                    .map_err(CreateTxError::CoinSelection)?;
+                match coin_selection.coin_select(
+                    required_for_attempt.clone(),
+                    optional_remaining.clone(),
+                    fee_rate,
+                    outgoing + fee_amount,
+                    &drain_script,
+                    rng,
+                ) {
+                    Ok(result) => {
+                        if !matches!(&result.excess, Excess::NoChange { .. }) {
+                            break result;
+                        }
 
-                if !matches!(&result.excess, Excess::NoChange { .. }) {
-                    break result;
+                        let Some(w) = optional_remaining.pop() else {
+                            break result;
+                        };
+                        last_successful_result = Some(result);
+                        required_for_attempt.push(w);
+                    }
+                    Err(err) => {
+                        if let Some(result) = last_successful_result {
+                            break result;
+                        }
+                        return Err(CreateTxError::CoinSelection(err));
+                    }
                 }
-
-                let Some(w) = optional_remaining.pop() else {
-                    break result;
-                };
-                required_for_attempt.push(w);
             }
         } else {
             coin_selection
--- a/src/wallet/mod.rs
+++ b/src/wallet/mod.rs
@@ -1467,16 +1467,37 @@ impl Wallet {
             }
         };
 
-        let coin_selection = coin_selection
-            .coin_select(
-                required_utxos,
-                optional_utxos,
-                fee_rate,
-                outgoing + fee_amount,
-                &drain_script,
-                rng,
-            )
-            .map_err(CreateTxError::CoinSelection)?;
+        // Retry coin selection to avoid dust/zero drain outputs (see issue #376). If the
+        // selection yields NoChange the loop promotes an optional UTXO to required and retries;
+        // it exits when optional_remaining is exhausted or a viable drain output is found.
+        let should_retry_for_dust_drain = params.recipients.is_empty()
+            && params.drain_to.is_some()
+            && (params.drain_wallet || !params.utxos.is_empty())
+            && !params.manually_selected_only;
+
+        let mut required_for_attempt = required_utxos;
+        let mut optional_remaining = optional_utxos;
+        let coin_selection = loop {
+            let result = coin_selection
+                .coin_select(
+                    required_for_attempt.clone(),
+                    optional_remaining.clone(),
+                    fee_rate,
+                    outgoing + fee_amount,
+                    &drain_script,
+                    rng,
+                )
+                .map_err(CreateTxError::CoinSelection)?;
+
+            if !should_retry_for_dust_drain || !matches!(&result.excess, Excess::NoChange { .. }) {
+                break result;
+            }
+
+            let Some(w) = optional_remaining.pop() else {
+                break result;
+            };
+            required_for_attempt.push(w);
+        };
 
         let excess = &coin_selection.excess;
         tx.input = coin_selection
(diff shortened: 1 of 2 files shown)


src/wallet/mod.rs
```diff
@@ -1467,16 +1467,40 @@ impl Wallet {
...
+                    fee_rate,
+                    outgoing + fee_amount,
+                    &drain_script,
+                    rng,
+                )
+                .map_err(CreateTxError::CoinSelection)?;
+
+            if !should_retry_for_dust_drain
+                || !matches!(&result.excess, Excess::NoChange { .. })
+                || optional_remaining.is_empty()
```
nikunjkumar05: The loop has both an outer || optional_remaining.is_empty() break condition and an inner let Some(w) = optional_remaining.pop() else { break result } guard. Since the outer check already breaks when the Vec is empty, the else { break result } branch of pop() is dead code. This is confusing — pick one:

`if !should_retry_for_dust_drain || !matches!(&result.excess, Excess::NoChange { .. }) {
    break result;
}`

alienx5499: Good catch!
Thanks!

PeteClubSeven:
Is there anything we can do to help this progress? We've encountered the same issue when trying to build an estimation API for bark exits and we're facing the choice of trying to implement our own coin selection to work around it so we can offer a stable API to devs. Would review and/or testing help?

ValuedMammal:
@PeteClubSeven Rather than looping on `.coin_select()` internally, which is guarded on a bespoke set of conditions, could you match on the error in your app logic and retry `Wallet::build_tx` after adding the required UTXO?

Are you concerned that automatically promoting an optional UTXO without adjusting the parameters could drive the feerate lower than originally intended?

In your test the target is so low that we know by construction the change is below dust, so if you expect a certain number of inputs in the transaction (say 2), can you add it manually using `add_utxo` as suggested in https://github.com/bitcoindevkit/bdk_wallet/issues/376#issuecomment-3959920155?

PeteClubSeven:
> @PeteClubSeven Rather than looping on `.coin_select()` internally, which is guarded on a bespoke set of conditions, could you match on the error in your app logic and retry `Wallet::build_tx` after adding the required UTXO?
> 
> Are you concerned that automatically promoting an optional UTXO without adjusting the parameters could drive the feerate lower than originally intended?
> 
> In your test the target is so low that we know by construction the change is below dust, so if you expect a certain number of inputs in the transaction (say 2), can you add it manually using `add_utxo` as suggested in [#376 (comment)](https://github.com/bitcoindevkit/bdk_wallet/issues/376#issuecomment-3959920155)?

We wanted to just let BDK handle UTXO selection and only request the foreign P2A be included. We were trying to avoid manual coin selection since we felt you guys have probably figured out a more optimal coin selection algorithm than we would try and hack together for this one part of the bark UX (unilateral exit).

One of the problems which prevents us from easily handling this as developers is that there are two classes of `InsufficientFunds` errors which can occur from this:
1. The wallet itself has ran out of funds, `available` will be the wallet balance and `needed` will be the total balance needed for the transaction, this can be escalated to the user as it is out of our control.
2. The drain output would be dust (this specific issue) and the `available` balance is actually `total_inputs - fee` (aka 220 sats) and `needed` is `ScriptBuf::minimal_non_dust` (330 sats for P2TR).

2 is a problem because the same error type is returned for what is clearly a very different class of error. If all we need to do is increase the fee by `needed - available` we can do that with a more descriptive error type.

Since I posted my last message we tried a new approach of wrapping the default coin selection algorithm with a little hack. It passes our integration tests so we think it works without impacting the currently working tests. Feel free to see what we did: https://gitlab.com/ark-bitcoin/bark/-/merge_requests/2273/diffs#03f2375aade6b4f63c8203ea3ab34ceef5a59134_153_159

Issue #239, opened 2025-05-22
Consider making all enum variants `#[non_exhaustive]`


evanlinjin:
**Describe the enhancement**  

This allows us to add error variants without breaking the API.

To Do: Figure out whether this is best practice and whether there are any downsides.

PR #449, opened 2026-04-19
 fix(wallet)!: make TxBuilder::current_height type-safe via absolute::Height

### LINKED ISSUE #49: Transaction builder should handle invalid tip height
"In the transaction builder, current_height will crash on an invalid absolute locktime height. If None was provided by the caller, this would let a remote chain source crash us by providing at invalid tip height. But that's considered out of scope. Although it seems unlikely that this parameter would be exposed to a third party, it's safer to only panic on inconsistent internal state and not on externally provided inputs."

"Out of scope since we trust our chain source but the transaction creation code will crash if the tip's height is not less than 500'000'000 (LockTime::from_height L1305), which could allow a remote Electrum or Esplora server to crash us."

none34829:
### Description

Closes #49.

`TxBuilder::current_height` and the fallback to `chain.tip().height()` in `create_tx` both called `absolute::LockTime::from_height(h).expect(...)`, which panics whenever `h >= 500_000_000`. The Wizardsardine audit flagged this because it lets a remote Electrum/Esplora server crash the wallet by reporting a tip height at or above that threshold — panics on externally-provided input are unsafe for a library.

This PR replaces the panics with a typed error:

- `TxParams.current_height` now stores a raw `u32` instead of an already-validated `absolute::LockTime`, so the builder setter no longer has to validate in a position where it can't return a `Result`.
- `Wallet::create_tx` performs the `u32 -> LockTime` conversion in one place, mapping the `ConversionError` to a new `CreateTxError::InvalidCurrentHeight(u32)` variant via `?`.
- Both the explicit-caller path (`TxBuilder::current_height`) and the implicit-fallback path (chain tip) now flow through the same validation and surface the same error.

### Notes to the reviewers

- The change to `TxParams.current_height` is crate-private (`pub(crate)`) so there's no external API break there. The `TxBuilder::current_height` public signature is unchanged.
- Adding a new variant to `CreateTxError` is technically an additive change, but it is an enum so downstream exhaustive matches would need to be updated. `CreateTxError` is not marked `#[non_exhaustive]`, which is tracked separately in #239.
- Follows the audit recommendation: "it's safer to only panic on inconsistent internal state and not on externally provided inputs."

### Changelog notice

- Fixed: `Wallet::create_tx` now returns `CreateTxError::InvalidCurrentHeight` instead of panicking when `TxBuilder::current_height` or the fallback chain tip height is `>= 500_000_000`.

#### All Submissions:

* [x] I ran `just p` before pushing (140 lib tests + 118 wallet + 22 fee_bump + 10 persisted + 13 descriptor_macro + 54 doctests all pass; clippy clean; fmt clean; cargo doc clean)

#### Bugfixes:

### DIFF
--- a/src/wallet/mod.rs
+++ b/src/wallet/mod.rs
@@ -1305,7 +1305,11 @@ impl Wallet {
             // If they didn't tell us the current height, we assume it's the latest sync height.
             None => {
                 let tip_height = self.chain.tip().height();
-                absolute::LockTime::from_height(tip_height).expect("invalid height")
+                // If the local chain tip height is not a valid block height for a locktime,
+                // that's an invariant violation in `LocalChain` (block heights are
+                // always below the locktime threshold) and should be addressed upstream.
+                absolute::LockTime::from_height(tip_height)
+                    .expect("LocalChain tip height should be a valid block height")
             }
             Some(h) => h,
         };
--- a/src/wallet/tx_builder.rs
+++ b/src/wallet/tx_builder.rs
@@ -649,9 +649,12 @@ impl<'a, Cs> TxBuilder<'a, Cs> {
     ///    them using [`TxBuilder::add_utxos`].
     ///
     /// In both cases, if you don't provide a current height, we use the last sync height.
-    pub fn current_height(&mut self, height: u32) -> &mut Self {
-        self.params.current_height =
-            Some(absolute::LockTime::from_height(height).expect("Invalid height"));
+    ///
+    /// This method takes an [`absolute::Height`], which is guaranteed by the type system to
+    /// be a valid block height for a locktime (below 500_000_000). Callers can construct one
+    /// with [`absolute::Height::from_consensus`].
+    pub fn current_height(&mut self, height: absolute::Height) -> &mut Self {
+        self.params.current_height = Some(absolute::LockTime::from(height));
         self
     }
 
--- a/tests/wallet.rs
+++ b/tests/wallet.rs
@@ -298,7 +298,7 @@ fn test_create_tx_custom_locktime() {
     let mut builder = wallet.build_tx();
     builder
         .add_recipient(addr.script_pubkey(), Amount::from_sat(25_000))
-        .current_height(630_001)
+        .current_height(absolute::Height::from_consensus(630_001).unwrap())
         .nlocktime(absolute::LockTime::from_height(630_000).unwrap());
     let psbt = builder.finish().unwrap();
 
@@ -2537,7 +2537,7 @@ fn test_spend_coinbase() {
     let mut builder = wallet.build_tx();
     builder
         .add_recipient(addr.script_pubkey(), balance.immature / 2)
-        .current_height(confirmation_height);
+        .current_height(absolute::Height::from_consensus(confirmation_height).unwrap());
     assert!(matches!(
         builder.finish(),
         Err(CreateTxError::CoinSelection(
@@ -2552,7 +2552,7 @@ fn test_spend_coinbase() {
     let mut builder = wallet.build_tx();
     builder
         .add_recipient(addr.script_pubkey(), balance.immature / 2)
-        .current_height(not_yet_mature_time);
+        .current_height(absolute::Height::from_consensus(not_yet_mature_time).unwrap());
     assert_matches!(
         builder.finish(),
         Err(CreateTxError::CoinSelection(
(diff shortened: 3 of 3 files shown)


src/wallet/tx_builder.rs
```diff
@@ -650,8 +650,7 @@ impl<'a, Cs> TxBuilder<'a, Cs> {
     ///
     /// In both cases, if you don't provide a current height, we use the last sync height.
     pub fn current_height(&mut self, height: u32) -> &mut Self {
-        self.params.current_height =
-            Some(absolute::LockTime::from_height(height).expect("Invalid height"));
+        self.params.current_height = Some(height);
```
ValuedMammal: You could remove the error case completely by making the function type safe, in fact I would prefer that way. Throwing CreateTxError is catching the problem too late, because if a `LockTime::from_height` can't be created from the wallet's chain tip, that indicates a problem with the LocalChain and should be addressed upstream.

```rust
pub fn current_height(&mut self, height: Height) -> &mut Self {
    self.params.current_height = Some(absolute::LockTime::from(height));
    self
}
```

none34829:
Thanks for the review- the type-safe approach is definitely cleaner. 

- `TxBuilder::current_height` now takes `absolute::Height` instead of `u32`, so the conversion to `LockTime` is infallible via `impl From<Height> for LockTime`. No `expect`, no new error variant.
- The chain tip fallback in `create_tx` keeps the `expect`, but now with a message noting that an invalid `LocalChain` tip height is an upstream invariant violation rather than silently swallowing it.
- `CreateTxError::InvalidCurrentHeight` removed entirely, along with the test that exercised it (the error case is gone).
- Existing test call sites (`test_create_tx_custom_locktime`, `test_spend_coinbase`) updated to construct the height via `absolute::Height::from_consensus(h).unwrap()`.

none34829:
CI failures here are unrelated to this PR- `bitcoin-units 0.1.3` was published and deprecated `FeeRate::from_sat_per_vb_unchecked`, which breaks `clippy -D warnings` on master too (47 errors locally when I reproduce against master).

I've opened #455 with a focused mechanical rename to `from_sat_per_vb_u32`. Once that lands I'll rebase this PR on top and CI should go green.

src/wallet/mod.rs
```diff
@@ -1305,7 +1305,11 @@ impl Wallet {
             // If they didn't tell us the current height, we assume it's the latest sync height.
             None => {
                 let tip_height = self.chain.tip().height();
-                absolute::LockTime::from_height(tip_height).expect("invalid height")
+                // If the local chain tip height is not a valid block height for a locktime,
+                // that's an invariant violation in `LocalChain` (block heights are
+                // always below the locktime threshold) and should be addressed upstream.
+                absolute::LockTime::from_height(tip_height)
+                    .expect("LocalChain tip height should be a valid block height")
             }
```
yan-pi, line 1313: The invariant described here is not enforced by `LocalChain`.

`CheckPoint::new` accepts a `BlockId` with any `u32` height, `CheckPoint::push` only checks ordering, and `LocalChain::from_changeset`/`apply_update` do not check the locktime threshold.

The update below succeeds, after which transaction creation reaches this `expect`:

```rust
let tip = wallet.latest_checkpoint().push(BlockId {
    height: 500_000_000,
    hash: BlockHash::all_zeros(),
}).unwrap();

wallet.apply_update(Update {
    chain: Some(tip),
    ..Default::default()
}).unwrap();

builder.add_recipient(script, amount);
let _ = builder.finish();
```

I reproduced the same panic after taking the staged `ChangeSet` and loading it with `Wallet::load().load_wallet_no_persist(...)`.

I'd keep `absolute::Height` for the explicit setter and make the fallback conversion return `CreateTxError::InvalidCurrentHeight` instead of relying on this invariant.

Please also add regression tests for the update and reload paths. The current tests only wrap existing valid constants in `Height::from_consensus(...).unwrap()` and do not exercise the audit finding.

yan-pi (commented):
NACK (6dac1a0) on the current patch as a resolution for #49. 

Accepting `absolute::Height` in the explicit setter is a good API improvement, but the fallback still panics for a height accepted through `Wallet::apply_update` and `Wallet::load`. I reproduced both paths at height `500_000_000` on the current head.

I think you can keep the typed setter, make the fallback conversion return `CreateTxError::InvalidCurrentHeight`, and add the regression test. 

The PR description and checklist also need to be updated: they describe the old force-pushed implementation, the current signature is API-breaking, and the branch needs a rebase.

PR #453, opened 2026-04-23
fix(descriptor): correct index tracking in `combinations()` function

tnull:
### Description

The combinations function was pushing `new_index` (the enumerate index relative to the skipped iterator) instead of `index + 1 + new_index` (the absolute index in the original vec). This caused duplicate and incorrect combinations to be generated for inputs larger than trivial sizes.

#### All Submissions:

#### Bugfixes:

### DIFF
--- a/src/descriptor/policy.rs
+++ b/src/descriptor/policy.rs
@@ -196,7 +196,7 @@ fn combinations(vec: &[usize], size: usize) -> Vec<Vec<usize>> {
             for (new_index, val) in vec.iter().skip(index + 1).enumerate() {
                 let mut cloned = vals.clone();
                 cloned.push(*val);
-                queue.push_front((new_index, cloned));
+                queue.push_front((index + 1 + new_index, cloned));
             }
         }
     }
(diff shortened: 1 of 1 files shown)


ValuedMammal:
Rebased

ValuedMammal (approved):
ACK d6bbe77a1fbf12fb28d23e4d53d0bd3f1691a2a6; ran the tests locally

ValuedMammal:
It's an obscure bug and the algorithm is not immediately clear, but the test confirms the bug.

ValuedMammal:
The key to understanding it is the meaning of the pair in the queue: (last_picked_index, partial_combo). The last_picked_index is the absolute position in vec of the most recently appended element. When we want to grow a partial combination, we only look at elements strictly to the right of that position — that one rule is what prevents duplicates and repetition.

Let's trace `combinations(&[0, 1, 2, 3], 3)` (4 choose 3) to make it concrete.

```sh
[0] ─── [0,1] ─── [0,1,2] ✓
    │          └── [0,1,3] ✓
    ├── [0,2] ─── [0,2,3] ✓
    └── [0,3]              (dead end — nothing to the right)
[1] ─── [1,2] ─── [1,2,3] ✓
    └── [1,3]              (dead end)
[2] ─── [2,3]              (dead end — only 2 elements, can't reach size 3)
[3]                        (dead end)
```

push_front makes it go deep before going wide — it finishes the [0,…] subtree entirely before visiting [1,…]. The index threaded through the queue is the "don't look left of here" cursor, and index + 1 + new_index converts the loop's relative offset back to an absolute position. That's the line the bug was on — the original code used new_index bare, which accidentally restarted the cursor from near the beginning of the slice on each recursive expansion.

(Claude 4.6)

PR #456, opened 2026-04-24
fix(wallet): enforce OP_RETURN standardness via try_add_data

### LINKED ISSUE #49: Enforce standardness rules for TxBuilder::add_data

tharu-jwd:
`add_data` has been silently accepting oversized payloads and multiple OP_RETURNs, both non-standard. Closes bitcoindevkit/bdk-tx#49.

Adds `try_add_data` that validates at call-site and deprecates `add_data`:

- Data payload capped at 80 bytes (Bitcoin Core's `MAX_OP_RETURN_RELAY` = 83 bytes total scriptPubKey)
- Errors if an OP_RETURN recipient is already present

Two new `CreateTxError` variants: `OpReturnInvalidDataSize(usize)` and `MultipleOpReturnOutputs`.

Deprecating rather than removing since this is a public API.

---

#### All Submissions:

#### New Features:

### DIFF
--- a/src/wallet/error.rs
+++ b/src/wallet/error.rs
@@ -215,6 +215,17 @@ pub enum CreateTxError {
     MissingNonWitnessUtxo(OutPoint),
     /// Miniscript PSBT error
     MiniscriptPsbt(MiniscriptPsbtError),
+    /// OP_RETURN data payload exceeds the 80-byte standardness limit
+    ///
+    /// Bitcoin Core enforces a maximum `scriptPubKey` size of 83 bytes for data carrier outputs
+    /// (`MAX_OP_RETURN_RELAY`), which constrains the data payload to at most 80 bytes.
+    /// See <https://github.com/bitcoin/bitcoin/blob/master/src/policy/policy.h>
+    OpReturnInvalidDataSize(usize),
+    /// Transaction already contains an OP_RETURN output
+    ///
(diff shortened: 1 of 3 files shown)


notmandatory:
This is meant to be fixed for the bdk-tx repo since that's the crate that new transaction building features will be developed for. The team's current plan is to only fix critical bugs for the existing bdk_wallet TxBuilder.

PR #457, opened 2026-04-24
fix(descriptor): reject bare descriptors in check_wallet_descriptor

### LINKED ISSUE #54: Return error for invalid descriptor types
"The library will crash on descriptors without an address form, as it assumes it can always get it but doesn't check it. For instance importing a raw() descriptor would make BDK crash internally. This can be an issue for a server providing an API to register descriptors and monitor them using BDK."

"The descriptor is not checked in Wallet::create_with_params (it only is in load_with_params), allowing to create a Wallet with an incompatible (multipath, hardened steps, ..) descriptor which would break a number of invariants during usage of the Wallet."

tharu-jwd:
Bare descriptors (e.g. `pk()`) have no standard address form. Passing one to the wallet wasn't caught at creation time, causing panics later when address derivation is attempted. Closes #54.

Adds `UnsupportedDescriptorType` to `DescriptorError` and checks for `DescriptorType::Bare` in `check_wallet_descriptor`, returning the new error before the descriptor reaches wallet internals.

---

#### All Submissions:

#### Bugfixes:

### DIFF
--- a/src/descriptor/mod.rs
+++ b/src/descriptor/mod.rs
@@ -321,6 +321,12 @@ pub(crate) fn check_wallet_descriptor(
         ));
     }
 
+    // Reject bare descriptors: they have no standard address form and would cause panics
+    // inside the wallet when address derivation is attempted.
+    if descriptor.desc_type() == DescriptorType::Bare {
+        return Err(DescriptorError::UnsupportedDescriptorType);
+    }
+
     // Run miniscript's sanity check, which will look for duplicated keys and other potential
     // issues.
     descriptor.sanity_check()?;
@@ -912,6 +918,16 @@ mod test {
         let result = check_wallet_descriptor(&descriptor);
 
         assert!(result.is_err());
+
+        // Bare descriptors (e.g. pk()) have no standard address form and must be rejected.
+        let descriptor = Descriptor::<DescriptorPublicKey>::from_str(
+            "pk(tpubD6NzVbkrYhZ4XHndKkuB8FifXm8r5FQHwrN6oZuWCz13qb93rtgKvD4PQsqC4HP4yhV3tA2fqr2RbY5mNXfM7RxXUoeABoDtsFUq2zJq6YK/0/*)",
+        )
+        .expect("must parse");
+        assert_matches!(
+            check_wallet_descriptor(&descriptor),
+            Err(DescriptorError::UnsupportedDescriptorType)
+        );
     }
 
     #[test]
--- a/src/descriptor/mod.rs
+++ b/src/descriptor/mod.rs
@@ -321,10 +321,20 @@ pub(crate) fn check_wallet_descriptor(
         ));
     }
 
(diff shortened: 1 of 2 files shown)


ValuedMammal:
Would you mind dropping 33fb97b7 and push this again with just the bare descriptor check fix, thank you.

tharu-jwd:
> Would you mind dropping [33fb97b](https://github.com/bitcoindevkit/bdk_wallet/commit/33fb97b7a086fe1ba9e8dad4b04a88db70da9a3b) and push this again with just the bare descriptor check fix, thank you.

Hey, sure.

ValuedMammal:
Approach ACK

ACK 15604e8c11e24309b952c640ffa380b8a2c5eb5f

src/descriptor/mod.rs
```diff
@@ -912,6 +918,16 @@ mod test {
         let result = check_wallet_descriptor(&descriptor);
 
         assert!(result.is_err());
+
+        // Bare descriptors (e.g. pk()) have no standard address form and must be rejected.
+        let descriptor = Descriptor::<DescriptorPublicKey>::from_str(
+            "pk(tpubD6NzVbkrYhZ4XHndKkuB8FifXm8r5FQHwrN6oZuWCz13qb93rtgKvD4PQsqC4HP4yhV3tA2fqr2RbY5mNXfM7RxXUoeABoDtsFUq2zJq6YK/0/*)",
```
sdmg15, line 934: I think it'll be also worth fixing the case for 

`let descriptor = "raw(deadbeef)";` 

I wrote a while back the reproducing of this audit report: https://gist.github.com/sdmg15/54ee406e3826ba4acc51312e30054df2

ItoroD: Since DescriptorType::Raw does not exist as one of the types in DescriptorType, it might be best to match desc.type on all supported types and then return Err is its not part of any of those.

ValuedMammal: > I wrote a while back the reproducing of this audit report

This is reproducing an error which is expected, and technically not a panic.

ValuedMammal:
@tharu-jwd I couldn't get your tests to pass.

src/descriptor/mod.rs
```diff
@@ -912,6 +928,24 @@ mod test {
...
+        )
+        .expect("must parse");
+        assert_matches!(
+            check_wallet_descriptor(&descriptor),
+            Err(DescriptorError::UnsupportedDescriptorType)
+        );
+
+        // Raw script descriptors have no standard address form and must be rejected.
+        let descriptor = Descriptor::<DescriptorPublicKey>::from_str("raw(deadbeef)")
```
ItoroD, line 943: Test is failing because of this. You cannot parse raw, (as @ValuedMammal mentioned) so this error is expected. Your pk works because thats a bare descriptor. Might want to take this out.

I think fmt issue as well.

PR #461, opened 2026-04-24
fix: always validate non_witness_utxo in add_foreign_utxo

tnull:
### Description
Previously, non_witness_utxo was only validated when witness_utxo was None. An attacker could supply both a valid witness_utxo and a non_witness_utxo with the wrong txid, bypassing the txid/outpoint checks entirely.

Restructure the validation so non_witness_utxo is always checked when present, regardless of whether witness_utxo is also set. Only require at least one of the two to exist.

#### All Submissions:

#### Bugfixes:

### DIFF
--- a/src/wallet/tx_builder.rs
+++ b/src/wallet/tx_builder.rs
@@ -407,23 +407,19 @@ impl<'a, Cs> TxBuilder<'a, Cs> {
         satisfaction_weight: Weight,
         sequence: Sequence,
     ) -> Result<&mut Self, AddForeignUtxoError> {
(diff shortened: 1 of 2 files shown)


ValuedMammal (approved):
ACK f0302e9760b63565c81aaae86af407f1d9a7074a

luisschwab (approved):
ACK f0302e9760b63565c81aaae86af407f1d9a7074a

oleonardolima (commented):
utACK f0302e9760b63565c81aaae86af407f1d9a7074a

PR #471, opened 2026-04-29
fix(audit): add non-witness-utxo validation for psbt

### LINKED ISSUE #468: [psbt] Validate non_witness_utxo txid with tx input previous output
`get_utxo_for()` uses `non_witness_utxo` without verifying its txid matches the input's `previous_output.txid`. An attacker can substitute a fake transaction with inflated values, causing incorrect `fee_amount()` and `fee_rate()` results.

https://github.com/bitcoindevkit/bdk_wallet/blob/db1eab42db6c75c1d15641825511115c7dbbc248/src/psbt/mod.rs#L41-L50

muhahahmad68:
### Description

This PR fixes a security vulnerability in `get_utxo_for()` within src/psbt/mod.rs where `non_witness_utxo` was being used without verifying that its txid matches the input's `previous_output.txid`.

### Notes to the reviewers

* The validation silently returns None on txid mismatch rather than propagating an error, which is consistent with the existing behavior of this function.

#### Fixed
- [psbt] Validate `non_witness_utxo` txid against `previous_output.txid` in
  `get_utxo_for()` to prevent fee calculation manipulation via substituted
  transactions (fixes #468)

#### All Submissions:

#### New Features:

* N/A

#### Bugfixes:

### DIFF
--- a/src/psbt/mod.rs
+++ b/src/psbt/mod.rs
@@ -43,8 +43,12 @@ impl PsbtUtils for Psbt {
 
         match (&input.witness_utxo, &input.non_witness_utxo) {
             (Some(_), _) => input.witness_utxo.clone(),
-            (_, Some(_)) => input.non_witness_utxo.as_ref().map(|in_tx| {
-                in_tx.output[tx.input[input_index].previous_output.vout as usize].clone()
+            (_, Some(_)) => input.non_witness_utxo.as_ref().and_then(|prev_tx| {
+                let outpoint = tx.input[input_index].previous_output;
+                if prev_tx.compute_txid() != outpoint.txid {
+                    return None;
+                }
+                prev_tx.output.get(outpoint.vout as usize).cloned()
             }),
             _ => None,
         }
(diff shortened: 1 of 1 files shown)


src/psbt/mod.rs
```diff
@@ -43,8 +43,15 @@ impl PsbtUtils for Psbt {
...
-                in_tx.output[tx.input[input_index].previous_output.vout as usize].clone()
+            (_, Some(_)) => input.non_witness_utxo.as_ref().and_then(|in_tx| {
+                // Validate that the non_witness_utxo txid matches the input's previous output txid
+                if in_tx.compute_txid() != tx.input[input_index].previous_output.txid {
+                    return None;
+                }
+                in_tx
+                    .output
+                    .get(tx.input[input_index].previous_output.vout as usize)
+                    .cloned()
```
ValuedMammal: small nit for readability
```suggestion
            (_, Some(_)) => input.non_witness_utxo.as_ref().and_then(|prev_tx| {
                let outpoint = tx.input[input_index].previous_output;
                if prev_tx.compute_txid() != outpoint.txid {
                    return None;
                }
                prev_tx.output.get(outpoint.vout as usize).cloned()
            }),
```

ValuedMammal (approved):
ACK 6b2f8ffb785c4a7c862fcc2be68b97af74c473ee

It'd be nice to see a test to check

1. `get_utxo_for` returns None on txid mismatch
2. `get_utxo_for` returns None on previous_output.vout out of bounds

commit should say

`fix(psbt): add non-witness-utxo validation for get_utxo_for`

ValuedMammal:
We need to drop the duplicate 081c3be3 commit from this branch that already exists on master with a different ID.

ValuedMammal:
ACK I would just suggest squashing the changes into a single commit keeping the `fix(psbt) ...` title. 🙏 

ACK 217dd85509af0b00bfd957f932eefc7fc8c274f7

muhahahmad68:
> ACK I would just suggest squashing the changes into a single commit keeping the `fix(psbt) ...` title. 🙏

gotcha!

PR #476, opened 2026-05-03
fix(signer): respect allow_all_sighashes in SignerWrapper::sign_input

### LINKED ISSUE #469: `sign_input` should respect `allow_all_sighashes` sign option
## Description

`SignOptions::allow_all_sighashes` is documented as controlling whether the signer will accept non-`SIGHASH_ALL` sighash types. However, this flag is only checked inside [`Wallet::sign`](https://github.com/bitcoindevkit/bdk_wallet/blob/db1eab42db6c75c1d15641825511115c7dbbc248/src/wallet/mod.rs#L1775) as a PSBT-wide pre-flight guard. The underlying `InputSigner::sign_input` never consults it although it's a public trait method. Callers that use `InputSigner` or `TransactionSigner` directly bypass the check entirely.

### Affected code

- `SignerWrapper<PrivateKey>::sign_input` — the leaf implementation for all BDK software signers.  
  `DescriptorXKey` and `DescriptorMultiXKey` both derive a key and then delegate here, so all three are affected through the same code path.
- All signing contexts (Legacy, Segwitv0, Taproot) are affected.

The ECDSA path calls `psbt.sighash_ecdsa()` which returns whatever type the PSBT declares and proceeds. The Taproot path calls `compute_tap_sighash()` which similarly reads the PSBT sighash type without checking `allow_all_sighashes`.

### Steps to reproduce

1. Obtain a signer directly (e.g. from wallet internals or a custom `SignersContainer`)
2. Set SIGHASH_NONE on a PSBT input
3. Call `signer.sign_input(&mut psbt, 0, &SignOptions::default(), &secp)` (signs successfully despite `allow_all_sighashes` being false)

### Expected behavior

`sign_input` should return `Err(SignerError::NonStandardSighash)` when `sign_options.allow_all_sighashes` is `false` and the PSBT input declares a sighash type other than `SIGHASH_ALL` / `SIGHASH_DEFAULT`.

## Proposed fix

Add the sighash guard in `SignerWrapper<PrivateKey>::sign_input`, after the "already finalized" early return but before the `match self.ctx` dispatch. This ensures the check covers all signing contexts and all delegating implementations in a single place:

```rust
if !sign_options.allow_all_sighashes {
    let sighash_type = psbt.inputs[input_index].sighash_type;
    if sighash_type.is_some()
        && sighash_type != Some(EcdsaSighashType::All.into())
        && sighash_type != Some(TapSighashType::All.into())
        && sighash_type != Some(TapSighashType::Default.into())
    {
        return Err(SignerError::NonStandardSighash);
    }
}
```

This mirrors the existing logic in `Wallet::sign`. Through the `Wallet::sign` path, behavior is unchanged since the PSBT-wide check still fires first.

A test should be added covering:

- Taproot input with `SIGHASH_NONE` and `allow_all_sighashes: false` returns `NonStandardSighash`
- The same with `allow_all_sighashes: true` proceeds normally.

muhahahmad68:
### Description
`SignOptions::allow_all_sighashes` is documented as controlling whether the signer will accept non-`SIGHASH_ALL`    sighash types. However, the check only existed inside `Wallet::sign` as a PSBT-wide pre-flight guard. The underlying `SignerWrapper<PrivateKey>::sign_input` never consulted it, meaning callers using `InputSigner` or `TransactionSigner` directly could bypass the check entirely.

Added a sighash guard in `SignerWrapper<PrivateKey>::sign_input` before the `match self.ctx` dispatch. This ensures the check covers all signing contexts (Legacy, Segwitv0, Taproot) and all delegating implementations in a single place.

### Tests Added

- Taproot input with `SIGHASH_NONE` and `allow_all_sighashes: false` returns `NonStandardSighash`
- Same setup with `allow_all_sighashes: true` passes the guard successfully

Fixes #469 

#### All Submissions:

#### New Features:

* N/A

#### Bugfixes:

### DIFF
--- a/src/wallet/signer.rs
+++ b/src/wallet/signer.rs
@@ -91,6 +91,7 @@ use core::ops::{Bound::Included, Deref};
 
 use bitcoin::bip32::{ChildNumber, DerivationPath, Fingerprint, Xpriv};
 use bitcoin::hashes::hash160;
+use bitcoin::psbt::PsbtSighashType;
 use bitcoin::secp256k1::Message;
 use bitcoin::sighash::{EcdsaSighashType, TapSighash, TapSighashType};
 use bitcoin::{ecdsa, psbt, sighash, taproot};
@@ -154,10 +155,10 @@ pub enum SignerError {
     MissingWitnessScript,
     /// The fingerprint and derivation path are missing from the psbt input
     MissingHdKeypath,
-    /// The psbt contains a non-`SIGHASH_ALL` sighash in one of its input and the user hasn't
+    /// The PSBT contains a non-`SIGHASH_ALL` sighash in one of its inputs and the user hasn't
     /// explicitly allowed them
     ///
-    /// To enable signing transactions with non-standard sighashes set
+    /// To enable signing transactions with non-`SIGHASH_ALL` set
     /// [`SignOptions::allow_all_sighashes`] to `true`.
     NonStandardSighash,
     /// Invalid SIGHASH for the signing context in use
@@ -462,6 +463,28 @@ impl InputSigner for SignerWrapper<PrivateKey> {
             || psbt.inputs[input_index].final_script_witness.is_some()
         {
             return Ok(());
+        }
+
+        fn is_sighash_all(ctx: SignerContext, sighash_type: PsbtSighashType) -> bool {
+            match ctx {
+                SignerContext::Legacy | SignerContext::Segwitv0 => {
+                    matches!(sighash_type.ecdsa_hash_ty(), Ok(EcdsaSighashType::All))
+                }
+                SignerContext::Tap { .. } => {
+                    matches!(
+                        sighash_type.taproot_hash_ty(),
+                        Ok(TapSighashType::All) | Ok(TapSighashType::Default)
+                    )
+                }
+            }
+        }
+
+        if !sign_options.allow_all_sighashes {
+            if let Some(sht) = psbt.inputs[input_index].sighash_type {
+                if !is_sighash_all(self.ctx, sht) {
+                    return Err(SignerError::NonStandardSighash);
+                }
+            }
         }
 
         let pubkey = PublicKey::from_private_key(secp, self);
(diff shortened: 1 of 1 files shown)


src/wallet/signer.rs
```diff
@@ -1010,6 +1021,99 @@ mod signers_container_tests {
...
+            ..Default::default()
+        };
+        assert_eq!(
+            signer.sign_input(&mut psbt, 0, &opts_reject, &secp),
+            Err(SignerError::NonStandardSighash),
+            "expected NonStandardSighash when allow_all_sighashes is false"
+        );
+
+        psbt.inputs[0].sighash_type = Some(TapSighashType::All.into());
+        assert_ne!(
```
benthecarman: instead of these `assert_ne!`s it'd be more readable if you're asserting if it is `Ok(())`

here and below

src/wallet/signer.rs
```diff
@@ -1010,6 +1021,99 @@ mod signers_container_tests {
...
+            signer.sign_input(&mut psbt, 0, &opts_reject, &secp),
+            Err(SignerError::NonStandardSighash),
+            "SIGHASH_ALL should not be rejected when allow_all_sighashes is false"
+        );
+
+        psbt.inputs[0].sighash_type = Some(TapSighashType::Default.into());
+        assert_ne!(
+            signer.sign_input(&mut psbt, 0, &opts_reject, &secp),
```
benthecarman: after you've had one successful sig, it'll always succeed because it is already signed. you also need to clear out the signatures before each. Imo would be cleanest to start with `base_psbt` and for each assertion clone it and make modifications as you need, that way you don't have any risk of secondary effects

benthecarman (changes requested):

muhahahmad68:
Makes more sense, I'll effect the changes as suggested. Much thanks @benthecarman

aagbotemi (commented):
Well done working on this @muhahahmad68. I left some reviews for you.

src/wallet/signer.rs
```diff
@@ -464,6 +464,17 @@ impl InputSigner for SignerWrapper<PrivateKey> {
...
+            let sighash_type = psbt.inputs[input_index].sighash_type;
+            if sighash_type.is_some()
+                && sighash_type != Some(EcdsaSighashType::All.into())
+                && sighash_type != Some(TapSighashType::All.into())
+                && sighash_type != Some(TapSighashType::Default.into())
+            {
+                return Err(SignerError::NonStandardSighash);
+            }
+        }
+
```
aagbotemi, line 489: The guard sits above `match self.ctx` but isn't context-aware. This causes two issues, `EcdsaSighashType::All.into()` and `TapSighashType::All.into()` produces the same `PsbtSighashType`, so one of the `!=` arms is dead, and a Legacy/Segwitv0 input with `sighash_type = Some(0x00)` passes this guard and gets rejected by `psbt.sighash_ecdsa()` as `SignerError::Psbt()`, not the documented `NonStandardSighash`.

A small context-aware helper handles both:
```rs
fn is_standard_sighash(ctx: SignerContext, sht: psbt::PsbtSighashType) -> bool {
    match ctx {
        SignerContext::Tap { .. } => matches!(
            sht.taproot_hash_ty(),
            Ok(TapSighashType::All) | Ok(TapSighashType::Default)
        ),
        SignerContext::Legacy | SignerContext::Segwitv0 => {
            matches!(sht.ecdsa_hash_ty(), Ok(EcdsaSighashType::All))
        }
    }
}

if !sign_options.allow_all_sighashes {
    if let Some(sht) = psbt.inputs[input_index].sighash_type {
        if !is_standard_sighash(self.ctx, sht) {
            return Err(SignerError::NonStandardSighash);
        }
    }
}
```

The dead comparison also disappears.

src/wallet/signer.rs
```diff
@@ -134,7 +134,7 @@ impl From<Fingerprint> for SignerId {
 }
 
 /// Signing error
-#[derive(Debug)]
+#[derive(Debug, PartialEq)]
```
aagbotemi: I think the `PartialEq` derived on `SignerError` was added only so the test could use `assert_eq!`, same as `MiniscriptPsbtError`. Switching the test to `assert_matches!` let you revert the derives.

muhahahmad68: Good catch. will update

src/wallet/signer.rs
muhahahmad68: well gotten!

ValuedMammal: Maybe more correct to call the function `is_sighash_all`. The docs are a bit confusing with regard to "standard" sighash types, but `allow_all_sighashes` clearly says that when `false` we only allow signing using `SIGHASH_ALL`.

ValuedMammal: Seems that without the fix the same non context-aware validation exists in `wallet.sign()` which is annoying because the ineffective guard now relies on the correct logic deeper down.

The `SignerError` variants are messy when it comes to invalid sighash types, but since we're sticking to `NonStandardSighash` for this particular case, maybe update the docs like this?

```rs
/// The PSBT contains a non-`SIGHASH_ALL` sighash type in one of its inputs and the user hasn't
/// explicitly allowed them.
///
/// To enable signing transactions with non-`SIGHASH_ALL` set
/// [`SignOptions::allow_all_sighashes`] to `true`.
NonStandardSighash,
```

src/wallet/signer.rs
```diff
@@ -464,6 +465,28 @@ impl InputSigner for SignerWrapper<PrivateKey> {
             return Ok(());
         }
 
+        fn is_sighash_all(ctx: SignerContext, sht: PsbtSighashType) -> bool {
```
ValuedMammal:
```suggestion
        fn is_sighash_all(ctx: SignerContext, sighash_type: PsbtSighashType) -> bool {
```

src/wallet/signer.rs
muhahahmad68: Agreed, "non-standard" sounds ambiguous.

muhahahmad68: I think the non context-aware validation for `Wallet::sign()` needs should be addressed in a separate pr?

ValuedMammal (approved):
ACK 9e45c56246f1dacb0e3ffdea293eb8cdf85fe5e3; ran the tests locally

Issue #477, opened 2026-05-04
feat:  add BIP-431 (TRUC) support


oleonardolima:
**Describe the enhancement**  
As reported in #419 we don't have proper support for BIP-431 (TRUC) transactions, which have different specification/policy for what outputs/nVersion can be used/mixed together, see [BIP-431](https://github.com/bitcoin/bips/blob/master/bip-0431.mediawiki#user-content-Specification).

As described in-depth on [BIP-431](https://github.com/bitcoin/bips/blob/master/bip-0431.mediawiki#user-content-Specification), and mentioned [here](https://github.com/bitcoindevkit/bdk_wallet/pull/442#pullrequestreview-4197234781) we need to support the following specification:

- [ ] Rule 1: A TRUC transaction signals replaceability, even if it does not signal BIP125 replaceability.
- [ ] Rule 2: Any TRUC transaction's unconfirmed ancestors must all be TRUC. Any descendant of an unconfirmed TRUC transaction must also be TRUC. #442
  > _NOTE: A TRUC transaction can spend outputs from confirmed non-TRUC transactions. A non-TRUC transaction can spend outputs from confirmed TRUC transactions._
- [ ] Rule 3: An unconfirmed TRUC transaction cannot have more than 1 unconfirmed ancestor. An unconfirmed TRUC transaction cannot have more than 1 unconfirmed descendant. CPFP Carve Out is not granted to TRUC transactions.
- [ ] Rule 4: A TRUC transaction cannot have a sigop-adjusted virtual size larger than 10,000 vB. #484
- [ ] Rule 5: A TRUC transaction that has an unconfirmed TRUC ancestor cannot have a sigop-adjusted virtual size larger than 1000 vB. #485
- [x] ~~Rule 6: An individual TRUC transaction is permitted to be below the mempool min relay feerate, assuming it is considered within a package that meets the mempool's feerate requirements.~~ **it's not applicable, lmk if i'm mistaken**.

**Use case**  

It allows users to properly build TRUC transactions, without the need for any workaround as mentioned here: https://github.com/bitcoindevkit/bdk_wallet/issues/419#issuecomment-4158215640 (specially L2s, such as: Lightning, Ark, ...)

**Impact**
- [x] Blocking production usage
- [ ] Nice-to-have / UX improvement
- [x] Developer experience / maintainability

**Are you using BDK in a production project?**
- [x] Yes
- [ ] No
- [ ] Not yet, but planning to

[...]

PR #478, opened 2026-05-04
fix(wallet): add support for bip-431 rule 2

oleonardolima:
partially addresses #477
supersedes #442

### Description

Adds support for BIP-431 (TRUC) rule 2 in filter_utxos. When building a transaction, unconfirmed UTXOs are filtered based on version compatibility with the transaction being constructed:

* If building a TRUC (v3) transaction, only unconfirmed outputs from TRUC transactions are eligible as optional UTXOs.
* If building a non-TRUC transaction, only unconfirmed outputs from non-TRUC transactions are eligible as optional UTXOs.
* Confirmed outputs may be spent by either TRUC or non-TRUC transactions regardless of the source transaction's version.

The filter is implemented inline in filter_utxos, which now takes a `version` parameter passed down from the transaction builder. The version compatibility check uses an equality comparison on the `== Version(3)` predicate, ensuring the unconfirmed output's transaction version matches the version class (TRUC or non-TRUC) of the transaction being built.

Note that manually selected UTXOs bypass this filter — callers are responsible for ensuring any manually selected unconfirmed UTXO is compatible with the transaction version being built.

### Changelog notice

```
### Fixed

- fix(wallet): filter unconfirmed UTXOs by BIP-431 (TRUC) rule-2
```

#### All Submissions:

#### New Features:

#### Bugfixes:

### DIFF
--- a/src/wallet/mod.rs
+++ b/src/wallet/mod.rs
@@ -41,8 +41,9 @@ use bitcoin::{
     psbt,
     secp256k1::Secp256k1,
     sighash::{EcdsaSighashType, TapSighashType},
-    transaction, Address, Amount, Block, FeeRate, Network, NetworkKind, OutPoint, Psbt, ScriptBuf,
-    Sequence, SignedAmount, Transaction, TxOut, Txid, Weight, Witness,
+    transaction::{self, Version},
+    Address, Amount, Block, FeeRate, Network, NetworkKind, OutPoint, Psbt, ScriptBuf, Sequence,
+    SignedAmount, Transaction, TxOut, Txid, Weight, Witness,
 };
 use miniscript::{
     descriptor::KeyMap,
@@ -1429,7 +1430,7 @@ impl Wallet {
         let (required_utxos, optional_utxos) = {
             // NOTE: manual selection overrides unspendable
             let mut required: Vec<WeightedUtxo> = params.utxos.clone();
-            let optional = self.filter_utxos(&params, current_height.to_consensus_u32());
+            let optional = self.filter_utxos(&params, current_height.to_consensus_u32(), version);
 
             // If `drain_wallet` is true, all UTxOs are required.
             if params.drain_wallet {
@@ -1978,7 +1979,12 @@ impl Wallet {
 
     /// Given the options returns the list of utxos that must be used to form the
     /// transaction and any further that may be used if needed.
-    fn filter_utxos(&self, params: &TxParams, current_height: u32) -> Vec<WeightedUtxo> {
+    fn filter_utxos(
+        &self,
+        params: &TxParams,
+        current_height: u32,
+        version: Version,
+    ) -> Vec<WeightedUtxo> {
         if params.manually_selected_only {
             vec![]
         // Only process optional UTxOs if manually_selected_only is false.
@@ -1988,6 +1994,7 @@ impl Wallet {
                 .iter()
                 .map(|wutxo| wutxo.utxo.outpoint())
                 .collect::<HashSet<OutPoint>>();
+
             self.tx_graph
                 .graph()
                 // Get all unspent UTxOs from wallet.
@@ -2006,6 +2013,29 @@ impl Wallet {
                     full_txo
                         .is_mature(current_height)
                         .then(|| new_local_utxo(k, i, full_txo))
+                })
+                // only add to optional UTXOs those that follows BIP-431 (TRUC) specification.
+                // see https://github.com/bitcoin/bips/blob/master/bip-0431.mediawiki#specification
+                .filter(|local_output| {
+                    local_output.chain_position.is_confirmed()
+                        || match self.tx_graph().get_tx(local_output.outpoint.txid) {
+                            Some(ancestor_tx) => match is_truc(version) {
+                                // if building TRUC; filter out all unconfirmed non-TRUC.
+                                true => {
+                                    is_truc(ancestor_tx.version)
+                                        && local_output.chain_position.is_unconfirmed()
+                                }
+                                // if building non-TRUC; filter out all unconfirmed TRUC.
+                                false => {
+                                    !is_truc(ancestor_tx.version)
+                                        && local_output.chain_position.is_unconfirmed()
+                                }
+                            },
+                            // if we don't have the full tx available we can't assure the ancestor
+                            // tx version it assumes it's a valid
+                            // candidate.
+                            None => true,
+                        }
                 })
                 // only process UTXOs not selected manually, they will be considered later in the
                 // chain
@@ -2885,6 +2915,12 @@ fn make_indexed_graph(
     Ok(indexed_graph)
 }
 
+/// Check if the given [`transaction::Version`] is TRUC (Topologically Restricted Until
+/// Confirmation)s.
+fn is_truc(version: transaction::Version) -> bool {
+    version.eq(&Version(3))
+}
+
 /// Transforms a [`FeeRate`] to `f64` with unit as sat/vb.
 #[macro_export]
 #[doc(hidden)]
@@ -2981,8 +3017,13 @@ mod test {
         let mut builder = wallet.build_tx();
         builder.add_utxo(outpoint).expect("should add local utxo");
         let params = builder.params.clone();
+        let version = params.version.unwrap_or(Version::TWO);
         // enforce selection of first output in transaction
-        let received = wallet.filter_utxos(&params, wallet.latest_checkpoint().block_id().height);
+        let received = wallet.filter_utxos(
+            &params,
+            wallet.latest_checkpoint().block_id().height,
+            version,
+        );
         // Notice expected doesn't include the first output from two_output_tx as it should be
         // filtered out.
         let expected = vec![wallet
--- a/src/wallet/mod.rs
+++ b/src/wallet/mod.rs
@@ -41,8 +41,9 @@ use bitcoin::{
     psbt,
     secp256k1::Secp256k1,
     sighash::{EcdsaSighashType, TapSighashType},
-    transaction, Address, Amount, Block, FeeRate, Network, NetworkKind, OutPoint, Psbt, ScriptBuf,
-    Sequence, SignedAmount, Transaction, TxOut, Txid, Weight, Witness,
+    transaction::{self, Version},
+    Address, Amount, Block, FeeRate, Network, NetworkKind, OutPoint, Psbt, ScriptBuf, Sequence,
+    SignedAmount, Transaction, TxOut, Txid, Weight, Witness,
 };
 use miniscript::{
     descriptor::KeyMap,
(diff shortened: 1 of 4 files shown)


oleonardolima:
cc @yan-pi as he already reviewed the other PR and has a reproducible test using bdk-cli too.

oleonardolima:
I noticed in CI I need to fix some bdk_testenv/electrs/esplora related issues. However, the fix in bc05f88 is ready for review/discussion.

src/wallet/mod.rs
```diff
@@ -2885,6 +2915,12 @@ fn make_indexed_graph(
     Ok(indexed_graph)
 }
 
+/// Check if the given [`transaction::Version`] is TRUC (Topologically Restricted Until
+/// Confirmation)s.
```
yan-pi: Nit: `(Topologically Restricted Until Confirmation)s` has a stray `s` after the closing parenthesis.

src/wallet/mod.rs
```diff
@@ -2007,6 +2014,29 @@ impl Wallet {
...
+                // only add to optional UTXOs those that follows BIP-431 (TRUC) specification.
+                // see https://github.com/bitcoin/bips/blob/master/bip-0431.mediawiki#specification
+                .filter(|local_output| {
+                    local_output.chain_position.is_confirmed()
+                        || match self.tx_graph().get_tx(local_output.outpoint.txid) {
+                            Some(ancestor_tx) => match is_truc(version) {
+                                // if building TRUC; filter out all unconfirmed non-TRUC.
+                                true => {
+                                    is_truc(ancestor_tx.version)
+                                        && local_output.chain_position.is_unconfirmed()
```
yan-pi: Nit: `local_output.chain_position.is_unconfirmed()` is always true here ? 

The `is_confirmed()` check on line 2020 short-circuits the keep case via `||`, so by the time the match runs, `local_output` is guaranteed unconfirmed. 

Maybe same occurs on line 2031. 
I think you can drop both checks.

yan-pi (commented):
tACK 1d32117.

IMO The `is_truc` helper is more functional and reads cleaner than the XNOR in #442, and definitely is a good foundation for the rest of #477.

Tested it locally by checking out test/unconfirmed-truc-txs and rebuilding the rust-repro binaries from https://github.com/yan-pi/bdk-pr442-repro against the PR via `[dependencies] bdk_wallet = { path = "../../bdk_wallet" }`:

The inverse direction of Rule 2: the non-v3 unconfirmed UTXO was filtered correclty.

One concern not flagged inline: the integration test is great, but maybe the unit-level coverage from #442 is also worth keeping? 

Plus two minor nits and typos.

src/wallet/mod.rs
```diff
@@ -2007,6 +2014,29 @@ impl Wallet {
                         .is_mature(current_height)
                         .then(|| new_local_utxo(k, i, full_txo))
                 })
+                // only add to optional UTXOs those that follows BIP-431 (TRUC) specification.
+                // see https://github.com/bitcoin/bips/blob/master/bip-0431.mediawiki#specification
+                .filter(|local_output| {
+                    local_output.chain_position.is_confirmed()
```
lorenzolfm: i think this logic can be simplified.

Maybe something like

```rust
                    if local_output.chain_position.is_confirmed() {
                        return true;
                    }

                    let Some(anscestor_tx) = self.tx_graph().get_tx(local_output.outpoint.txid)
                    else {
                        return true;
                    };

                    is_truc(version) && is_truc(anscestor_tx.version)
```

oleonardolima: Thanks! We can't use the `&&` because we're filtering out different things, though it looks much better like this:

``` rust
if local_output.chain_position.is_confirmed() {
    return true;
}

let Some(ancestor_tx) = self.tx_graph().get_tx(local_output.outpoint.txid)
else {
    return true;
};

match is_truc(version) {
    // if building TRUC; filter out all unconfirmed non-TRUC.
    true => is_truc(ancestor_tx.version),
    // if building non-TRUC; filter out all unconfirmed TRUC.
    false => !is_truc(ancestor_tx.version),
}
```

oleonardolima: done in cd9290cccf387478f34b4da7da0a25364ec47b46

src/wallet/mod.rs
oleonardolima: I updated it to used what we discussed with @lorenzolfm, solved in cd9290cccf387478f34b4da7da0a25364ec47b46

lorenzolfm (approved):
ACK 85143c28a93d5ba322ef07adeb69e4c0eaa8e93a

lorenzolfm (approved):
ACK 6145afc49239b9fade7a0058b3ad4ecf4599a916

src/wallet/mod.rs
```diff
@@ -2058,6 +2065,27 @@ impl Wallet {
...
+                    else {
+                        return true;
+                    };
+
+                    match is_truc(version) {
+                        // if building TRUC; filter out all unconfirmed non-TRUC.
+                        true => is_truc(ancestor_tx.version),
+                        // if building non-TRUC; filter out all unconfirmed TRUC.
+                        false => !is_truc(ancestor_tx.version),
+                    }
```
ValuedMammal: I think #442 had the right idea with the bidirectional filter.
```suggestion
                    local_output.chain_position.is_confirmed()
                        || self
                            .tx_graph()
                            .get_tx(local_output.outpoint.txid)
                            .is_some_and(|tx| (tx.version == Version(3)) == (version == Version(3)))
```

src/wallet/mod.rs
```diff
@@ -2942,6 +2970,12 @@ fn make_indexed_graph(
     Ok(indexed_graph)
 }
 
+/// Check if the given [`transaction::Version`] is TRUC (Topologically Restricted Until
+/// Confirmation).
+fn is_truc(version: transaction::Version) -> bool {
+    version.eq(&Version(3))
+}
```
ValuedMammal: If going for a private helper, I'd argue this is clearer.
```suggestion
/// The transaction is considered to have opted in to v3 TRUC policy if the transaction version is 3
/// (BIP431).
fn has_opted_into_v3_truc_policy(tx: &Transaction) -> bool {
    tx.version == Version(3)
}
```

ValuedMammal (changes requested):

src/wallet/mod.rs
oleonardolima: I don't see the difference, it's filtering the same thing. However, having the predicate makes the code easier to follow.

src/wallet/mod.rs
oleonardolima: The docs looks better, though we don't have the full transaction when creating it, only the version is available.

src/wallet/mod.rs
oleonardolima: @ValuedMammal I'd say something like the code below is better, though it's does not read well for users not aware of what's happening. 

```
                    local_output.chain_position.is_confirmed()
                        || self
                            .tx_graph()
                            .get_tx(local_output.outpoint.txid)
                            .is_some_and(|tx| is_truc(tx.version) && is_truc(version))
```

ValuedMammal: > I don't see the difference

The else branch of the let-else is logically unreachable; if we have the FullTxOut, then we also have the transaction.

You used a match statement to test a boolean expression. if/else would be the idiomatic approach.

> I'd say something like the code below is better

You're missing the case where `!is_truc(tx.version) && !is_truc(version)`.

> though it's does not read well for users not aware of what's happening

You have a code comment saying the filter implements BIP431 rule 2, which should be enough, but to be extra clear you can say something like this

```rust
/// If the target version is 3, the unconfirmed output must be version 3. If the target version
/// is not 3, the output must not be version 3. In either case if the output is confirmed,
/// then it may be selected.
```

Does that help @oleonardolima ?

src/wallet/mod.rs
ValuedMammal: For what it's worth we constructed the tx on L1399. `is_truc(version)` isn't a useful abstraction and tells the reader nothing about how it's implemented, but since the implementation is trivial, you might as well evaluate this condition inline. Does that help @oleonardolima ?

ValuedMammal (changes requested):
#442 added docs explaining the manual selection caveat https://github.com/bitcoindevkit/bdk_wallet/pull/442#discussion_r3166888824, but that seems to be missing from this branch.

src/wallet/mod.rs
oleonardolima: Thanks! Got it, I didn't notice the note above regarding the usage of `get_tx_node` infallibly.

I still think that the predicate helps, however as it's only being used here the inline check is enough.

I updated it to use it inline, and updated your suggested docs to follow the BIP terminology, I kept the confirmed step separate as early return so the code reads better.

src/wallet/mod.rs
oleonardolima: Yes, but passing the transaction around does not seem good. Anyway, as it's only being used here I removed to use the inline check only.

oleonardolima:
> #442 added docs explaining the manual selection caveat [#442 (comment)](https://github.com/bitcoindevkit/bdk_wallet/pull/442#discussion_r3166888824), but that seems to be missing from this branch.

@ValuedMammal Thanks! I updated the code following some of the suggestions, let me know if there's anything else to address.

ValuedMammal (approved):
ACK 7982b6e8fe0b47984d0b148b325f8b4bd9e44481; ran the tests locally

tests/wallet.rs
```diff
@@ -3331,3 +3332,141 @@ fn test_tx_ordering_untouched_preserves_insertion_ordering_bnb_success() {
...
+    let _ = wallet.sign(&mut psbt, SignOptions::default())?;
+    let tx_c = psbt.extract_tx()?;
+
+    // txC MUST ONLY use the available confirmed UTXOs AND/OR unconfirmed TRUC UTXOs.
+    assert!(
+        tx_c.input
+            .iter()
+            .all(|txin| txin.previous_output.txid.ne(&txid_b)),
+        "SHOULD NOT try to spend an unconfirmed non-TRUC output in a TRUC tx!"
+    );
```
nymius, line 3443: Why not checking we are funding the TRUC tx with another TRUC tx?

nymius (approved):
ACK 7982b6e8fe0b47984d0b148b325f8b4bd9e44481

This solution is inline with the current approach for UTXO filtering, so I agree with it, however, the number of constraints we impose on the local utxo seems to be always growing.
I usually come back to the idea that this filtering should be done in a detachable pipeline, with customizable filters rather than manually adding new ones each time we need.
In that case I would prefer to create a wrapper around txouts injecting all the information we need to filter them later, or providing an oracle that can be called with custom made closure to apply these filters.
Has anyone have any ideas about this?

oleonardolima:
> ACK [7982b6e](https://github.com/bitcoindevkit/bdk_wallet/commit/7982b6e8fe0b47984d0b148b325f8b4bd9e44481)
> 
> This solution is inline with the current approach for UTXO filtering, so I agree with it, however, the number of constraints we impose on the local utxo seems to be always growing. I usually come back to the idea that this filtering should be done in a detachable pipeline, with customizable filters rather than manually adding new ones each time we need. In that case I would prefer to create a wrapper around txouts injecting all the information we need to filter them later, or providing an oracle that can be called with custom made closure to apply these filters. Has anyone have any ideas about this?

In the new create_psbt and PsbtParams flow, you could have a custom filter instead of this always-growing one, doesn't that solve this issue ? Let me know if I'm missing your point here.

tests/wallet.rs
oleonardolima: Yes, that would also work.

PR #479, opened 2026-05-05
fix(coin_selection): avoid per-utxo fee rounding

benthecarman:
select_sorted_utxos computed fee_rate * weight per utxo and summed
the results. Because FeeRate * Weight ceils to whole sats, each
input added up to 0.999 sats of fractional fee. The error grew
linearly with the input count: at 1 sat/vb on 271-wu inputs
(67.75 sat each), four inputs overpay by 1 sat; ten overpay by
2 sats; large selections drift further.

Accumulate input weight during selection and compute the fee once
on the total. Ceiling now happens at most once for the whole tx,
so the result equals the fee a sender would compute by weighing
the finished transaction and never exceeds the requested fee rate
by more than one sat over the whole transaction.

This was found when comparing fee calculation differences between LDK
and BDK.

#### All Submissions:

#### Bugfixes:

### DIFF
--- a/src/wallet/coin_selection.rs
+++ b/src/wallet/coin_selection.rs
@@ -326,17 +326,16 @@ fn select_sorted_utxos(
     drain_script: &Script,
 ) -> Result<CoinSelectionResult, InsufficientFunds> {
     let mut selected_amount = Amount::ZERO;
-    let mut fee_amount = Amount::ZERO;
+    let mut total_weight = Weight::ZERO;
     let selected = utxos
         .scan(
-            (&mut selected_amount, &mut fee_amount),
-            |(selected_amount, fee_amount), (must_use, weighted_utxo)| {
-                if must_use || **selected_amount < target_amount + **fee_amount {
-                    **fee_amount += fee_rate
-                        * TxIn::default()
-                            .segwit_weight()
-                            .checked_add(weighted_utxo.satisfaction_weight)
-                            .expect("`Weight` addition should not cause an integer overflow");
+            (&mut selected_amount, &mut total_weight),
+            |(selected_amount, total_weight), (must_use, weighted_utxo)| {
+                if must_use || **selected_amount < target_amount + fee_rate * **total_weight {
+                    **total_weight += TxIn::default()
+                        .segwit_weight()
+                        .checked_add(weighted_utxo.satisfaction_weight)
+                        .expect("`Weight` addition should not cause an integer overflow");
                     **selected_amount += weighted_utxo.utxo.txout().value;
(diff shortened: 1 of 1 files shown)


ValuedMammal:
I see what you mean about avoiding per-utxo fee rounding, just confused by the "floors to whole sats" comment since bitcoin's [`impl Mul<FeeRate> for Weight`](https://docs.rs/bitcoin/latest/bitcoin/struct.FeeRate.html#impl-Mul%3CFeeRate%3E-for-Weight) states that it uses a ceiling computation, so rounding up instead of down?

benthecarman:
Ah yeah, that should be `ceils`

ValuedMammal:
> Ah yeah, that should be ceils

So is BDK underpaying the fee? If so could it be for a different reason than per-UTXO rounding in `select_sorted_utxos`?

benthecarman:
> So is BDK underpaying the fee? If so could it be for a different reason than per-UTXO rounding in select_sorted_utxos?

No it is slightly overpaying because we are rounding up every input. My commit message/description wrote it wrong, sorry. Updated

src/wallet/coin_selection.rs
```diff
@@ -996,6 +996,46 @@ mod test {
         assert_eq!(result.fee_amount, Amount::from_sat(68));
     }
 
+    #[test]
+    fn test_select_sorted_utxos_accumulates_weight_before_fee() {
+        // Per-input weight of 271 wu makes (1 sat/vb) * weight = 67.75 sat,
+        // so flooring per-utxo loses 0.75 sat each. Selecting 4 such inputs
+        // and summing the floors yields 268 sat; computing the fee once on
+        // the accumulated weight yields 271 sat.
```
ValuedMammal, line 1004:
```suggestion
        // Per-input weight of 271 wu makes (1 sat/vb) * weight = 67.75 sat,
        // so rounding per-utxo gains 0.25 sat each. Selecting 4 such inputs
        // and summing the error yields 272 sat; computing the fee once on
        // the accumulated weight yields 271 sat.
```

benthecarman: fixed. sorry relied on claude too much :sweat_smile:

ValuedMammal (approved):
ACK daf8d810a7e2bddd1bee5c1cedf38e9292f92bf5; ran the tests locally

110CodingP (approved):
ACK [daf8d81](https://github.com/bitcoindevkit/bdk_wallet/pull/479/commits/daf8d810a7e2bddd1bee5c1cedf38e9292f92bf5).
Also verified that the new test fails without the changes.

PR #480, opened 2026-05-09
feat(wallet)!: include non-canonical txs in wallet history

### LINKED ISSUE #295: Preserve evicted outgoing transactions in wallet / Improve "WalletTx"
## We should ensure correct behavior in the following scenarios

**Scenario 1:** Outgoing transaction is evicted from the mempool due to insufficient fees.

- **Current behavior (`master`):** The transaction disappears from the wallet’s transaction list.  
- **Risk with current behavior:** The caller may create another outgoing transaction, and the coin selection algorithm may choose outputs that can coexist in the same 
  subgraph as the original transaction. This can result in an accidental double payment if the original transaction later reappears in the mempool or gets confirmed.  
- **Ideal behavior:** The transaction remains in the wallet’s transaction list. We can see that it is uncanonical and does not conflict with any canonical transactions. 
  The caller can then choose to fee-bump or cancel it.

**Scenario 2:** Outgoing transaction is evicted from the mempool because its parent (an incoming transaction) was cancelled.

- **Current behavior (`master`):** The transaction disappears from the wallet’s transaction list.  
- **Risk with current behavior:** Same risk as in Scenario&nbsp;1.  
- **Ideal behavior:** The transaction remains in the wallet’s transaction list. We can see that one of its inputs conflicts with the cancellation transaction.  
  - If the transaction has only this one input, the caller should have the option to wait until the conflicting cancellation transaction reaches a chosen confirmation 
    threshold before forgetting the original transaction.  
  - If the transaction has multiple inputs, the caller should have the option to replace or cancel it.  
- **Additional notes:** Spending from unconfirmed incoming transactions should be discouraged to avoid this scenario. There should be an option allowing the caller to set 
  a minimum confirmation count before an incoming transaction becomes eligible for spending.

**Scenario 3:** A chain reorg unconfirms a previously confirmed transaction, and it is **not re-introduced** to the mempool.

- **Current behavior (`master`):** The transaction can vanish from view, alongside it's descendants.
- **Risk with current behavior:** The caller may create a duplicate or conflicting payments for itself and all of it's descendants.
- **Ideal behavior:** The transaction and descendants remain in the wallet's transaction list. The caller can check whether these newly-uncanonical transactions has conflicts and whether those conflicts are confirmed or not. Can we somehow merge/replace the whole package?

> TODO: More scenarios

## Notes on RBF

In `bdk_wallet::Wallet`, the current RBF implementation reuses all inputs from the original transaction.  
This guarantees that any subsequent RBF cannot be part of the same subgraph as the original transaction or its earlier replacements, reducing the risk of accidental duplicate payments. This is also the [behavior in Bitcoin Core](https://github.com/bitcoin/bitcoin/blob/26e9db2df0762e41b1964cdd4c67e5bc2cacf404/src/wallet/feebumper.cpp#L202).

We should incentivize this behavior for now.

noahjoeris:
### Description

`WalletTx` was an alias for `CanonicalTx`, so `wallet.transactions()` / `wallet.get_tx(txid)` only returned **canonical** wallet-relevant txs. Replaced/evicted txs disappeared even when still wallet-relevant.

This PR renames `WalletTx` → `TransactionInfo` and surfaces **all wallet-relevant txs**, including non-canonical ones:

```rust
pub struct TransactionInfo {
    pub details: TxDetails,              // includes chain_position: Option<…>
    pub anchors: BTreeSet<ConfirmationBlockTime>,
    pub first_seen: Option<u64>,
    pub last_seen: Option<u64>,
    pub evicted_at: Option<u64>,
    pub conflicts: Vec<Txid>,            // direct input conflicts
}
```

Fixes #295

**Supported use cases**
- Keep dropped txs visible in wallet history.
- Show whether an evicted tx has conflicts.
- Let users decide when it is safe to forget an old dropped tx.
- Give users enough information to warn before accidentally paying twice.

### Notes to the reviewers

- Performance of `transactions()` takes a small hit. Worth confirming whether it's an issue in practice.
- Still need to add tests.

### Changelog notice
```
- Changed: `WalletTx` → owned `TransactionInfo` (details, anchors/seen/evicted, direct conflicts).
- Changed: `transactions()` / `get_tx()` also return non-canonical wallet-relevant txs.
- Changed: `TxDetails::chain_position` → `Option<ChainPosition<_>>` (`None` = non-canonical).
- Added: `Wallet::conflicts(txid)` (ancestor-aware).
```

#### All Submissions:

- [x] I've signed all my commits
- [x] I ran `just p` before pushing

#### New Features:

- [x] I've added tests for the new feature
- [x] I've added docs for the new feature
- [x] This pull request breaks the existing API

### DIFF
--- a/src/wallet/mod.rs
+++ b/src/wallet/mod.rs
@@ -30,7 +30,7 @@ use bdk_chain::{
         FullScanRequest, FullScanRequestBuilder, FullScanResponse, SyncRequest, SyncRequestBuilder,
         SyncResponse,
     },
-    tx_graph::{CalculateFeeError, CanonicalTx, TxGraph, TxUpdate},
+    tx_graph::{CalculateFeeError, TxGraph, TxNode, TxUpdate},
     BlockId, CanonicalizationParams, ChainPosition, ConfirmationBlockTime, DescriptorExt,
     FullTxOut, Indexed, IndexedTxGraph, Indexer, Merge,
 };
@@ -42,7 +42,7 @@ use bitcoin::{
     secp256k1::Secp256k1,
     sighash::{EcdsaSighashType, TapSighashType},
     transaction, Address, Amount, Block, FeeRate, Network, NetworkKind, OutPoint, Psbt, ScriptBuf,
-    Sequence, SignedAmount, Transaction, TxOut, Txid, Weight, Witness,
+    Sequence, Transaction, TxOut, Txid, Weight, Witness,
 };
 use miniscript::{
     descriptor::KeyMap,
@@ -64,7 +64,7 @@ pub mod signer;
 pub mod tx_builder;
 pub(crate) mod utils;
 
-use crate::collections::{BTreeMap, HashMap, HashSet};
+use crate::collections::{BTreeMap, BTreeSet, HashMap, HashSet};
 use crate::descriptor::{
     check_wallet_descriptor, error::Error as DescriptorError, policy::BuildSatisfaction,
     DerivedDescriptor, DescriptorMeta, ExtendedDescriptor, ExtractPolicy, IntoWalletDescriptor,
@@ -181,8 +181,49 @@ impl fmt::Display for AddressInfo {
     }
 }
 
-/// A `CanonicalTx` managed by a `Wallet`.
-pub type WalletTx<'a> = CanonicalTx<'a, Arc<Transaction>, ConfirmationBlockTime>;
+/// A wallet-relevant transaction and its metadata.
+#[derive(Clone, Debug)]
+pub struct TransactionInfo<'a> {
+    /// Wallet-specific amounts, fees, and canonicality.
+    pub details: TxDetails,
+    /// Graph metadata such as anchors, first_seen, and last_seen.
+    pub tx_node: TxNode<'a, Arc<Transaction>, ConfirmationBlockTime>,
+    /// Latest backend observation that this tx was absent from the mempool.
+    pub last_evicted: Option<u64>,
+    /// Direct conflicts spending the same inputs.
+    ///
+    /// See [`Wallet::canonical_blockers`] to get canonical blockers, including those via
+    /// ancestors.
+    pub conflicts: Vec<Txid>,
+}
+
+/// The canonicality of a wallet-relevant transaction.
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub enum TxCanonicality {
+    /// The transaction is currently canonical.
+    Canonical(ChainPosition<ConfirmationBlockTime>),
+    /// The transaction is currently not canonical.
+    NonCanonical,
+}
+
+impl From<Option<ChainPosition<ConfirmationBlockTime>>> for TxCanonicality {
+    fn from(chain_position: Option<ChainPosition<ConfirmationBlockTime>>) -> Self {
+        match chain_position {
+            Some(chain_position) => Self::Canonical(chain_position),
+            None => Self::NonCanonical,
+        }
+    }
+}
+
+impl TransactionInfo<'_> {
+    /// Returns the chain position if the transaction is currently canonical.
+    pub fn chain_position(&self) -> Option<&ChainPosition<ConfirmationBlockTime>> {
+        match &self.details.canonicality {
+            TxCanonicality::Canonical(chain_position) => Some(chain_position),
+            TxCanonicality::NonCanonical => None,
+        }
+    }
+}
 
 impl Wallet {
     /// Build a new single descriptor [`Wallet`].
@@ -789,33 +830,6 @@ impl Wallet {
             .map(|((k, i), full_txo)| new_local_utxo(k, i, full_txo))
     }
 
-    /// Get the [`TxDetails`] of a wallet transaction.
-    ///
-    /// If the transaction with txid [`Txid`] cannot be found in the wallet's transactions, `None`
-    /// is returned.
-    pub fn tx_details(&self, txid: Txid) -> Option<TxDetails> {
-        let tx: WalletTx = self.transactions().find(|c| c.tx_node.txid == txid)?;
-
-        let (sent, received) = self.sent_and_received(&tx.tx_node.tx);
-        let fee: Option<Amount> = self.calculate_fee(&tx.tx_node.tx).ok();
-        let fee_rate: Option<FeeRate> = self.calculate_fee_rate(&tx.tx_node.tx).ok();
-        let balance_delta: SignedAmount = self.tx_graph.index.net_value(&tx.tx_node.tx, ..);
-        let chain_position = tx.chain_position;
-
-        let tx_details: TxDetails = TxDetails {
-            txid,
-            received,
-            sent,
-            fee,
-            fee_rate,
-            balance_delta,
-            chain_position,
-            tx: tx.tx_node.tx,
-        };
-
-        Some(tx_details)
-    }
-
     /// List all relevant outputs (includes both spent and unspent, confirmed and unconfirmed).
     ///
     /// To list only unspent outputs (UTXOs), use [`Wallet::list_unspent`] instead.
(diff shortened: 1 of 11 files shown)


src/wallet/mod.rs
```diff
@@ -1030,45 +1100,96 @@ impl Wallet {
...
+
+        if !self.tx_graph.index.is_tx_relevant(&tx_node.tx) {
+            return None;
+        }
+
+        let canonical_chain_positions_by_txid = self.canonical_chain_positions_by_txid();
+        Some(WalletTx {
+            status: self.tx_status(&tx_node, &canonical_chain_positions_by_txid),
+            tx_node,
+        })
```
aagbotemi: This allocates a full HashMap of every canonical wallet tx to resolve a single txid, which is more work than necessary when we only need one txid's position. I think using `find()` over the canonical iterator is better than the HashMap because it doesn't allocate and short-circuits on a canonical hit. What do you think?

noahjoeris: Makes sense. Thanks for the review! I updated.

src/wallet/mod.rs
```diff
@@ -1030,45 +1083,114 @@ impl Wallet {
...
+            .flat_map(|c| {
+                graph
+                    .direct_conflicts(&c)
+                    .map(|(_, txid)| txid)
+                    .collect::<Vec<_>>()
+            })
+            .filter(|t| is_canonical(t))
+            .collect();
+
+        blockers.remove(&txid);
```
aagbotemi: nit: the `is_canonical(t)` filter above only allows canonical txids into `blockers`, and `txid` is non-canonical here, so this line is unreachable. So, I think it's safe to remove.

ValuedMammal:
Concept ACK

noahjoeris:
After some ideas by ValuedMammal I simplified the model a bit:

- Renamed `WalletTx` → `TransactionInfo`, `WalletTxStatus` → `TxCanonicality`.
- Replaced `canonical_blockers()` with a direct-only `conflicts: Vec<Txid>` field; moved `last_evicted` to a top-level field on `TransactionInfo`.

PR description updated accordingly.

ValuedMammal:
`canonical_blockers` was a nice idea because it could tell you if a tx is indirectly conflicted through one of its ancestors, even if it has no direct conflicts of its own. How can we capture that information so it's clear why a wallet tx isn't canonical? Maybe include the tx's unconfirmed ancestor and descendant sets somewhere so we can trace its ancestry. But as mentioned before we may not want to bake a lot of txgraph traversal into every `wallet.transactions()` call

noahjoeris:
Yep I agree. I added `canonical_blockers()` back.

src/wallet/mod.rs
```diff
@@ -181,8 +181,49 @@ impl fmt::Display for AddressInfo {
...
+}
+
+/// The canonicality of a wallet-relevant transaction.
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub enum TxCanonicality {
+    /// The transaction is currently canonical.
+    Canonical(ChainPosition<ConfirmationBlockTime>),
+    /// The transaction is currently not canonical.
+    NonCanonical,
+}
```
ValuedMammal: TxCanonicality looks like a reimplementation of `Option<T>`, so I'd remove it in favor of `Option<ChainPosition>`. None means "not canonical". The other reason is to avoid leaking the "canonical" jargon into the wallet API.

src/wallet/mod.rs
```diff
@@ -181,8 +181,49 @@ impl fmt::Display for AddressInfo {
...
 
-/// A `CanonicalTx` managed by a `Wallet`.
-pub type WalletTx<'a> = CanonicalTx<'a, Arc<Transaction>, ConfirmationBlockTime>;
+/// A wallet-relevant transaction and its metadata.
+#[derive(Clone, Debug)]
+pub struct TransactionInfo<'a> {
+    /// Wallet-specific amounts, fees, and canonicality.
+    pub details: TxDetails,
+    /// Graph metadata such as anchors, first_seen, and last_seen.
+    pub tx_node: TxNode<'a, Arc<Transaction>, ConfirmationBlockTime>,
```
ValuedMammal: In `TransactionInfo` flatten the tx_node into separate first-seen and last-seen fields, clone out the anchors set, and remove the lifetime `'a`. The details already contain the transaction and txid. So you'd have

```rust
pub struct TransactionInfo {
    pub details: TxDetails,
    pub anchors: BTreeSet<ConfirmationBlockTime>,
    pub first_seen: Option<u64>,
    pub last_seen: Option<u64>,
    pub evicted_at: Option<u64>,
    pub conflicts: Vec<Txid>,
}
```

src/wallet/mod.rs
```diff
@@ -1000,114 +1014,236 @@ impl Wallet {
...
+
+    /// Return canonical transactions that block a wallet transaction from being canonical.
+    ///
+    /// This includes canonical transactions that directly conflict with `txid`, and canonical
+    /// transactions that directly conflict with a non-canonical ancestor of `txid`. Ancestors are
+    /// checked because a transaction is also blocked if one of its ancestors was replaced.
+    ///
+    /// Returns `None` if `txid` is unknown or not wallet-relevant; returns `Some(empty)` if `txid`
+    /// is canonical or no canonical blockers are known.
+    pub fn canonical_blockers(&self, txid: Txid) -> Option<Vec<Txid>> {
```
ValuedMammal: Change `canonical_blockers` to
```rust
fn conflicts(&self, txid) -> Result<Vec<(Txid, Option<ChainPosition>), UnknownTransaction>
```
that way finding a confirmed conflict of `txid` can be done by iterating over the conflicts to see if one is confirmed. With `Option<Vec<...>>` the caller might wrongly assume None implies the tx has no conflicts if in fact it was never found.

src/wallet/mod.rs
```diff
@@ -1000,114 +1014,236 @@ impl Wallet {
...
+        let mut non_canonical_txs = tx_graph
+            .full_txs()
+            .filter(|tx_node| !canonical_txids.contains(&tx_node.txid))
+            .filter(|tx_node| tx_index.is_tx_relevant(&tx_node.tx))
+            .map(|tx_node| self.build_transaction_info(tx_node, None))
+            .collect::<Vec<_>>();
+        non_canonical_txs.sort_unstable_by_key(|tx_info| tx_info.tx_node.txid);
+        wallet_txs.extend(non_canonical_txs);
+
+        wallet_txs.into_iter()
```
ValuedMammal: `Wallet::transactions` doesn't have any documented guarantee about the returned order so I'm not sure it's worth sorting the non-canonical set by Txid as opposed to just chaining the two iterators together. If a caller needs a specific ordering they can use `transactions_sort_by`.

src/wallet/mod.rs
```diff
@@ -789,33 +830,6 @@ impl Wallet {
             .map(|((k, i), full_txo)| new_local_utxo(k, i, full_txo))
     }
 
-    /// Get the [`TxDetails`] of a wallet transaction.
-    ///
-    /// If the transaction with txid [`Txid`] cannot be found in the wallet's transactions, `None`
-    /// is returned.
-    pub fn tx_details(&self, txid: Txid) -> Option<TxDetails> {
```
ValuedMammal, line 796: Why get rid of Wallet::tx_details; it could be left as a convenience for get_tx().details

src/wallet/mod.rs
```diff
@@ -1000,114 +1014,236 @@ impl Wallet {
...
+    }
+
+    /// Return canonical transactions that block a wallet transaction from being canonical.
+    ///
+    /// This includes canonical transactions that directly conflict with `txid`, and canonical
+    /// transactions that directly conflict with a non-canonical ancestor of `txid`. Ancestors are
+    /// checked because a transaction is also blocked if one of its ancestors was replaced.
+    ///
+    /// Returns `None` if `txid` is unknown or not wallet-relevant; returns `Some(empty)` if `txid`
+    /// is canonical or no canonical blockers are known.
```
oleonardolima: This is not accurate, a canonical transaction cannot block another one from being canonical.

oleonardolima (commented):
I don't fully grasp the rationale for the changes. I mean if you need specific group of transactions (e.g non-canonical, evicted, and so on), you could have other methods for those specific scenarios, and even a newer single one that list everything (_same as accessing the inner TxGraph_).

It's valid to notice that TxGraph is monotone, so if we are getting all transactions for the Wallet being canonical or not, so the list can be really big, and basically a duplicate of what already exists there.

noahjoeris:
Thanks for the reviews! 



> I don't fully grasp the rationale for the changes. I mean if you need specific group of transactions (e.g non-canonical, evicted, and so on), you could have other methods for those specific scenarios, and even a newer single one that list everything (_same as accessing the inner TxGraph_).
> 
> It's valid to notice that TxGraph is monotone, so if we are getting all transactions for the Wallet being canonical or not, so the list can be really big, and basically a duplicate of what already exists there.

It addresses the problem stated in #295 that relevant non-canonical txs (fee-evicted, replaced, reorged) disappear from transactions(), so callers can miss them and accidentally double-pay. Ideal behavior is to keep them in the default wallet history with status/conflicts so apps can act accordingly. A separate method / raw TxGraph leaves that safe path opt-in.

Users can rebuild this from existing APIs, but they could get it wrong and it's more convenient this way.
And I assume non-canonical relevant txs are usually rare so the list of txs won't be much longer. Perf-wise we can lighten transactions() and move sent/received/fee/fee_rate/net_value to on-demand tx_details if we want.

PR #482, opened 2026-05-11
Fix wallet name generation to use explicit public descriptor checksum

### LINKED ISSUE #481: Clarify `wallet_name_from_descriptor` behavior and compute checksum explicitly
### **Description**
The [wallet_name_from_descriptor](https://github.com/bitcoindevkit/bdk_wallet/blob/6dca318c9d29bec5bedfed27a6d36a21e6bc8204/src/wallet/mod.rs#L2790) currently relies on an implicit [to_string()](https://github.com/bitcoindevkit/bdk_wallet/blob/6dca318c9d29bec5bedfed27a6d36a21e6bc8204/src/wallet/mod.rs#L2804) contract to extract the descriptor checksum. The docs do not clearly state how the wallet name is derived, and users may not know that private descriptor input should produce the same name as the equivalent public form.

AmosOO7:
### Summary
- Update [wallet_name_from_descriptor](https://github.com/bitcoindevkit/bdk_wallet/blob/6dca318c9d29bec5bedfed27a6d36a21e6bc8204/src/wallet/mod.rs#L2790) to compute checksum directly from parsed public descriptor strings
- Remove reliance on [to_string()](https://github.com/bitcoindevkit/bdk_wallet/blob/6dca318c9d29bec5bedfed27a6d36a21e6bc8204/src/wallet/mod.rs#L2804) implicitly containing #checksum
- Clarify in docs that wallet names are defined by public descriptor checksums
- Add regression coverage for equivalent xpub/xprv descriptors producing the same wallet name

## Testing
```
cargo test test_wallet_name_from_descriptor_public_key_check
```

#### All Submissions:

Closes #481

### DIFF
--- a/src/wallet/mod.rs
+++ b/src/wallet/mod.rs
@@ -2784,16 +2784,19 @@ impl AsRef<bdk_chain::tx_graph::TxGraph<ConfirmationBlockTime>> for Wallet {
     }
 }
 
-/// Generate a deterministic wallet name from descriptor checksums.
+/// Generate a deterministic wallet name from the provided descriptors.
 ///
-/// The wallet name is computed from the parsed public descriptor:
-/// - the checksum of the parsed public descriptor
-/// - if present, the checksum of the parsed change descriptor is appended
+/// The wallet name is the concatenation of the [checksum] of the external and (if provided)
+/// internal public descriptors. If descriptors containing private keys are provided, the name
+/// is computed from the corresponding public descriptors; the result is identical to calling
+/// this function with the equivalent public (xpub) descriptors.
 ///
-/// This is based on the public descriptor form, so equivalent xprv and xpub
-/// inputs produce the same wallet name.
+/// # Errors
 ///
-/// Returns an error if descriptor parsing fails or if checksum computation fails.
+/// If descriptor parsing fails or if checksum computation fails then a [`DescriptorError`] is
+/// returned.
+///
+/// [checksum]: crate::descriptor::checksum::calc_checksum
 pub fn wallet_name_from_descriptor<T>(
     descriptor: T,
     change_descriptor: Option<T>,
--- a/src/wallet/mod.rs
+++ b/src/wallet/mod.rs
@@ -3096,5 +3096,7 @@ mod test {
         );
         assert!(result_with_change.is_ok());
         // Wallet name should be main_checksum + change_checksum
+        let wallet_name = result_with_change.unwrap();
+        assert_eq!(wallet_name, "vn4aqs37jgrerlc3");
     }
 }
--- a/src/wallet/mod.rs
+++ b/src/wallet/mod.rs
@@ -66,9 +66,9 @@ pub(crate) mod utils;
 
 use crate::collections::{BTreeMap, HashMap, HashSet};
 use crate::descriptor::{
-    check_wallet_descriptor, error::Error as DescriptorError, policy::BuildSatisfaction,
-    DerivedDescriptor, DescriptorMeta, ExtendedDescriptor, ExtractPolicy, IntoWalletDescriptor,
-    Policy, XKeyUtils,
+    check_wallet_descriptor, checksum::calc_checksum, error::Error as DescriptorError,
+    policy::BuildSatisfaction, DerivedDescriptor, DescriptorMeta, ExtendedDescriptor,
+    ExtractPolicy, IntoWalletDescriptor, Policy, XKeyUtils,
 };
 use crate::psbt::PsbtUtils;
 use crate::types::*;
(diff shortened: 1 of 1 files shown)


src/wallet/mod.rs
```diff
@@ -2797,16 +2797,23 @@ where
     T: IntoWalletDescriptor,
 {
     // TODO: check descriptors contains only public keys
-    let descriptor = descriptor
-        .into_wallet_descriptor(secp, network_kind)?
-        .0
-        .to_string();
+    let (descriptor, keymap) = descriptor.into_wallet_descriptor(secp, network_kind)?;
+    if !keymap.is_empty() {
+        return Err(DescriptorError::Miniscript(miniscript::Error::Unexpected(
```
busayo-OD: Since there are already dedicated variants for validation rules like this (HardenedDerivationXpub and ExternalAndInternalAreTheSame), would it make sense to add one here so callers can match on this condition without inspecting the message string?

src/wallet/mod.rs
```diff
@@ -3064,4 +3071,56 @@ mod test {
...
+        let result = wallet_name_from_descriptor(
+            public_change_without_private,
+            Some(public_descriptor),
+            NetworkKind::Test,
+            &secp,
+        );
+        assert!(result.is_ok());
+
+        // Test with empty descriptor
+        let public_empty = "";
```
busayo-OD: This test returns early at parse_descriptor before keymap.is_empty() is ever checked. It's testing parse error propagation, not the new check. Maybe clarify the comment to reflect that?

src/wallet/mod.rs
```diff
@@ -3064,4 +3071,56 @@ mod test {
...
+            Some(private_descriptor),
+            NetworkKind::Test,
+            &secp,
+        );
+        assert!(matches!(
+            result,
+            Err(DescriptorError::Miniscript(Unexpected(..)))
+        ));
+        // Test with change descriptor containing public keys (should be ok)
+        let public_change_without_private = "wpkh([9a6a2580/84'/1'/0']tpubDDnGNapGEY6AZAdQbfRJgMg9fvz8pUBrLwvyvUqEgcUfgzM6zc2eVK4vY9x9L5FJWdX8WumXuLEDV5zDZnTfbn87vLe9XceCFwTu9so9Kks/1/*)";
```
aagbotemi: This is a duplicate of `public_change` above, as they have same descriptor string. One can be removed.

aagbotemi (commented):
Well done working on this @AmosOO7. The docs for `wallet_name_from_descriptor` look slightly stale.

src/wallet/mod.rs
```diff
@@ -2797,16 +2797,23 @@ where
     T: IntoWalletDescriptor,
 {
     // TODO: check descriptors contains only public keys
```
aagbotemi: This can be removed now that the check is implemented

src/wallet/mod.rs
AmosOO7: Thanks, after the comments from @ValuedMammal in #481, this will be removed entirely.

src/wallet/mod.rs
AmosOO7: Thanks, This will be removed entirely.

src/wallet/mod.rs
AmosOO7: Nice catch, thanks

src/wallet/mod.rs
AmosOO7: Thanks, nice catch

ValuedMammal:
In ab1d73b959d976d15bc0ac702beb5908c3f89e1e

For clarity you should have somewhere in the test the expected wallet name, for example

```rs
assert_eq!(public_result.as_ref().unwrap(), "vn4aqs37");
```

AmosOO7:
> In [ab1d73b](https://github.com/bitcoindevkit/bdk_wallet/commit/ab1d73b959d976d15bc0ac702beb5908c3f89e1e)
> 
> For clarity you should have somewhere in the test the expected wallet name, for example
> 
> ```rust
> assert_eq!(public_result.as_ref().unwrap(), "vn4aqs37");
> ```

Thank you for that, I will update that now.

ValuedMammal:
Since we're touching this part of the code we should make the documentation super clear so users know what to expect

- Deterministic calculation
- Checksum of the parsed public descriptor
- If present, the change descriptor checksum is appended
- Accepts private descriptors, and produces the same name as the public form
- Errors if descriptor parsing fails, or computing the checksum failed

ValuedMammal:
We should update the PR description as well once you've settled on a direction. If you're using links in the PR description, they're not showing up well 🧐

AmosOO7:
> Since we're touching this part of the code we should make the documentation super clear so users know what to expect
> 
> * Deterministic calculation
> * Checksum of the parsed public descriptor
> * If present, the change descriptor checksum is appended
> * Accepts private descriptors, and produces the same name as the public form
> * Errors if descriptor parsing fails, or computing the checksum failed

Will push an update for this.

ValuedMammal:
ACK 18ea7eea45a9160d99e5557f6889b0b32792f7d8

ValuedMammal:
- Added assertion to `test_wallet_name_from_descriptor_public_key_check` when change descriptor is present
- Updated documentation for `wallet_name_from_descriptor`

ACK 687c4153ef3d6627bc1d4428841baddfb4f87c5e

ValuedMammal (approved):

Issue #52, opened 2024-11-14
Sanity check PSBT before signing


notmandatory:
"Wallet::sign, which could presumably be called with externally-provided PSBTs, calls update_psbt_with_descriptor which assumes multiple invariants on the PSBT. The PSBT is not sanity checked beforehand and would therefore make it possible to crash an application using the library and exposing a sign endpoint. The invariants are assumed in PsbtUtils::get_utxo_for's implementation for Psbt (called from update_psbt_with_descriptor): it assumes the inner tx has at least as many inputs as the PSBT and it assumes that the transaction provided in the PSBT input's non_witness_utxo field has does contain the output index referenced by the inner transaction."

"In the signer module, the previous transaction contained in a PSBT input is not validated against the outpoint for legacy and segwit v0 transactions. This is checked when creating a transaction, but this module may be used to sign a PSBT as an external participant."

oleonardolima:
It won't be applicable after #70 and #235 in v3.0 release, as the `signers` module and its `Wallet::sign` APIs will be removed. Unless we choose to have a helper sanity check method, to be used by the users before the use of `Psbt::sign`, it could be in rust-bitcoin, though.

nymius:
I agree rust-bitcoin, or at least the PSBT code in rust-bitcoin org is better suited for this, but shouldn't be a blocker to start implementing some logic here and later migrate it there, for anyone interested in trying to implement this now.
I did that recently in `bdk-sp` with `send::psbt::derive_sp`.

PR #486, opened 2026-05-12
fix(psbt): sanity check psbt before signing

sdmg15:
### Description
This PR attempts to sanity check the PSBT before signing. It closes audit issue #52

#### All Submissions:

#### Bugfixes:

### DIFF
--- a/tests/wallet.rs
+++ b/tests/wallet.rs
@@ -17,8 +17,8 @@ use bitcoin::script::PushBytesBuf;
 use bitcoin::sighash::{EcdsaSighashType, TapSighashType};
 use bitcoin::taproot::TapNodeHash;
 use bitcoin::{
-    absolute, transaction, Address, Amount, BlockHash, FeeRate, Network, OutPoint, ScriptBuf,
-    Sequence, SignedAmount, Transaction, TxIn, TxOut, Txid,
+    absolute, transaction, Address, Amount, BlockHash, FeeRate, Network, OutPoint, Psbt, ScriptBuf,
+    Sequence, SignedAmount, Transaction, TxIn, TxOut, Txid, Witness,
 };
 use rand::rngs::StdRng;
 use rand::SeedableRng;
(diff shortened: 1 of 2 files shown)


110CodingP (commented):
Thanks for working on this!
Do you also plan to work on this part of #52 ?
"In the signer module, the previous transaction contained in a PSBT input is not validated against the outpoint for legacy and segwit v0 transactions. This is checked when creating a transaction, but this module may be used to sign a PSBT as an external participant."

tests/wallet.rs
```diff
@@ -3007,3 +3007,61 @@ fn test_tx_ordering_untouched_preserves_insertion_ordering_bnb_success() {
         "UTXOs should be ordered with required first, then selected"
     );
 }
+
+#[test]
+fn test_wallet_sign_rejects_malformed_psbt_input_count_mismatch() {
```
110CodingP, line 3012: Plus  I am sorry I wasn't able to understand what this test is trying to check. Could you please add some more comments here?

sdmg15: This test effectively calls the `wallet::sign` with a malformed Psbt and expects the wallet to return an error instead of panicking.

So what I did here is crafting a raw transaction that initially just has 1 input then transformed it into a Psbt type to which I added more inputs (3 vs 1 in original tx).

110CodingP: I see but how does the input count relate to the error raised?

sdmg15: I'm not sure I get your question but that's the essence of the audit report.
When `Wallet::sign` will be called with the current setup (3 inputs in PSBT but 1 in raw tx), the `get_utxo_for` will return `None` and then the `SignerError::MissingNonWitnessUtxo` will be returned.
Not sure if it answers the question.

110CodingP: We get the error even without the other 2 inputs because the first input itself does not contain a non_witness_utxo!

sdmg15: You're right about the test, it doesn't check anything unless we modify the function `update_psbt_with_descriptor` or `get_utxo_for` to actually return an error type. Maybe the test should just be discarded.

Musab1258:
Thanks @sdmg15 for working on this!

I hope you are aware that PR #235 by @oleonardolima, when completed and merged, will remove the `signers` module altogether from `bdk_wallet` to rely on `rust-bitcoin`'s native `Psbt::sign`, which will eventually make the changes here obsolete. However, since that PR is still a draft and might take time, this PR can still act as a stopgap fix if it is merged on time. 

I reviewed the code changes, and the logic you added to the `get_utxo_for` should safely prevent the crash. However, I think you should address these two things in this PR.


1. Fixing the Test Setup

Currently, `test_wallet_sign_rejects_malformed_psbt_input_count_mismatch` does not test the out-of-bounds logic because the first input is missing a `non_witness_utxo`. `Wallet::sign` catches that very first input during its `trust_witness_utxo` validation block and returns the error before it even attempts to sign the out-of-bounds inputs.

To fix this and allow the test to actually reach the signing phase for the malformed inputs, you just need to inject the `non_witness_utxo` into the first generated input. You can add this right after initializing the PSBT:

```rust
psbt.inputs[0].non_witness_utxo = Some(prev_tx.clone()); 
```

Once you add this, run the test again to see what error the signer actually throws when it hits the out-of-bounds inputs, and update your assert! statement accordingly.



2. Completing the Scope of Issue #52 (Outpoint Validation)

As @110CodingP mentioned earlier, Issue #52 also points out another vulnerability, which is that the signer module currently doesn't validate if the previous transaction contained in the PSBT input actually matches the outpoint for legacy and segwit v0 transactions.

To fix this, you can add a quick validation loop at the beginning of `Wallet::sign`. It would just need to verify that `psbt.inputs[i].non_witness_utxo.txid() == psbt.unsigned_tx.input[i].previous_output.txid`.

Chibey-max:
I think `test_wallet_sign_rejects_malformed_psbt_input_count_mismatch` may be passing before it exercises the malformed input-count path. Since `psbt.inputs[0]` does not have `non_witness_utxo` set, `Wallet::sign` can return `SignerError::MissingNonWitnessUtxo` from the upfront non-witness UTXO check before signing reaches the extra PSBT inputs.

Would it make sense to set `psbt.inputs[0].non_witness_utxo = Some(prev_tx.clone())` before pushing the extra PSBT inputs? That should let the test reach the intended mismatch/out-of-bounds path and prove the new guard is what prevents the malformed PSBT from causing trouble.

Issue #474, opened 2026-05-01
Reachable panic in `Utxo::txout`


ValuedMammal:
`Utxo::txout()` returns `&TxOut` but contains two reachable panics for the `Foreign` variant:

1. `prev_tx.output[outpoint.vout as usize]` — panics if the attached `non_witness_utxo` has fewer outputs than the `outpoint.vout` indicates.
2. `unreachable!("Foreign UTXOs will always have one of these set")` — asserts an invariant the type does not enforce. Because `Utxo::Foreign` is a public enum variant with public fields, a caller can theoretically construct one with an empty `psbt::Input`, making this branch reachable.

The validation that guards against these states lives in `TxBuilder::add_foreign_utxo_with_sequence` but is bypassed by any direct construction of `Utxo::Foreign`.

**To Reproduce**

```rust
let utxo = Utxo::Foreign {
    outpoint: OutPoint::null(),
    sequence: Sequence::MAX,
    psbt_input: Box::new(psbt::Input::default()),
};
utxo.txout();
```

**Expected behavior**

`txout()` should not panic on a value that is constructable through the public API.

**Build environment**

- BDK tag/commit: current master
- OS+version:
- Rust/Cargo version:
- Rust/Cargo target:

**Which backend(s) are relevant (if any)?**
- [x] None / not backend-related

**Is this blocking production use?**
- [ ] Yes
- [x] No

**Additional context**

Raised during review of #445. Three options identified:

1. Return `Option<&TxOut>` from `txout()` (breaking).
2. Make `Utxo::Foreign` wrap a type with private fields — restrict construction to the validated `add_foreign_utxo` path, maintaining the invariant at the type level. `txout()` could then keep returning `&TxOut` correctly (breaking).
3. Short-term non-breaking — collapse the two `if let` arms and the `unreachable!` into a single chained `Option` with an honest `.expect("Foreign UTXOs should have one of witness_utxo, non_witness_utxo set")`.

muhahahmad68:
Would you like me to implement the third option as a short-term non-breaking fix or proceed with the long-term fix?

ValuedMammal:
Yeah I'm in favor of option 3 for now, and we should plan on doing a follow-up to change the function signature to `pub fn txout(&self) -> Option<&TxOut>`.

PR #487, opened 2026-05-13
fix(types): avoid reachable panics in Utxo::txout for Foreign variant

muhahahmad68:
### Description

`Utxo::txout()` contained two reachable panics for the `Foreign` variant:

1. `prev_tx.output[outpoint.vout as usize]` — panics if `non_witness_utxo` has fewer outputs than `outpoint.vout` 
2. `unreachable!("Foreign UTXOs will always have one of these set")` — asserts an invariant the type does not enforce, reachable via direct construction of `Utxo::Foreign` with an empty `psbt::Input`

Addresses part of #474

### Notes to the reviewers

Collapsed the two `if let` arms and the `unreachable!` into a single chained `Option`, replacing the index access with `.get()` to safely handle out-of-bounds vout, and replacing `unreachable!` with `.expect()`.

This is the short-term non-breaking fix. A follow-up will change the function signature to `pub fn txout(&self) -> Option<&TxOut>` as a breaking change targeting Wallet 4.0.0.

#### All Submissions:

#### New Features:

* N/A
#### Bugfixes:

### DIFF
--- a/src/types.rs
+++ b/src/types.rs
@@ -122,17 +122,13 @@ impl Utxo {
                 outpoint,
                 psbt_input,
                 ..
-            } => {
-                if let Some(prev_tx) = &psbt_input.non_witness_utxo {
-                    return &prev_tx.output[outpoint.vout as usize];
-                }
-
-                if let Some(txout) = &psbt_input.witness_utxo {
-                    return txout;
-                }
-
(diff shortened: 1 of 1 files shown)


ValuedMammal (approved):
ACK edb0f523da584cccf4a98a45fd4da707f19f8ade

PR #490, opened 2026-05-26
feat(wallet): add Wallet::sign_with_signers

noahjoeris:
### Description

Adds `Wallet::sign_with_signers`, allowing callers to pass signer containers. `Wallet::sign` now delegates to this new method, so the existing API remains unchanged.

This is a small step toward removing signer ownership from `Wallet` and improving separation of concerns. In the future we can deprecate `Wallet::sign` and remove the wallet signer fields.

### Changelog notice
```
- Added `Wallet::sign_with_signers`.
```

#### All Submissions:

#### New Features:

### DIFF
--- a/src/wallet/mod.rs
+++ b/src/wallet/mod.rs
@@ -1729,11 +1729,6 @@ impl Wallet {
     /// [`SignerOrdering`]. This function returns the `Result` type with an encapsulated `bool` that
     /// has the value true if the PSBT was finalized, or false otherwise.
     ///
-    /// The [`SignOptions`] can be used to tweak the behavior of the software signers, and the way
-    /// the transaction is finalized at the end. Note that it can't be guaranteed that *every*
-    /// signers will follow the options, but the "software signers" (WIF keys and `xprv`) defined
-    /// in this library will.
-    ///
     /// ## Example
     ///
     /// ```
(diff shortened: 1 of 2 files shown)


ValuedMammal (approved):
ACK 24564ede8f961679e3bffb43ce54ba520cf3a1ae

src/wallet/mod.rs
```diff
@@ -1753,7 +1753,64 @@ impl Wallet {
...
+    ///     &[external_signers],
+    ///     SignOptions::default(),
+    /// )?;
+    /// assert!(finalized);
+    /// # Ok::<(), anyhow::Error>(())
+    /// ```
+    pub fn sign_with_signers(
+        &self,
+        psbt: &mut Psbt,
+        signers: &[Arc<SignersContainer>],
```
110CodingP: nit: Do we need the `Arc` ? Like what if we use &[SignersContainer] instead?

src/wallet/mod.rs
110CodingP: nit: Should we move this comment about `SignOptions` to `sign_with_signers`  since we plan to deprecate `sign` method?

110CodingP (approved):

110CodingP:
ACK [24564ed](https://github.com/bitcoindevkit/bdk_wallet/pull/490/commits/24564ede8f961679e3bffb43ce54ba520cf3a1ae)

src/wallet/mod.rs
ValuedMammal: I think we should go with `&[&SignersContainer]`, makes the API cleaner and avoids the BTreeMap allocation in wallet.sign().

ValuedMammal:
ACK 53d8f425132a4f6c3153e0a87565f4c7aba00cff

tests/wallet.rs
```diff
@@ -1368,6 +1369,33 @@ fn test_sign_single_xprv() {
     assert_eq!(extracted.input[0].witness.len(), 2);
 }
 
+#[test]
+fn test_sign_with_signers() {
+    let descriptor = "wpkh(tprv8ZgxMBicQKsPd3EupYiPRhaMooHKUHJxNsTfYuScep13go8QFfHdtkG9nRkFGb7busX4isf6X9dURGCoKgitaApQ6MupRhZMcELAxTBRJgS/*)";
+    let secp = Secp256k1::new();
```
oleonardolima: nit: you don't need this if you build the `Wallet` first, you can use the `wallet.secp_ctx()` as below.

I'd actually build a full test with both external/internal signers, not a blocker though.

oleonardolima (commented):
utACK 53d8f425132a4f6c3153e0a87565f4c7aba00cff

tests/wallet.rs
noahjoeris: Sounds good thanks.

110CodingP:
Thanks for the changes!
ACK [`35f753f`](https://github.com/bitcoindevkit/bdk_wallet/pull/490/commits/35f753f357b5b80859d56bd3f03b39e73ead35e1)

oleonardolima (approved):
utACK 35f753f357b5b80859d56bd3f03b39e73ead35e1

Issue #485, opened 2026-05-11
Rule 5: A TRUC transaction that has an unconfirmed TRUC ancestor cannot have a sigop-adjusted virtual size larger than 1000 vB.


oleonardolima:

oleonardolima:
@yan-pi will work on this one, for some reason i can't assign him here.

Issue #484, opened 2026-05-11
Rule 4: A TRUC transaction cannot have a sigop-adjusted virtual size larger than 10,000 vB.


oleonardolima:

oleonardolima:
@yan-pi will work on this one, for some reason i can't assign him here.

PR #493, opened 2026-06-01
feat(wallet): add support for bip-431 rules 4 and 5

yan-pi:
partially addresses #477
depends on #478
solves #484
solves #485

### Description

Enforces BIP-431 Rules 4 and 5 inside `create_tx`, before returning a PSBT:

- **Rule 4**: rejects any TRUC (nVersion=3) transaction with sigop-adjusted vsize over 10,000 vB
- **Rule 5**: tightens the cap to 1,000 vB when the transaction has an unconfirmed TRUC ancestor

Without this, `build_tx().version(3).finish()` can hand back a PSBT that bitcoind rejects at broadcast.

When the cap is exceeded, `create_tx` returns the new `CreateTxError::TrucSizeExceeded { cap_vb, actual_vb }`, so callers can tell which rule fired and by how much.

Adding a variant to the public `CreateTxError` is a breaking change, this is acceptable since the PR targets the 4.0 release.

This adds `estimate_truc_vsize` as a private helper and reuses the `is_truc` helper introduced in #478.

This does **not** address Rules 1–3 (topology / package limits), that's a separate concern tracked in #477 and #478.

### Notes to the reviewers

The check runs after `coin_select` and before `complete_transaction`.

`CoinSelection::coin_select` consumes `Vec<WeightedUtxo>` and returns `Vec<Utxo>`, so the per-input `satisfaction_weight` is gone by the time we'd want it.

We pre-collect a `HashMap<OutPoint, Weight>` from the `WeightedUtxo` list before handing it to the selector.
The map is only allocated when `version == 3`.

For the Rule 5 ancestor check, foreign UTXOs are conservatively treated as non-TRUC.
Same posture as the existing TRUC filter in `filter_utxos`.

We use plain `weight / 4`. Bitcoind's TRUC check uses sigop-adjusted vsize (`max(weight, sigops * 20) / 4`), which matches plain weight/4 for P2WPKH, P2TR and ordinary P2WSH.

#477 should have a follow-up to track proper sigop accounting.

### Changelog notice

```
Added CreateTxError::TrucSizeExceeded { cap_vb, actual_vb }.
`create_tx` now enforces BIP-431 Rules 4 and 5 (vsize caps) for nVersion=3 transactions.
```

#### All Submissions:

#### New Features:

#### Bugfixes:

### DIFF
--- a/src/wallet/error.rs
+++ b/src/wallet/error.rs
@@ -215,6 +215,16 @@ pub enum CreateTxError {
     MissingNonWitnessUtxo(OutPoint),
     /// Miniscript PSBT error
     MiniscriptPsbt(MiniscriptPsbtError),
+    /// TRUC (BIP-431) virtual size cap exceeded.
+    ///
+    /// `cap_vb == 10_000` means Rule 4 (any TRUC tx).
+    /// `cap_vb == 1_000` means Rule 5 (TRUC tx with unconfirmed TRUC ancestor).
+    TrucSizeExceeded {
+        /// The cap that was exceeded, in virtual bytes.
+        cap_vb: u64,
+        /// The estimated virtual size of the candidate transaction.
+        actual_vb: u64,
+    },
 }
 
 impl fmt::Display for CreateTxError {
@@ -280,6 +290,12 @@ impl fmt::Display for CreateTxError {
             }
             CreateTxError::MiniscriptPsbt(err) => {
                 write!(f, "Miniscript PSBT error: {err}")
+            }
+            CreateTxError::TrucSizeExceeded { cap_vb, actual_vb } => {
+                write!(
+                    f,
+                    "TRUC virtual size cap exceeded: estimated {actual_vb} vB > {cap_vb} vB"
+                )
             }
         }
     }
--- a/src/wallet/mod.rs
+++ b/src/wallet/mod.rs
@@ -1441,6 +1441,23 @@ impl Wallet {
             }
         };
 
+        const TRUC_MAX_VSIZE_VB: u64 = 10_000;
+        const TRUC_CHILD_MAX_VSIZE_VB: u64 = 1_000;
+
+        let is_truc_tx = is_truc(version);
+
+        // BIP-431: keep per-input satisfaction weights for the vsize check below;
+        // coin_select returns plain Utxos and the weights would otherwise be lost.
+        let satisfaction_weights: HashMap<OutPoint, Weight> = if is_truc_tx {
+            required_utxos
+                .iter()
+                .chain(optional_utxos.iter())
+                .map(|w| (w.utxo.outpoint(), w.satisfaction_weight))
+                .collect()
+        } else {
+            HashMap::new()
+        };
+
         // Get drain script.
         let mut drain_index = Option::<(KeychainKind, u32)>::None;
         let drain_script = match params.drain_to {
@@ -1534,6 +1551,39 @@ impl Wallet {
 
         // Sort inputs/outputs according to the chosen algorithm.
         params.ordering.sort_tx_with_aux_rand(&mut tx, rng);
+
+        // BIP-431 Rules 4 and 5: a TRUC transaction's sigop-adjusted vsize is capped at
+        // 10,000 vB, or 1,000 vB when it has an unconfirmed TRUC ancestor.
+        if is_truc_tx {
+            let total_satisfaction_weight: Weight = coin_selection
+                .selected
+                .iter()
+                .filter_map(|u| satisfaction_weights.get(&u.outpoint()).copied())
+                .sum();
+            let estimated_vb = estimate_truc_vsize(tx.weight(), total_satisfaction_weight);
+
+            let has_unconf_truc_ancestor = coin_selection.selected.iter().any(|utxo| match utxo {
+                Utxo::Local(local) if local.chain_position.is_unconfirmed() => self
+                    .tx_graph
+                    .graph()
+                    .get_tx(local.outpoint.txid)
+                    .is_some_and(|tx| is_truc(tx.version)),
+                // Foreign UTXOs carry no chain position; treat them as non-TRUC.
+                Utxo::Local(..) | Utxo::Foreign { .. } => false,
+            });
+
+            let cap_vb = if has_unconf_truc_ancestor {
+                TRUC_CHILD_MAX_VSIZE_VB
+            } else {
+                TRUC_MAX_VSIZE_VB
+            };
+            if estimated_vb > cap_vb {
+                return Err(CreateTxError::TrucSizeExceeded {
+                    cap_vb,
+                    actual_vb: estimated_vb,
+                });
+            }
+        }
 
         let psbt = self.complete_transaction(tx, coin_selection.selected, params)?;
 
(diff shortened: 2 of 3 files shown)


src/wallet/mod.rs
```diff
@@ -1535,6 +1547,44 @@ impl Wallet {
         // Sort inputs/outputs according to the chosen algorithm.
         params.ordering.sort_tx_with_aux_rand(&mut tx, rng);
 
+        // BIP-431 Rules 4 and 5: a TRUC transaction's sigop-adjusted vsize is capped at
+        // 10,000 vB, or 1,000 vB when it has an unconfirmed TRUC ancestor.
+        if is_truc(version) {
```
lorenzolfm: maybe you should do let is_truct = is_truct(version) to avoid calling the function two time

src/wallet/mod.rs
```diff
@@ -1535,6 +1547,44 @@ impl Wallet {
...
+                                .get_tx(local.outpoint.txid)
+                                .is_some_and(|tx| is_truc(tx.version))
+                    }
+                    // Foreign UTXOs carry no chain position; treat them as non-TRUC.
+                    Utxo::Foreign { .. } => false,
+                }
+            });
+
+            let cap_vb = if has_unconf_truc_ancestor {
+                1_000
```
lorenzolfm: consts?

src/wallet/mod.rs
lorenzolfm: Suggestion from PR review with @vinteumorg

```rust
        // BIP-431 Rules 4 and 5: a TRUC transaction's sigop-adjusted vsize is capped at
        // 10,000 vB, or 1,000 vB when it has an unconfirmed TRUC ancestor.
        if is_truc(version) {
            let satisfaction_weights: HashMap<OutPoint, Weight> = required_utxos
                .iter()
                .chain(optional_utxos.iter())
                .map(|w| (w.utxo.outpoint(), w.satisfaction_weight))
                .collect();

            let total_satisfaction_weight: Weight = coin_selection
                .selected
                .iter()
                .filter_map(|u| satisfaction_weights.get(&u.outpoint()).copied())
                .sum();

            let estimated_vb = estimate_truc_vsize(tx.weight(), total_satisfaction_weight);

            let has_unconf_truc_ancestor = coin_selection.selected.iter().any(|utxo| {
                match utxo {
                    Utxo::Local(local) if local.chain_position.is_unconfirmed() => self
                        .tx_graph
                        .graph()
                        .get_tx(local.outpoint.txid)
                        .is_some_and(|tx| is_truc(tx.version)),
                    // Foreign UTXOs carry no chain position; treat them as non-TRUC.
                    Utxo::Local(..) | Utxo::Foreign { .. } => false,
                }
            });

            let cap_vb = if has_unconf_truc_ancestor {
                1_000
            } else {
                10_000
            };

            if estimated_vb > cap_vb {
                return Err(CreateTxError::TrucSizeExceeded {
                    cap_vb,
                    actual_vb: estimated_vb,
                });
            }
        }
```

oleonardolima:
@yan-pi I updated my branch to remove the bdk_testenv dependency, you can rebase this one to get a green CI (though you might need to update your tests).

oleonardolima:
@yan-pi the CI still failing

yan-pi:
Rebased this on top of the updated #478 branch.

I dropped the old `bdk_testenv` e2e test and kept the TRUC Rule 4/5 coverage in-memory.

src/wallet/error.rs
```diff
@@ -215,6 +215,16 @@ pub enum CreateTxError {
...
+    /// TRUC (BIP-431) virtual size cap exceeded.
+    ///
+    /// `cap_vb == 10_000` means Rule 4 (any TRUC tx).
+    /// `cap_vb == 1_000` means Rule 5 (TRUC tx with unconfirmed TRUC ancestor).
+    TrucSizeExceeded {
+        /// The cap that was exceeded, in virtual bytes.
+        cap_vb: u64,
+        /// The estimated virtual size of the candidate transaction.
+        actual_vb: u64,
+    },
```
oleonardolima, line 227: @yan-pi did you manage to do it as a non-breaking change ? I agree that having the error would be the right approach, though it would only land in the next major release.

yan-pi: I dropped the new variant and mapped the TRUC size-cap rejection to the existing `CreateTxError::CoinSelection(InsufficientFunds { .. })` instead.

This one seems to be the best one to handle the current error without the breaking change.

src/wallet/mod.rs
```diff
@@ -1534,6 +1552,40 @@ impl Wallet {
...
+                TRUC_CHILD_MAX_VSIZE_VB
+            } else {
+                TRUC_MAX_VSIZE_VB
+            };
+            if estimated_vb > cap_vb {
+                let available = coin_selection.selected_amount();
+                return Err(CreateTxError::CoinSelection(InsufficientFunds {
+                    needed: available + Amount::from_sat(1),
+                    available,
+                }));
```
ValuedMammal: I'm worried this won't adequately convey the true cause of the error - that the vsize limit was exceeded. Probably the right way is to pass a max-weight parameter to a coin selector which selects inputs while the total weight is within the size limit or fails if the cap is exceeded before the target is funded.

yan-pi: I agree with the concern.

I talked with @oleonardolima too, and our current thinking is to keep this PR non-breaking. `InsufficientFunds` is not the ideal error, but adding `TrucSizeExceeded` would be a breaking API change

Longer term, I agree this should probably be modeled as a coin-selection constraint, not just a post-selection check

And AFAIK `bdk_tx` is still being built, that maybe would be better for this kind of policy-aware transaction construction

For now, I’d like to keep this PR as the minimal non-breaking guard because Second/Ark users need Rules 4/5 enforced now, and without this BDK can return PSBTs that bitcoind rejects

Happy to track the cleaner design as a follow-up

yan-pi:
rebased on top of 7982b6e8fe0b47984d0b148b325f8b4bd9e44481

yan-pi:
Since this moved to 4.0, I want to check if I understood the direction.

[coin-select#51](https://github.com/bitcoindevkit/coin-select/pull/51) now has `Target::max_weight` and `MaxWeightExceeded`. The current wallet `CoinSelectionAlgorithm` does not use `bdk_coin_select` or take a weight limit though, so bringing that here would mean changing the public trait and built-in selectors. [bdk-tx#47](https://github.com/bitcoindevkit/bdk-tx/pull/47) seems to be handling something similar in the new path.

@ValuedMammal is that what you had in mind for #493: change the current coin-selection API, or keep the final Rules 4/5 check here and leave max-weight-aware selection for `bdk_tx`?

I also noticed the cap isn't fixed: it's 10k without an unconfirmed ancestor and 1k when the selection has one, so I think we may need two selection attempts.

btw, I used AI while researching this, mostly to understand how the current coin-selection work fits together across the repos.

ValuedMammal:
@yan-pi I would say bring back the error you had originally designed for and keep the `TRUC_MAX_VSIZE_VB` validation logic the same.

yan-pi:
Ive restored the dedicated `CreateTxError::TrucSizeExceeded { cap_vb, actual_vb }` variant. I Think its ready for another review round, or ready to be merged.

110CodingP (commented):
@oleonardolima @ValuedMammal  we were planning to remove the `TxBuilder` in 4.0 right? so these changes should be done to `create_psbt` instead?

yan-pi:
@110CodingP i think `create_psbt` isn't ready. Gated on `bdk-tx` and `--cfg bdk_wallet_unstable`.

`TxBuilder` is still the stable API. It wasn't deprecated (the 3.2.0 marks are on the signer (#505))

I'd leave this PR as-is. Im gonna look foward to take the `create_psbt` / `bdk_tx` port as a follow-up.

110CodingP:
>i think `create_psbt` isn't ready. Gated on `bdk-tx` and `--cfg bdk_wallet_unstable`.
> 
> `TxBuilder` is still the stable API. It wasn't deprecated (the 3.2.0 marks are on the signer (#505))
> 
> I'd leave this PR as-is. Im gonna look foward to take the `create_psbt` / `bdk_tx` port as a follow-up.

But this PR is meant for 4.0 right? I mean it has a breaking change right, so would need a 4.0 ? I may be wrong but afaik `TxBuilder` is supposed to be removed by 4.0.

Issue #410, opened 2026-03-17
Improve CI persistence and MSRV testing, automate issue triage


notmandatory:
## Overview  

Improve BDK Wallet CI processes for persistence and MSRV testing, and add automated issue triage.

**Description**

This is a tracking issue to improve our automated CI (continuous integration) testing and issue triage. For the CI improvements we need GitHub actions that run on every PR to confirm that no persisted data incompatibilities were introduced. We already have Github actions that confirm the codebase compiles both on the stable rust version and our chosen MSRV (minimum supported rust version), but these tests are brittle and hard to maintain. And finally for issue triage it would be nice to have a GitHub action that helps close stale issues so project contributors have an accurate list of issues to be worked on.

**Expected Outcomes**  

The issues and sub-issues below should be completed.

- #234
  - Add CI process to verify data created with the prior release version can be loaded with a new version
  - Add CI process to verify data created with new version can be loaded with the prior release version
- Improve MSRV testing
  - #400 
  - #79
- https://github.com/bitcoindevkit/bdk_wallet/issues/386

**Resources**  

- [Github actions](https://docs.github.com/en/actions)
- [cargo-rbmt](https://github.com/rust-bitcoin/rust-bitcoin-maintainer-tools/tree/master/cargo-rbmt)
- [stale action](https://github.com/actions/stale)

**Skills Required**  

- hands on experience with git and GitHub
- basic rust programming skills
- basic sqlite DB skills
- be able to sign your git commits

**Mentor(s)**  

@notmandatory 

**Difficulty**  

Medium

**Competency Test (optional)**  

- Create a repo in your personal GitHub account
- Create and commit a simple rust "hello world" app with properly signed commits
- Add a simple Github Actions job to build and check the formatting of your app
- Add a feature branch to change your app in some way
- Create a PR from your feature branch

lambaaryan011:
Hello @notmandatory,  I’m Aryan Lamba, a 3rd-year B.Tech Computer Science student, and I’m very interested in working on this issue for Summer of Bitcoin 2026.

I’ve completed the competency test with signed commits, a Rust hello-world project, and a GitHub Actions workflow (build + fmt):
https://github.com/lambaaryan011/bdk-competency-aryan

I’ve started going through the issue and related sub-issues (#234, #400, #386, #79) and had a few initial thoughts:

[...]

PR #496, opened 2026-06-04
chore(ci): add signed commits check to integration

EliteCoder18:
### Description

Part of #410 

A new check-signed-commits job is added to the integration workflow. It runs only on pull requests and calls `ci/check-signed-commits.sh`, which iterates over every commit between the PR base and head, checks the signature status via `git log --format="%H %G?"`, and fails the job if any commit is unsigned (status = N). The check is implemented as a standalone bash script in ci/ consistent with the existing bash scripts, making it easy to run independently of CI.
  
### Notes to the reviewers

The signature check uses git's built-in %G? format specifier, which returns N specifically for commits with no signature at all distinct from E (signature present but key not in local keyring) or U(unknown key validity). This means the check correctly identifies unsigned commits without needing any GPG keys loaded in the CI environment.

The job only runs on pull_request events (guarded by if:` github.event_name == 'pull_request'`) so it doesn't affect push-triggered runs on protected branches.

### Changelog notice

ci: enforce GPG signed commits on pull requests

### Before submitting

### DIFF
--- a/.github/workflows/cont_integration.yml
+++ b/.github/workflows/cont_integration.yml
@@ -169,3 +169,16 @@ jobs:
         env: 
           RUSTDOCFLAGS: "-D warnings --cfg bdk_wallet_unstable"
         run: cargo doc --workspace --all-features --no-deps
+
+  check-signed-commits:
+    name: Check signed commits
+    runs-on: ubuntu-latest
+    if: github.event_name == 'pull_request'
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v6
+        with:
+          fetch-depth: 0
+          persist-credentials: false
+      - name: Verify all commits are GPG signed
+        run: ./ci/check-signed-commits.sh "${{ github.event.pull_request.base.sha }}" "${{ github.event.pull_request.head.sha }}"
--- a/ci/check-signed-commits.sh
+++ b/ci/check-signed-commits.sh
@@ -1,0 +1,26 @@
+#!/bin/bash
+
+set -euo pipefail
+
+BASE_SHA="${1:-}"
+HEAD_SHA="${2:-}"
+
+if [ -z "$BASE_SHA" ] || [ -z "$HEAD_SHA" ]; then
+    echo "Usage: $0 <base_sha> <head_sha>"
+    exit 1
+fi
+
+UNSIGNED=0
+while IFS=' ' read -r commit status; do
+    if [ "$status" = "N" ]; then
+        echo "Commit $commit is not GPG signed."
+        UNSIGNED=$((UNSIGNED + 1))
+    fi
+done < <(git log --format="%H %G?" "${BASE_SHA}..${HEAD_SHA}")
+
+if [ "$UNSIGNED" -gt 0 ]; then
+    echo "Error: $UNSIGNED commit(s) are not GPG signed. See CONTRIBUTING.md."
+    exit 1
+fi
+
+echo "All commits are GPG signed."


notmandatory (approved):
ACK eba7f046ce0999a902189066bc66a07d3eae2193

Will wait for team feedback before merging.

.github/workflows/cont_integration.yml
```diff
@@ -161,3 +161,16 @@ jobs:
...
+    runs-on: ubuntu-latest
+    if: github.event_name == 'pull_request'
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v6
+        with:
+          fetch-depth: 0
+          persist-credentials: false
+      - name: Verify all commits are GPG signed
+        run: ./ci/check-signed-commits.sh "${{ github.event.pull_request.base.sha }}" "${{ github.event.pull_request.head.sha }}"
```
oleonardolima, line 184: i'd rather have the bash call here instead of a new file, if it's only being used  here, such as:

```
"$(git log --pretty='format:%G?' -1 HEAD)" = "N"  ] && \
       echo "\nERROR: unsigned commit: BDK requires that commits be signed." || \
       true
```

oleonardolima: @EliteCoder18 you can see how i did here: https://github.com/oleonardolima/bdk_wallet/commit/974822093636014f94705eb64b7e89957d210f1c

oleonardolima: if you'd rather keep the simpler version of the file, it could be used in the justfile check command too.

EliteCoder18: Thanks for the suggestion! the justfile already uses a similar one-liner for local development, but intentionally with `|| true` so it acts as a warning rather than blocking contributors mid-flow.

For CI, I feel the goal is stricter enforcement across the entire PR range (`base_sha..head_sha`), not just the latest commit. An inline `HEAD` check could miss cases where only the most recent commit is signed while earlier commits are not.

That said, I like the idea of reusing the same logic in the justfile as well, and I'm happy to wire the script into the local check recipe if you think that would be beneficial.

notmandatory:
Also would be best to do a manual test of this in your cloned repo to confirm it works as expected.

EliteCoder18:
> Also would be best to do a manual test of this in your cloned repo to confirm it works as expected.

For testing, I added an unsigned commit to this PR and reran the workflow. The signed-commit check failed as expected, confirming that the action correctly detects unsigned commits in the PR history.

I'll revert the test commit after verification.

image

oleonardolima (approved):
ACK eba7f046ce0999a902189066bc66a07d3eae2193

It looks good, thanks for working on this! I can't trigger the CI for some reason, though.

notmandatory:
This looks ready I just need you to re-base it (if needed) on the latest `master` branch so it merges cleanly.

EliteCoder18:
> This looks ready I just need you to re-base it (if needed) on the latest `master` branch so it merges cleanly.

Done! I've rebased the branch onto the latest master and force-pushed the updated commits. It should now merge cleanly.

oleonardolima (approved):
utACK 21fd312ad183e3c7f9506507bfec1e3db329e4bc

.github/workflows/cont_integration.yml
ValuedMammal: > I'm happy to wire the script into the local check recipe if you think that would be beneficial.

I think it would be beneficial.

notmandatory (approved):
tACK 

Confirmed the script does what it's supposed to for unsigned and signed commits:

```shell
$ ci/check-signed-commits.sh a260a33d8d00bbd4b34c94fc73500c52e9eec0a4 c5e8e844e1b8215db0c0270defdf4c6b6bffc88d
Commit c5e8e844e1b8215db0c0270defdf4c6b6bffc88d is not GPG signed.
Error: 1 commit(s) are not GPG signed. See CONTRIBUTING.md.

$ ci/check-signed-commits.sh 58fe631177047ef1c49947ff453110f47138adcd 21fd312ad183e3c7f9506507bfec1e3db329e4bc
All commits are GPG signed.
```

ValuedMammal (approved):
ACK a2f61e2280eedd525533fbc7d6c0e3986648deae

PR #502, opened 2026-06-15
feat(wallet): Three-stage PSBT creation over bdk_tx::TxTemplate (PoC for discussion)

evanlinjin:
### Description

> **This is a proof-of-concept opened for discussion — not a merge-ready PR.** It explores an alternative design to #297 so we can compare API shapes side by side and decide on a direction *before* polishing either one. Plenty here (naming, the missing convenience façade, the upstream `bdk_tx` dependency) is deliberately unfinished; I'm interested in feedback on the *shape*, not the finish.

Both approaches add PSBT creation to `Wallet` and produce identical PSBTs; they differ in **where the complexity lives**.

To cover the real cases (manual coin selection, foreign inputs, assets, sweep, RBF, fee policy, version/locktime/anti-fee-sniping, ordering, global xpubs), a single `create_psbt(params)` entry point needs one very large params type — in #297, a ~30-method typestate builder. That shape carries three costs that compound over time:

1. **The wallet ends up re-declaring `bdk_tx`.** Version, locktime, anti-fee-sniping, ordering, shuffling, and per-input/fallback sequence already exist in `bdk_tx`. Surfacing them as wallet options means wrapping and maintaining each, so every `bdk_tx` change becomes a wallet change and the option list grows to mirror a dependency we should simply be *using*.
2. **Invalid or conflicting options only surface at the final call.** A single params type only sorts two things out when `create_psbt` finally runs: *conflicts* (an explicit `locktime` next to `anti_fee_sniping_height`, both setting `nLockTime`, precedence hidden in docs) and *invalid values* (a sequence violating an input's CSV requirement isn't rejected when set — it surfaces only at build time, failing the whole process). The staged flow applies operations in explicit order, and each one (e.g. `template.input_mut(op).set_sequence(..)`) validates and errors *at the call site*, leaving the rest intact. This shape also suits **frontend / interactive** use well: a UI can resolve candidates once, let the user adjust selection and outputs against a live, inspectable `CandidateSet`/`TxTemplate`, and surface each error inline at the control that caused it — rather than filling a whole form and getting one wholesale failure at submit.
3. **Nothing is reusable.** A single call returning a finished PSBT can't resolve a candidate set once and build several transactions, or inspect/adjust the selection before committing.

The direction this PoC explores: **the wallet should be a thin, composable layer over `bdk_tx`, not a façade that hides it.** PSBT creation is three genuinely distinct concerns — *what can I spend*, *what do I spend to hit this target*, *turn that into a PSBT* — so it's modeled as three stages instead of one call:

```rust
let coins    = wallet.candidates()?;                               // 1. resolve
let template = wallet.select(coins, SelectParams {                 // 2. select → TxTemplate
    recipients, fee_rate, ..Default::default()
})?;
let template = template.shuffle_outputs(&mut rng);                 //    shape (bdk_tx methods)
let (psbt, finalizer) = wallet.finish(template, FinishParams::default())?;  // 3. emit
```

This makes the three costs go away: each stage has a small, focused params struct; transaction shaping happens on the returned `bdk_tx::TxTemplate` using `bdk_tx`'s own methods (no re-wrapping, no drift); sweep and RBF become composable *axes* rather than special types (a sweep is `DrainAll` with no recipients; RBF is a field on the candidate params); and the intermediate values are real and inspectable.

Because the candidate set and the template are now real values the caller holds, several capabilities that the single-call design structurally **could not** expose fall out for free:

- **Input grouping** (`CandidateSet::regroup`) — group UTXOs by a key (e.g. address / script pubkey) so coin selection treats a group as one unit. This is privacy-relevant: it avoids spending *some but not all* UTXOs of an address. There was no way to express this through `PsbtParams`.
- **Post-resolution filtering** (`CandidateSet::filter`) — apply arbitrary predicates over the resolved inputs (by value, script, confirmation status, coinbase/maturity), beyond the fixed `CandidateParams` knobs.
- **Resolve-once, select-many** — reuse one resolved candidate set across multiple independent selections. This one isn't hypothetical: a large `bdk_wallet` user has already asked for exactly this, and the single-call design can't accommodate it without re-resolving the whole candidate set every time.
- **Foreign inputs go straight onto the set** — `CandidateSet::push_must_select` / `push_can_select` take pre-built `bdk_tx::Input`s, so `CandidateParams` stays purely about *deriving candidates from the wallet* rather than also being a pass-through for caller-supplied `bdk_tx` values.
- **RBF batching** — for a replacement, `CandidateSet::replaced_unspent()` exposes the replaced txs' still-unspent wallet outputs, so a caller can re-create those payments when batching several txs into one replacement.
- **`bdk_tx` escape hatch** — `CandidateSet::into_parts()` (yielding the `InputCandidates` and any `RbfParams`) and the returned `TxTemplate` let a caller drop down to `bdk_tx` primitives for anything the wallet doesn't wrap, instead of being limited to the options the wallet chose to surface.

These aren't features bolted on — they're a direct consequence of modeling the stages as values rather than hiding them behind one call.

If anything the staged design is *more* ergonomic and transparent: the single-call approach is one large params struct where the options' relationships are hard to follow, whereas each stage here scopes its options so what's valid where is obvious. The one thing it gives up is the single-call one-liner for the trivial "send X to Y" case — and that convenience can be layered *on top of* the composable stages later, whereas composability can't be retrofitted onto a sealed call.

Full API surface (`CandidateParams`/`SelectParams`/`FinishParams`, `SelectionStrategy`, `CandidateSet`, and the per-stage `CandidatesError`/`CreatePsbtError`) is in the diff and rustdoc. Examples: `examples/psbt.rs`, `examples/replace_by_fee.rs`.

### Notes to the reviewers

As a PoC, these are open for discussion rather than settled — they're the main things I'd want a decision on:

- **Privacy default.** `select` returns an *unshuffled* template; the caller must call `shuffle_outputs`. Deliberate and loudly documented, but it gives up #297's shuffle-by-default safety. **Should `finish` shuffle by default with an opt-out instead?**
- **No convenience façade.** Intentionally omitted — add a one-call helper now, or wait for demand?
- **Params are not `#[non_exhaustive]`,** so adding a field is breaking; kept this way to allow inline `..Default::default()` construction. Open to revisiting.
- **Determinism.** Candidate construction is RNG-free; randomness lives only in `select` (via `Selector::select_with_algorithm`).
- **Upstream dependency.** Requires `bdk_tx`'s `TxTemplate` (bitcoindevkit/bdk-tx#73) plus two small additions currently on the fork: a `single_random_draw` selection algorithm and `InputCandidates::push_must_select` / `push_can_select`. The branch points `bdk_tx` at that fork and **must be repointed to upstream before merge.**

### Changelog notice

- Added a three-stage PSBT-creation API to `Wallet`: `candidates`/`rbf_candidates`/`candidates_with`, `select`/`select_with_rng`, and `finish`, along with `CandidateParams`, `SelectParams`, `FinishParams`, `SelectionStrategy`, `CandidateSet`, and the `CandidatesError`/`CreatePsbtError` error types. `CandidateSet` supports `push_must_select`/`push_can_select` (foreign inputs), `filter`/`regroup`, `replaced_unspent` (RBF batching), and `into_parts` (`bdk_tx` escape hatch).

#### All Submissions:

#### New Features:

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -79,3 +79,9 @@ name = "esplora_blocking"
 
 [[example]]
 name = "bitcoind_rpc"
+
+[[example]]
+name = "psbt"
+
+[[example]]
+name = "replace_by_fee"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,7 +21,7 @@ unexpected_cfgs = { level = "warn", check-cfg = ['cfg(coverage_nightly)'] }
 
 [dependencies]
 bdk_chain = { version = "0.23.3", features = ["miniscript", "serde"], default-features = false }
-bdk_tx = { version = "0.2.0", default-features = false }
+bdk_tx = { git = "https://github.com/evanlinjin/bdk-tx", branch = "feature/tx-template", default-features = false }
 bitcoin = { version = "0.32.8", features = ["serde", "base64"], default-features = false }
 miniscript = { version = "12.3.5", features = ["serde"], default-features = false }
 rand_core = { version = "0.6.4" }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,6 +21,7 @@ unexpected_cfgs = { level = "warn", check-cfg = ['cfg(coverage_nightly)'] }
 
 [dependencies]
 bdk_chain = { version = "0.23.3", features = ["miniscript", "serde"], default-features = false }
+bdk_tx = { version = "0.2.0", default-features = false }
 bitcoin = { version = "0.32.8", features = ["serde", "base64"], default-features = false }
 miniscript = { version = "12.3.5", features = ["serde"], default-features = false }
 rand_core = { version = "0.6.4" }
@@ -33,9 +34,10 @@ bdk_file_store = { version = "0.22.0", optional = true }
 bip39 = { version = "2.2.2", optional = true }
 tempfile = { version = "3.26.0", optional = true }
 
+
 [features]
 default = ["std"]
-std = ["bitcoin/std", "bitcoin/rand-std", "miniscript/std", "bdk_chain/std"]
+std = ["bitcoin/std", "bitcoin/rand-std", "miniscript/std", "bdk_chain/std", "bdk_tx/std"]
 compiler = ["miniscript/compiler"]
 all-keys = ["keys-bip39"]
 keys-bip39 = ["bip39"]
--- a/examples/psbt.rs
+++ b/examples/psbt.rs
@@ -5,10 +5,10 @@ use std::str::FromStr;
 
 use bdk_chain::BlockId;
 use bdk_chain::ConfirmationBlockTime;
-use bdk_wallet::psbt::{PsbtParams, SelectionStrategy::*};
+use bdk_wallet::psbt::{FinishParams, SelectParams, SelectionStrategy::*};
 use bdk_wallet::test_utils::*;
 use bdk_wallet::{KeychainKind::External, Wallet};
-use bitcoin::{consensus, secp256k1::rand, Address, Amount, TxIn, TxOut};
+use bitcoin::{consensus, secp256k1::rand, transaction::Version, Address, Amount, TxIn, TxOut};
 use rand::Rng;
 
 // This example shows how to create a PSBT using BDK Wallet.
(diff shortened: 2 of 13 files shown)


notmandatory:
There are some good ideas in this PoC that are worth exploring but I don't think it should be considered as a replacement for #297. #297 has a similar API as the current TxBuilder which will make it easier to adopt by users, and since #297 has already gone through a good amount of review we can get it to users sooner. 

Based on user feedback we can consider improving the API in future releases but for now I think the most important thing is to give users an easy transition from TxBuilder to a `bdk-tx` based way to create transactions.

evanlinjin:
@notmandatory unfortunately it's the shape that can't cheaply be revisited. Whereas, correctness gaps can easily patched later. That being said, logically, it mostly reuses what was already done in #297.

I think it's important to get the shape right. This PR isn't merely "a few good ideas", it's fixing a problematic shape that already existed in the original `TxBuilder`.

notmandatory:
Is there a functional defect in #297 that requires the changes you're proposing or only an API "shape" improvement? 

I think we need to first solve the long overdue problem of replacing the old TxBuilder policy with the new miniscript planning feature. The API in #297:

1. gives us miniscript planning without drastically changing the API users are familiar with
2. has had plenty of design and code review time
3. can be delivered to users to use and evaluate in the 3.2 milestone

evanlinjin:
@notmandatory There is no functional defect with #297 - I'm just worried about the shape as it is inflexible and obscures functionality that `bdk_tx` already provides.

I agree with the general idea of the goals you have mentioned. However, I would like to provide some clarifications.

1. First off, the `TxBuilder` API is not the right shape so basing anything on that does not provide any value even if users are "used to it". That being said, #297 is still a new method with a new params struct which users are not used to. Additionally, "three calls vs one" is a documentation problem, not a burden for users.

2. This PR builds upon the foundation of #297. The majority of the transaction-construction logic is not changed - only repackaged for the staged API. Sure, the `psbt/params.rs` diff is large, but that's mostly removing the old builder.

3. I also agree it will be great to ship to users soon. However, if we merge #297 (and even this PR) now, we will be put in a position where we won't be able to iterate fast (refer to my comment [here](https://github.com/bitcoindevkit/bdk_wallet/pull/297#issuecomment-4810411011)). Additionally, greatly reworking the API shape to implement features later will be a hassle for the users imo. It's better to get the general API shape right now.

evanlinjin:
Closing in favor of bitcoindevkit/bdk-tx#84.

Rationale: https://github.com/bitcoindevkit/bdk_wallet/pull/297#issuecomment-4810411011

notmandatory:
@evanlinjin for our first release with `bdk-tx` I really want to focus on delivering incremental improvement, which #297 does. That doesn't mean there aren't good ideas in this draft PR, only that they need more time for review and comment to figure out if they're needed. If so then we can target them for a future breaking release and give the team plenty of time to review and discuss.

Issue #123, opened 2023-11-15
Do not include satisfaction in policy id calculation


afilini:
Currently every node in the `Policy` tree has an id which is calculated as a checksum on the string of the json-serialized object:

https://github.com/bitcoindevkit/bdk/blob/d6a0cf0795d9f8a1550e614e96b339a18fc537dc/crates/bdk/src/descriptor/policy.rs#L174-L178

The object includes details about whether the current wallet can satisfy the policy in some ways (aka `contribution`), and these are also included in the checksum. The same is true for what in a psbt is currently satisfied, which is the `satisfaction` field of `Policy`.

This causes the id of the nodes to change from wallet to wallet, depending on which private keys a wallet has or depending on which psbt you are looking at. Ideally if both wallets have the same public descriptor, they should also have the same id for the nodes, which makes sharing `policy_path`s much easier.

I would recommend changing the way the id is calculated, if you have any ideas let me know. I could also try working on it myself.

PR #505, opened 2026-06-23
Deprecate wallet-owned signing APIs

noahjoeris:
### Description

Partially addresses #70

Deprecates wallet-owned signing APIs, migrating examples, tests, and documentation toward caller-owned keys using `bitcoin::Psbt::sign` and `Wallet::sign_with_signers`.

Depends on:

- https://github.com/rust-bitcoin/rust-miniscript/pull/872
- https://github.com/rust-bitcoin/rust-miniscript/pull/867

### Notes to the reviewers

- I migrated examples, docs and tests to use `bitcoin::Psbt::sign` where possible. And used `Wallet::sign_with_signers` if we need SignOptions. 

### Changelog notice

```
- Deprecated `Wallet::{add_signer, set_keymap, set_keymaps, get_signers, sign, policies}`, `CreateParams::keymap`, `LoadParams::{keymap, extract_keys}`, and `TxBuilder::policy_path` in favor of caller-owned keys with `bitcoin::Psbt::sign` / `Wallet::sign_with_signers`, and `TxBuilder::set_condition`.
- Changed `FullyNodedExport::export_wallet` to not export secret keys anymore.
- Added `TxBuilder::set_condition` as replacement for `TxBuilder::policy_path`.
```

### Before submitting

### DIFF
--- a/examples/electrum.rs
+++ b/examples/electrum.rs
@@ -4,7 +4,10 @@ use bdk_wallet::Wallet;
 use bdk_wallet::bitcoin::Amount;
 use bdk_wallet::bitcoin::FeeRate;
 use bdk_wallet::bitcoin::Network;
+use bdk_wallet::bitcoin::secp256k1::Secp256k1;
 use bdk_wallet::chain::collections::HashSet;
+use bdk_wallet::descriptor::IntoWalletDescriptor;
+use bdk_wallet::miniscript::descriptor::KeyMapWrapper;
 use bdk_wallet::psbt::PsbtUtils;
 use bdk_wallet::rusqlite::Connection;
 use bdk_wallet::{KeychainKind, SignOptions};
@@ -23,16 +26,22 @@ const INTERNAL_DESC: &str = "wpkh(tprv8ZgxMBicQKsPdy6LMhUtFHAgpocR8GC6QmwMSFpZs7
 const ELECTRUM_URL: &str = "ssl://mempool.space:40002";
 
 fn main() -> Result<(), anyhow::Error> {
+    let secp = Secp256k1::new();
+    let (external_descriptor, mut external_keymap) =
+        EXTERNAL_DESC.into_wallet_descriptor(&secp, NETWORK.into())?;
+    let (internal_descriptor, internal_keymap) =
+        INTERNAL_DESC.into_wallet_descriptor(&secp, NETWORK.into())?;
+    external_keymap.extend(internal_keymap);
+    let signer = KeyMapWrapper::from(external_keymap);
     let mut db = Connection::open(DB_PATH)?;
     let wallet_opt = Wallet::load()
-        .descriptor(KeychainKind::External, Some(EXTERNAL_DESC))
-        .descriptor(KeychainKind::Internal, Some(INTERNAL_DESC))
-        .extract_keys()
+        .descriptor(KeychainKind::External, Some(external_descriptor.clone()))
+        .descriptor(KeychainKind::Internal, Some(internal_descriptor.clone()))
         .check_network(NETWORK)
         .load_wallet(&mut db)?;
     let mut wallet = match wallet_opt {
         Some(wallet) => wallet,
-        None => Wallet::create(EXTERNAL_DESC, INTERNAL_DESC)
+        None => Wallet::create(external_descriptor, internal_descriptor)
             .network(NETWORK)
             .create_wallet(&mut db)?,
     };
@@ -89,7 +98,9 @@ fn main() -> Result<(), anyhow::Error> {
     tx_builder.fee_rate(target_fee_rate);
 
     let mut psbt = tx_builder.finish()?;
-    let finalized = wallet.sign(&mut psbt, SignOptions::default())?;
+    psbt.sign(&signer, wallet.secp_ctx())
+        .map_err(|(_, e)| anyhow::anyhow!("{e:?}"))?;
+    let finalized = wallet.finalize_psbt(&mut psbt, SignOptions::default())?;
     assert!(finalized);
     let original_fee = psbt.fee_amount().unwrap();
     let tx_feerate = psbt.fee_rate().unwrap();
@@ -124,7 +135,10 @@ fn main() -> Result<(), anyhow::Error> {
     let mut builder = wallet.build_fee_bump(txid).expect("failed to bump tx");
     builder.fee_rate(feerate);
     let mut bumped_psbt = builder.finish().unwrap();
-    let finalize_btx = wallet.sign(&mut bumped_psbt, SignOptions::default())?;
+    bumped_psbt
+        .sign(&signer, wallet.secp_ctx())
+        .map_err(|(_, e)| anyhow::anyhow!("{e:?}"))?;
+    let finalize_btx = wallet.finalize_psbt(&mut bumped_psbt, SignOptions::default())?;
     assert!(finalize_btx);
     let new_fee = bumped_psbt.fee_amount().unwrap();
     let bumped_tx = bumped_psbt.extract_tx()?;
--- a/src/wallet/export.rs
+++ b/src/wallet/export.rs
@@ -50,9 +50,9 @@
 //!     "wpkh([c258d2e4/84h/1h/0h]tpubDD3ynpHgJQW8VvWRzQ5WFDCrs4jqVFGHB3vLC3r49XHJSqP8bHKdK4AriuUKLccK68zfzowx7YhmDN8SiSkgCDENUFx9qVw65YyqM78vyVe/0/*)",
 //!     "wpkh([c258d2e4/84h/1h/0h]tpubDD3ynpHgJQW8VvWRzQ5WFDCrs4jqVFGHB3vLC3r49XHJSqP8bHKdK4AriuUKLccK68zfzowx7YhmDN8SiSkgCDENUFx9qVw65YyqM78vyVe/1/*)",
 //! )
-//! .network(Network::Testnet)
-//! .create_wallet_no_persist()?;
-//! let export = FullyNodedExport::export_wallet(&wallet, "exported wallet", true).unwrap();
+//!     .network(Network::Testnet)
+//!     .create_wallet_no_persist()?;
+//! let export = FullyNodedExport::export_wallet(&wallet, "exported wallet", true)?;
 //!
 //! println!("Exported: {}", export.to_string());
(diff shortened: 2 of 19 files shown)


examples/electrum.rs
```diff
@@ -89,7 +98,10 @@ fn main() -> Result<(), anyhow::Error> {
     tx_builder.fee_rate(target_fee_rate);
 
     let mut psbt = tx_builder.finish()?;
-    let finalized = wallet.sign(&mut psbt, SignOptions::default())?;
+    wallet
+        .sign_psbt(&mut psbt, &signer)
+        .map_err(|(_, e)| anyhow::anyhow!("{e:?}"))?;
```
oleonardolima: question: this is something remaining from the previous commits, right ? it should use the `psbt.sign` directly.

_NOTE_: this also applies to the other examples and updated docs.

tests/psbt.rs
```diff
@@ -221,3 +258,81 @@ fn test_psbt_multiple_internalkey_signers() {
...
+    let (wallet, _) = get_funded_wallet(desc, change_desc);
+
+    let xpriv = Xpriv::from_str(
+        "tprv8ZgxMBicQKsPdy6LMhUtFHAgpocR8GC6QmwMSFpZs7h6Eziw3SpThFfczTDh5rW2krkqffa11UpX3XkeTTB2FvzZKWXqPY54Y6Rq4AQ5R8L",
+    )
+    .unwrap();
+
+    let signing_keys = wallet
+        .sign_psbt(&mut psbt, &xpriv)
+        .expect("sign_psbt should succeed with the correct xpriv");
```
oleonardolima: these tests are also using the API from previous (removed) commits.

oleonardolima (commented):
overall it's looking good, it's best if you fix the first commit and let the last one just adding the deprecation notice. let's us know when it's ready for final review.

noahjoeris:
Thanks for spotting this. It seems I added the change to the last commit and not the first, my bad. I cleaned up the commits.

oleonardolima (commented):
utACK c8109380fb6278dd35039e8b1de54ca59e9c9beb

noahjoeris:
I still need to look into `Wallet::policies` and `Wallet::create_tx` and what to do with them. They also use `self.signers`

noahjoeris:
For `Wallet::policies()`, I suggest deprecating it and adding something like `Wallet::policies_with_signers(keychain, signers)`.

One tricky part is `TxBuilder::finish() -> Wallet::create_tx()`. It currently extracts policies with the wallet signers. 

A possible migration path is adding `TxBuilder::policy(policy, keychain)`. Then callers that use `policies_with_signers(...)` can pass the resulting policy into the builder. Existing callers can keep the current fallback behavior of using wallet-owned signers during the deprecation window.

Downside: policy-path users get one more builder call, and `Policy` becomes more directly part of the API. Maybe that's acceptable with our intention of slowly moving away from TxBuilder.

vera945:
Neither `policies` nor `create_tx` needs deprecation. Here's the reasoning:

`create_tx` calls `extract_policy` only to invoke `requires_path()` and `get_condition()`. Both methods operate entirely on the descriptor's structural tree — `SatisfiableItem` — and never read the contribution field, which is the only output that depends on the signers container. So passing `SignersContainer::default()` is valid without changing the behavior. This is verifiable by creating a wallet from xpub descriptors and calling `build_tx` normally.

`policies` similarly behaves this way for watch-only wallets. The contribution field will be `Satisfaction::None` for all leaves, which correctly describes a wallet that holds no signing keys. If a caller wants to evaluate what their external signing keys can contribute, the right answer is `wallet.public_descriptor(keychain).extract_policy(their_signers, ...)`, which is already the pattern #505 is establishing.

noahjoeris:
No, passing empty signers won't work. It can create different `Policy` IDs, and `policy_path` relies on those IDs.

A path built from `Wallet::policies()` can fail when `create_tx()` re-extracts with empty signers, because `get_condition()` looks up selections by `self.id`.

noahjoeris:
So after some discussion with @thunderbiscuit and @notmandatory we concluded that we cannot remove the signers fields unless we update TxBuilder to accept a spending policy to be passed in. If we don't want to update it (from what I heard) then the wallet signer removal is coupled to the TxBuilder removal. And it might make sense to deprecate both at the same time once we integrated `create_psbt`.

examples/electrum.rs
```diff
@@ -89,7 +98,9 @@ fn main() -> Result<(), anyhow::Error> {
     tx_builder.fee_rate(target_fee_rate);
 
     let mut psbt = tx_builder.finish()?;
-    let finalized = wallet.sign(&mut psbt, SignOptions::default())?;
+    psbt.sign(&signer, wallet.secp_ctx())
```
ItoroD, line 101: @noahjoeris Just curious, is there any reason we call secp from the wallet `wallet.secp_ctx()` instead of just creating one? Does wallet add anything special to Secp256k1<c> ?

noahjoeris: We simply reuse it to avoid initializing another one. nothing special is added, creating and using your own context is equivalent.

ValuedMammal:
> So after some discussion with @thunderbiscuit and @notmandatory we concluded that we cannot remove the signers fields unless we update TxBuilder to accept a spending policy to be passed in. If we don't want to update it (from what I heard) then the wallet signer removal is coupled to the TxBuilder removal. And it might make sense to deprecate both at the same time once we integrated `create_psbt`.

That sounds extreme. The whole policy path ceremony in `create_tx` exists to come up with the ["requirements"](https://github.com/bitcoindevkit/bdk_wallet/blob/b01f2f61ca9a1625131b914f080be77c86c17e65/src/wallet/mod.rs#L1330) which is just a `Condition` struct with two optional fields, so what you could do is deprecate `TxBuilder::policy_path` and provide a `set_condition` method to pass the `Condition` directly and that will forever decouple the policy from TxBuilder. `policy_path` is broken anyway because of the unstable policy id as you mention (#123).

src/wallet/export.rs
```diff
@@ -194,19 +204,18 @@ impl FullyNodedExport {
     ///
     /// If the database is empty or `include_blockheight` is false, the `blockheight` field
     /// returned will be `0`.
-    pub fn export_wallet(
+    pub fn export_wallet_with_keymaps(
         wallet: &Wallet,
+        external_keymap: &KeyMap,
+        internal_keymap: &KeyMap,
```
ValuedMammal: 🔧 Keymaps can hold multiple entries, so `internal_keymap` appears like an extra parameter here.

```rust
pub fn export_wallet_with_keymap(
    wallet: &Wallet,
    key_map: &KeyMap,
    label: &str,
    include_blockheight: bool,
) -> Result<Self, &'static str> {
```

src/wallet/export.rs
```diff
@@ -244,6 +248,30 @@ impl FullyNodedExport {
         Ok(export)
     }
 
+    /// Export a wallet to FullyNoded format using keys stored in the wallet.
+    #[deprecated(
+        since = "3.2.0",
+        note = "pass external and internal KeyMaps explicitly; use FullyNodedExport::export_wallet_with_keymaps instead."
+    )]
+    #[allow(deprecated)]
+    pub fn export_wallet(
```
ValuedMammal: What's the reason for deprecating `export_wallet`? We can still create a `FullyNodedExport` from a wallet without keymaps.

ValuedMammal (changes requested):

src/wallet/export.rs
noahjoeris: Yeah I didn't think about this. Makes sense. Thanks!

src/wallet/tx_builder.rs
```diff
@@ -191,8 +193,95 @@ impl<'a, Cs> TxBuilder<'a, Cs> {
...
+    /// // Inputs of the resulting transaction will carry nSequence = 6.
+    /// let mut builder = wallet.build_tx();
+    /// builder.set_condition(condition);
+    ///
+    /// # Ok::<(), anyhow::Error>(())
+    /// ```
+    pub fn set_condition(&mut self, condition: Condition) -> &mut Self {
+        self.params.condition = Some(condition);
+        self
+    }
```
ValuedMammal, line 278: One downside is you now have to set the expected Condition even when the spending policy doesn't require an explicit path.

ValuedMammal:
On second thought I think it should be merged to master and backported to 3.x, so we don't lose track of the tests migration ~~and export feature~~.

src/wallet/tx_builder.rs
ValuedMammal: For example if your policy is `and(pk(A),older(6))`, then the "requirement" consists of a single `csv: Some(Sequence(6))` condition. In the old code this is set automatically by the descriptor and no `policy_path` is required because there's only 1 spending path. Once `TxBuilder::policy_path` is gone, you'll have to either use `set_exact_sequence` or `set_condition` to ensure the input has the appropriate Sequence. It's not a problem, just something to be aware of.

@noahjoeris

ValuedMammal:
Do you want to leave the `export_with_keymap` feature for a future discussion and treat the wallet export as a watch-only format for now?

src/wallet/tx_builder.rs
noahjoeris: Okay. Probably we need to add some error for this when no condition is provided. once we remove the signers/deprecated things.

ValuedMammal (approved):
ACK 515bee8797d7bfd9708eb87afe5fc78f8f073adf; ran the tests locally

110CodingP (approved):
ACK [73c6736](https://github.com/bitcoindevkit/bdk_wallet/pull/505/commits/73c673693a1845a4b9c2aa25440b92950e4c72ac)

110CodingP (approved):
ACK 73c6736 
(did twice by mistake :sweat_smile: )

oleonardolima (commented):
utACK 73c673693a1845a4b9c2aa25440b92950e4c72ac

ValuedMammal:
reACK 3a3cd6f9c43e688b0a777b152ce4d4a1064c2c32

110CodingP:
ACK 3a3cd6f

110CodingP (approved):

ValuedMammal:
Next steps are to backport to release/3.x, and proceed with signer refactoring and pre-work for #524.

notmandatory:
I moved this to the 4.0 release since it was merged to `master` branch.

PR #507, opened 2026-06-27
fix(wallet): map keychain in mark_used and unmark_used

evanlinjin:
### Description

The bug that is fixed: `Wallet::mark_used` and `Wallet::unmark_used` passed the caller-supplied `KeychainKind` straight through to the underlying index. For single-descriptor wallets, this meant marking or unmarking an index on `KeychainKind::Internal` was a silent no-op.

### Changelog notice

```rust
- Fixed `Wallet::mark_used` and `Wallet::unmark_used` being a no-op on single-descriptor wallets when passed `KeychainKind::Internal`.
```

### Before submitting

~- [ ] This PR breaks the existing API~

### DIFF
--- a/src/wallet/mod.rs
+++ b/src/wallet/mod.rs
@@ -729,7 +729,9 @@ impl Wallet {
     ///
     /// Returns whether the given index was present and then removed from the unused set.
     pub fn mark_used(&mut self, keychain: KeychainKind, index: u32) -> bool {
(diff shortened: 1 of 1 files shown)


noahjoeris (approved):
ACK eb5f5b5f9e4eaa5532eb8b5a32df6b3a8869eaa0

PR #516, opened 2026-07-18
feat: Add Wallet::create_psbt

ValuedMammal:
### Description

This PR introduces `Wallet::create_psbt` and `Wallet::replace_by_fee` - a new PSBT construction API built on top of [`bdk-tx`](https://github.com/bitcoindevkit/bdk-tx). This is a second attempt at integrating the planning module that started with #297. The scope is mostly the same and incorporates a number of improvements noted on #297. Notably the entire surface area is gated behind two compile flags.

flags:
- the `bdk-tx` Cargo feature
- the `--cfg bdk_wallet_unstable` rustc flag

Enabling `bdk-tx` without also passing `bdk_wallet_unstable` is a compile error with a clear diagnostic. This allows us to ship the API for early adopters while reserving the right to break it in a minor release.

Users can opt in by adding the following to your project's `.cargo/config.toml`:

```toml
[build]
rustflags = ["--cfg", "bdk_wallet_unstable"]
```

and enable the feature in `Cargo.toml`:

```toml
bdk_wallet = { version = "...", features = ["bdk-tx"] }
```

#### New public API

| Item | Notes |
|------|-------|
| `Wallet::create_psbt` | Build a new PSBT with coin selection |
| `Wallet::create_psbt_with_rng` | Same with caller-supplied entropy |
| `Wallet::replace_by_fee` | Construct an RBF replacement for one or more transactions |
| `Wallet::replace_by_fee_with_rng` | Same with caller-supplied entropy |
| `PsbtParams<C>` | Builder for PSBT construction; `C` is `CreateTx` or `ReplaceTx`/`Rbf` |
| `SelectionStrategy` | Enum: `All`, `SingleRandomDraw`, `LowestFee`, `Custom` |
| `CreatePsbtError` | Error type for `create_psbt` |
| `ReplaceByFeeError` | Error type for `replace_by_fee` |
| `bdk_tx` re-export | For callers who need direct access to `bdk_tx` |

`TxOrdering` is also generalized to `TxOrdering<In, Out>` to accommodate
`bdk-tx`'s typed input/output types. `TxOrdering<In, Out>` adds default type parameters
(`TxIn`, `TxOut`) that are identical to the previous concrete type, so remains
[semver-compatible](https://doc.rust-lang.org/cargo/reference/semver.html#generic-generalize-identical).
Existing callers that don't use the `Custom` variant are unaffected.

#### Differences from #297

- **Non-essential API surface removed** - Some wallet methods present in #297 were trimmed; the public API is narrower and more focused.
- **`bdk_tx` re-export** - `bdk_wallet::bdk_tx` is re-exported so callers who need `bdk-tx` types directly don't need a separate `bdk_tx` dependency.
- **`SelectionStrategy::All` replaces `drain_wallet`** - modeling "sweep everything" as a strategy is cleaner than a `PsbtParams` option.
- **`SelectionStrategy::Custom`** - users can supply their own coin-selection algorithm as a closure without having to implement a trait.
- **`SingleRandomDraw` correctness fix** - shuffling now only runs when `SingleRandomDraw` is active; the extra `shuffle_slice` calls are removed.
- **`bdk_wallet_unstable` cfg gate + `bdk-tx` Cargo feature** - the full `create_psbt` / `replace_by_fee` surface is hidden unless explicitly opted into.

### Changelog notice

```
### Added
- feat(wallet): add unstable `Wallet::create_psbt` and `Wallet::replace_by_fee`
  behind the `bdk-tx` cargo feature and the `--cfg bdk_wallet_unstable` rustc
  flag. These APIs are explicitly **unstable**: breaking changes may land in
  minor releases without a semver bump. To opt in, enable the `bdk-tx` feature
  **and** pass `--cfg bdk_wallet_unstable` to rustc.
```

### Before submitting

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -82,3 +82,11 @@ name = "esplora_blocking"
 [[example]]
 name = "bitcoind_rpc"
 
+[[example]]
+name = "psbt"
+required-features = ["bdk-tx"]
+
+[[example]]
+name = "replace_by_fee"
+required-features = ["bdk-tx"]
+
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,13 +14,14 @@ rust-version = "1.85.0"
 
 [package.metadata.docs.rs]
 all-features = true
-rustdoc-args = ["--cfg", "docsrs"]
+rustdoc-args = ["--cfg", "docsrs", "--cfg", "bdk_wallet_unstable"]
 
 [lints.rust]
-unexpected_cfgs = { level = "warn", check-cfg = ['cfg(coverage_nightly)'] }
+unexpected_cfgs = { level = "warn", check-cfg = ['cfg(coverage_nightly)', 'cfg(bdk_wallet_unstable)'] }
 
 [dependencies]
 bdk_chain = { version = "0.23.3", features = ["miniscript", "serde"], default-features = false }
+bdk_tx = { version = "0.2.0", default-features = false, optional = true }
 bitcoin = { version = "0.32.8", features = ["serde", "base64"], default-features = false }
 miniscript = { version = "12.3.5", features = ["serde"], default-features = false }
 rand_core = { version = "0.6.4" }
@@ -33,14 +34,16 @@ bdk_file_store = { version = "0.22.0", optional = true }
 bip39 = { version = "2.2.2", optional = true }
 tempfile = { version = "3.26.0", optional = true }
 
+
 [features]
 default = ["std"]
-std = ["bitcoin/std", "bitcoin/rand-std", "miniscript/std", "bdk_chain/std"]
+std = ["bitcoin/std", "bitcoin/rand-std", "miniscript/std", "bdk_chain/std", "bdk_tx?/std"]
 compiler = ["miniscript/compiler"]
 all-keys = ["keys-bip39"]
 keys-bip39 = ["bip39"]
 rusqlite = ["bdk_chain/rusqlite"]
 file_store = ["bdk_file_store"]
+bdk-tx = ["dep:bdk_tx"]
 test-utils = ["std", "anyhow", "tempfile"]
 
 [dev-dependencies]
@@ -49,6 +52,7 @@ assert_matches = "1.5.0"
 bdk_bitcoind_rpc = { version = "0.22.0" }
 bdk_electrum = { version = "0.23.2" }
 bdk_esplora = { version = "0.22.1", features = ["async-https", "blocking-https", "tokio"] }
+bdk_tx = { version = "0.2.0" }
 bdk_wallet = { path = ".", features = ["rusqlite", "file_store", "test-utils"] }
 clap = { version = "4.5.60", features = ["derive", "env"] }
 ctrlc = "3.5.2"
@@ -77,3 +81,4 @@ name = "esplora_blocking"
 
 [[example]]
 name = "bitcoind_rpc"
+
--- a/justfile
+++ b/justfile
@@ -7,6 +7,9 @@ alias t := test
 alias p := pre-push
 alias d := doc
 
+# All features not including the unstable ones
+FEATURES := "std,compiler,all-keys,rusqlite,file_store,test-utils"
+
 _default:
    @just --list
 
@@ -14,11 +17,22 @@ _default:
 build:
    cargo build
 
-# Check code: formatting, compilation, linting, and commit signature
-check:
+# Check formatting, compilation, linting
+_check:
    cargo +nightly fmt --all -- --check
-   cargo check --all-features --all-targets
-   cargo clippy --all-features --all-targets -- -D warnings
+   RUSTFLAGS="" cargo check --all-targets --no-default-features --features miniscript/no-std,bdk_chain/hashbrown
+   RUSTFLAGS="" cargo check --all-targets --features {{FEATURES}}
+   RUSTFLAGS="-D warnings" cargo clippy --all-targets --features {{FEATURES}}
+
+# Check formatting, compilation, linting of the unstable API surface
+_check-unstable:
+   cargo +nightly fmt --all -- --check
+   RUSTFLAGS="--cfg bdk_wallet_unstable" cargo check --all-targets --no-default-features --features miniscript/no-std,bdk_chain/hashbrown
+   RUSTFLAGS="--cfg bdk_wallet_unstable" cargo check --all-targets --all-features
+   RUSTFLAGS="--cfg bdk_wallet_unstable -D warnings" cargo clippy --all-targets --all-features
+
+# Check formatting, compilation, linting, and commit signature
+check: _check _check-unstable
    @[ "$(git log --pretty='format:%G?' -1 HEAD)" = "N" ] && \
        echo "\n⚠️  Unsigned commit: BDK requires that commits be signed." || \
        true
@@ -27,13 +41,20 @@ check:
 fmt:
    cargo +nightly fmt
 
+# Run tests on the stable API surface
+_test:
+   RUSTFLAGS="" cargo test --workspace --features {{FEATURES}}
+
+# Run tests on the unstable API surface (with --cfg bdk_wallet_unstable)
+_test-unstable:
+   RUSTDOCFLAGS="--cfg bdk_wallet_unstable" RUSTFLAGS="--cfg bdk_wallet_unstable" cargo test --workspace --all-features
+
 # Run all tests on the workspace with all features
-test:
-   cargo test --all-features
+test: _test _test-unstable
 
 # Check docs on the workspace
 doc:
-   RUSTDOCFLAGS='-D warnings' cargo doc --workspace --all-features --no-deps
+   RUSTDOCFLAGS="-D warnings --cfg bdk_wallet_unstable" cargo doc --workspace --all-features --no-deps
 
 # Run pre-push suite: format, check, and test
 pre-push: fmt check test doc
--- a/src/psbt/params.rs
+++ b/src/psbt/params.rs
@@ -1,0 +1,902 @@
+//! Parameters for creating a PSBT.
+
+use alloc::sync::Arc;
+use alloc::vec::Vec;
+use core::fmt;
+
+use bdk_chain::{BlockId, CanonicalizationParams, ConfirmationBlockTime, FullTxOut, TxGraph};
+use bdk_tx::{ChangeScript, Input, Output, Selector, SelectorError};
+use bitcoin::{
+    absolute, transaction::Version, Amount, FeeRate, OutPoint, ScriptBuf, Sequence, Transaction,
+    Txid,
+};
+use miniscript::plan::Assets;
+
+use crate::collections::{HashMap, HashSet};
+use crate::TxOrdering;
+
+/// Marker type representing the PSBT creation state.
+#[derive(Debug)]
+pub struct CreateTx;
+
+/// Marker type representing the Replace-By-Fee (RBF) state.
+#[derive(Debug)]
+pub struct ReplaceTx;
+
+/// Parameters to create a PSBT.
+// TODO: Can we derive `Clone` for this?
+#[derive(Debug)]
+pub struct PsbtParams<C> {
+    /// Set of selected UTXO outpoints, `HashSet` ensures uniqueness
+    pub(crate) set: HashSet<OutPoint>,
+    /// Ordered list of manually-selected spends.
+    pub(crate) must_spend: Vec<MustSpend>,
+    /// List of recipient script/amount pairs.
+    pub(crate) recipients: Vec<(ScriptBuf, Amount)>,
+    /// Optional script or descriptor designated for change.
+    pub(crate) change_script: Option<ChangeScript>,
+    /// Optional assets for creating a spend plan.
+    pub(crate) assets: Option<Assets>,
+    /// Target fee rate.
+    pub(crate) fee_rate: FeeRate,
+    /// Coin selection strategy to use.
+    pub(crate) coin_selection: SelectionStrategy,
+    /// Parameters for transaction canonicalization.
+    pub(crate) canonical_params: CanonicalizationParams,
+    /// UTXO filtering function.
+    pub(crate) utxo_filter: UtxoFilter,
+    /// Optional height for evaluating coinbase maturity.
+    pub(crate) maturity_height: Option<u32>,
+    /// Only allow spending UTXOs which are selected manually.
+    pub(crate) manually_selected_only: bool,
+    /// Optional transaction [`Version`].
+    pub(crate) version: Option<Version>,
+    /// Minimum transaction locktime — a floor on the resulting `tx.lock_time`.
+    pub(crate) min_locktime: Option<absolute::LockTime>,
+    /// Optional height for BIP326 anti-fee sniping.
+    pub(crate) anti_fee_sniping: Option<absolute::Height>,
+    /// Ordering of the transaction's inputs and outputs.
+    pub(crate) ordering: TxOrdering<Input, Output>,
+    /// Only set the [`witness_utxo`](bitcoin::psbt::Input::witness_utxo) in PSBT inputs. This
+    /// allows opting out of setting the
+    /// [`non_witness_utxo`](bitcoin::psbt::Input::non_witness_utxo).
+    pub(crate) only_witness_utxo: bool,
+    /// Whether to try filling in the PSBT global xpubs from the wallet's descriptors.
+    pub(crate) add_global_xpubs: bool,
+    /// Set of txids being replaced if this is a RBF transaction.
+    pub(crate) replace: HashSet<Txid>,
+    /// Per-input sequence overrides keyed by outpoint.
+    ///
+    /// Only applies to inputs added via [`PsbtParams::add_utxos`]. Takes precedence over
+    /// [`fallback_sequence`](Self::fallback_sequence).
+    pub(crate) sequence_overrides: HashMap<OutPoint, Sequence>,
+    /// Fallback sequence applied to wallet-managed inputs that have no per-input override and
+    /// no CSV-derived sequence requirement.
+    pub(crate) fallback_sequence: Option<Sequence>,
+    /// The context in which the params are used.
+    pub(crate) marker: core::marker::PhantomData<C>,
+}
+
+impl Default for PsbtParams<CreateTx> {
+    fn default() -> Self {
+        Self {
+            set: Default::default(),
+            must_spend: Default::default(),
+            assets: Default::default(),
+            recipients: Default::default(),
+            change_script: Default::default(),
+            fee_rate: FeeRate::BROADCAST_MIN,
+            coin_selection: Default::default(),
+            canonical_params: Default::default(),
+            utxo_filter: Default::default(),
+            maturity_height: Default::default(),
+            manually_selected_only: Default::default(),
+            version: Default::default(),
+            min_locktime: Default::default(),
+            anti_fee_sniping: Default::default(),
+            ordering: Default::default(),
+            only_witness_utxo: Default::default(),
+            add_global_xpubs: Default::default(),
+            replace: Default::default(),
+            sequence_overrides: Default::default(),
+            fallback_sequence: Default::default(),
+            marker: core::marker::PhantomData,
+        }
+    }
+}
+
+impl PsbtParams<CreateTx> {
+    /// Create a new [`PsbtParams`].
+    pub fn new() -> Self {
+        Self::default()
+    }
+
+    /// Add UTXOs by outpoint to fund the transaction.
+    ///
+    /// A single outpoint may appear at most once in the list of UTXOs to spend. The caller is
+    /// responsible for ensuring that items of `outpoints` correspond to outputs of previous
+    /// transactions and are currently unspent.
+    ///
+    /// If an outpoint doesn't correspond to an indexed script pubkey, an [`UnknownUtxo`]
+    /// error will occur. See [`Wallet::create_psbt`] for more.
+    ///
+    /// To add a UTXO that did not originate from this wallet (i.e. a "foreign" UTXO), see
+    /// [`PsbtParams::add_planned_input`].
+    ///
+    /// [`UnknownUtxo`]: crate::wallet::error::CreatePsbtError::UnknownUtxo
+    /// [`Wallet::create_psbt`]: crate::Wallet::create_psbt
+    pub fn add_utxos(&mut self, outpoints: &[OutPoint]) -> &mut Self {
+        for outpoint in outpoints.iter().copied() {
+            if self.set.insert(outpoint) {
+                self.must_spend.push(MustSpend::Utxo(outpoint));
+            }
+        }
+        self
+    }
+
+    /// Add a planned input.
+    ///
+    /// This can be used to add inputs that come with a [`Plan`] or [`psbt::Input`] provided.
+    /// See [`Input`] for more on how to create inputs manually. Be aware that creating inputs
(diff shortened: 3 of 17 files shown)


src/wallet/error.rs
```diff
@@ -365,3 +367,119 @@ impl fmt::Display for BuildFeeBumpError {
...
+#[non_exhaustive]
+pub enum ReplaceByFeeError {
+    /// There was a problem creating the PSBT
+    CreatePsbt(CreatePsbtError),
+    /// Failed to compute the fee of an original transaction
+    PreviousFee(bdk_chain::tx_graph::CalculateFeeError),
+    /// Original transaction could not be found
+    MissingTransaction(Txid),
+    /// One of the transactions to be replaced is already confirmed
+    TransactionConfirmed(Txid),
```
evanlinjin, line 439: I don't think this error variant earns it's keep.

* Most callers would want to check whether the transaction is still unconfirmed before trying to replace it. A caller that doesn't do that obviously wants to create a transaction that conflicts with a confirmed one.

* The guard is partial. It only catches "already confirmed at build time". The original transaction can be confirmed before the replacement is ready to broadcast.

src/wallet/mod.rs
```diff
@@ -2895,6 +2934,774 @@ impl Wallet {
...
+                .get(&txid)
+                .is_some_and(|chain_position| chain_position.is_confirmed())
+            {
+                return Err(ReplaceByFeeError::TransactionConfirmed(txid));
+            }
+        }
+
+        // For each txid being replaced, verify that at least one of its original inputs
+        // remains in the selected set. A replacement must conflict with every transaction it
+        // replaces — two transactions cannot spend the same UTXO.
```
evanlinjin, line 3463:
```suggestion
        // A replacement must conflict (double-spend) with each tx it replaces - so require at least
        // one of each original's inputs to remain selected.
```

wdyt?

src/wallet/mod.rs
```diff
@@ -2895,6 +2934,774 @@ impl Wallet {
...
+            .collect::<Result<_, _>>()?;
+
+        // Sum fees from all descendants known to the tx graph. This assumes every
+        // descendant is currently in the mempool, which could slightly overestimate
+        // the fee floor if a descendant was evicted or never relayed.
+        let descendant_fee: Amount = descendants
+            .iter()
+            .filter_map(|&txid| {
+                let tx = self.tx_graph.graph().get_tx(txid)?;
+                self.calculate_fee(&tx).ok()
```
evanlinjin, line 3574: Do you think we should return an error if an original's descendant cannot have it's fee calculated? The current implementation has a risk of undershooting the RBF fee floor.

src/wallet/mod.rs
```diff
@@ -2895,6 +2934,774 @@ impl Wallet {
...
+            })
+            .collect::<Result<_, _>>()?;
+
+        // Sum fees from all descendants known to the tx graph. This assumes every
+        // descendant is currently in the mempool, which could slightly overestimate
+        // the fee floor if a descendant was evicted or never relayed.
+        let descendant_fee: Amount = descendants
+            .iter()
+            .filter_map(|&txid| {
+                let tx = self.tx_graph.graph().get_tx(txid)?;
```
evanlinjin, line 3573:
```suggestion
                let tx = self.tx_graph.graph().get_tx(txid).expect("tx must exist as they were obtained via the TxGraph");
```

We obtained `descendants` directly from `TxGraph`. So they must exist.

Cargo.toml
```diff
@@ -21,6 +21,7 @@ unexpected_cfgs = { level = "warn", check-cfg = ['cfg(coverage_nightly)'] }
 
 [dependencies]
 bdk_chain = { version = "0.23.3", features = ["miniscript", "serde"], default-features = false }
+bdk_tx = { version = "0.2.0", default-features = false }
```
oleonardolima: nit: shouldn't this be optional ?

src/psbt/params.rs
```diff
@@ -0,0 +1,665 @@
...
+
+/// Marker type representing the Replace-By-Fee (RBF) state.
+#[derive(Debug)]
+pub struct ReplaceTx;
+
+/// Alias for [`ReplaceTx`] context marker.
+pub type Rbf = ReplaceTx;
+
+/// Parameters to create a PSBT.
+// TODO: Can we derive `Clone` for this?
```
oleonardolima, line 27: nit: are you looking forward to address this in a follow-up ?

src/psbt/params.rs
```diff
@@ -0,0 +1,665 @@
...
+use crate::collections::{HashMap, HashSet};
+use crate::TxOrdering;
+
+/// Marker type representing the PSBT creation state.
+#[derive(Debug)]
+pub struct CreateTx;
+
+/// Marker type representing the Replace-By-Fee (RBF) state.
+#[derive(Debug)]
+pub struct ReplaceTx;
```
oleonardolima, line 24: question: food for thought, do you think having these structs is better than having an enum (as it's representing a state), for example ?

src/psbt/params.rs
```diff
@@ -0,0 +1,665 @@
...
+    /// ancestor necessarily invalidates the descendant.
+    ///
+    /// # Panics
+    ///
+    /// Panics if `txs` is empty.
+    pub fn replace_txs(self, txs: &[Arc<Transaction>]) -> PsbtParams<Rbf> {
+        assert!(
+            !txs.is_empty(),
+            "replace_txs requires at least one transaction"
+        );
```
oleonardolima: fix: it's way better if we return an error instead of panicking here.

src/psbt/params.rs
```diff
@@ -0,0 +1,665 @@
...
+    /// Any txouts for which the `predicate` returns `false` will be excluded from coin selection,
+    /// otherwise any coin in the wallet that is mature and spendable will be eligible for
+    /// selection.
+    pub fn filter_utxos<F>(&mut self, predicate: F) -> &mut Self
+    where
+        F: Fn(&FullTxOut<ConfirmationBlockTime>) -> bool + Send + Sync + 'static,
+    {
+        self.utxo_filter = UtxoFilter(Arc::new(predicate));
+        self
+    }
```
oleonardolima, line 397: nit: the method name is a bit odd, it's not filtering anything, it's setting the utxo_filter,  updating it to `utxo_filter` is better.

src/wallet/error.rs
```diff
@@ -365,3 +366,95 @@ impl fmt::Display for BuildFeeBumpError {
 }
 
 impl core::error::Error for BuildFeeBumpError {}
+
+/// Error when creating a PSBT.
+#[derive(Debug)]
+#[non_exhaustive]
+pub enum CreatePsbtError {
+    /// No Bnb solution.
+    Bnb(bdk_coin_select::NoBnbSolution),
```
oleonardolima, line 377: Does `bdk_coin_select` has a single `Error` enum ? If that's the case, it's better to have a single `CoinSelection(bdk_coin_select::Error)` instead.

src/wallet/mod.rs
```diff
@@ -2895,6 +2961,779 @@ impl Wallet {
...
+        // TODO: Currently BDK has no notion of MTP, we can use the confirmation block time for now.
+        let time =
+            absolute::Time::from_consensus(anchor.confirmation_time.try_into().ok()?).ok()?;
+        Some(ConfirmationStatus {
+            height,
+            prev_mtp: Some(time),
+        })
+    } else {
+        None
+    }
```
oleonardolima, line 2959: nit: using `match` would read way better than if-let.

src/wallet/mod.rs
```diff
@@ -2895,6 +2961,779 @@ impl Wallet {
...
+    /// When the caller supplies a [`ChangeScript`] via `override_script`, it is passed through
+    /// unchanged and `None` is returned for the change info (no address needs to be revealed).
+    ///
+    /// Otherwise we select the next unused address (already-revealed-but-unused first, then the
+    /// next-to-be-revealed index) **without** mutating wallet state. Revelation is deferred to
+    /// after all error paths have been cleared by the caller.
+    fn peek_change_info(
+        &self,
+        override_script: Option<ChangeScript>,
+    ) -> (Option<(KeychainKind, u32, ScriptBuf)>, ChangeScript) {
```
oleonardolima: fix: it's better if you keep the peek method pure, the override is a short-circuit that should be handled by the callers instead.

src/wallet/mod.rs
```diff
@@ -2895,6 +2961,779 @@ impl Wallet {
...
+                            .next_index(change_keychain)
+                            .expect("keychain must exist");
+                        let spk = self
+                            .peek_address(change_keychain, next_index)
+                            .script_pubkey();
+                        (next_index, spk)
+                    });
+                let descriptor = self.public_descriptor(change_keychain);
+                let definite_descriptor = descriptor
+                    .at_derivation_index(index)
```
oleonardolima: question: why go the `from_descriptor` route if you already have the `spk` ?

src/wallet/mod.rs
```diff
@@ -2895,6 +2961,779 @@ impl Wallet {
...
+        (assets, txouts)
+    }
+
+    /// Filters wallet `txos` by the spending criteria.
+    ///
+    /// - `policy`: Closure indicating whether the output should be kept, used by some callers to
+    ///   apply additional filters as in the case of RBF.
+    fn filter_spendable<'a, I, C, F>(
+        &'a self,
+        txos: I,
```
oleonardolima, line 3047: s/txos/txouts/g

src/wallet/mod.rs
```diff
@@ -2895,6 +2961,779 @@ impl Wallet {
...
+        params: &'a PsbtParams<C>,
+        policy: F,
+    ) -> impl Iterator<Item = FullTxOut<ConfirmationBlockTime>> + 'a
+    where
+        I: IntoIterator<Item = FullTxOut<ConfirmationBlockTime>> + 'a,
+        F: Fn(&FullTxOut<ConfirmationBlockTime>) -> bool + 'a,
+    {
+        let current_height = params.maturity_height.unwrap_or(self.chain.tip().height());
```
oleonardolima, line 3055: nit: it'd be clear if you stick to `maturity_height`, or just `height`.

src/wallet/mod.rs
```diff
@@ -2895,6 +2961,779 @@ impl Wallet {
...
+            // Exclude immature outputs.
+            if !txo.is_mature(current_height) {
+                return false;
+            }
+            // Exclude spent outputs.
+            if txo.spent_by.is_some() {
+                return false;
+            }
+            true
+        })
```
oleonardolima, line 3078: nit: IIRC you could do it lazily by chaining the `.filter`'s instead.

src/psbt/params.rs
oleonardolima: I've been reviewing commit-by commit, so you can ignore this comment now.

src/psbt/params.rs
```diff
@@ -30,7 +30,7 @@ pub type Rbf = ReplaceTx;
 // TODO: Can we derive `Clone` for this?
 #[derive(Debug)]
 pub struct PsbtParams<C> {
-    /// Set of selected UTXO outpoints.
+    /// Set of selected UTXO outpoints, `HashSet` ensures uniqueness
     pub(crate) set: HashSet<OutPoint>,
```
oleonardolima, line 31: question: woulndn't be using something as `select` be clearer ?

src/wallet/error.rs
oleonardolima: I even notice it throughout the commits, there's no BnB in the `SelectionStrategy` enum, it's LowestFee instead 🤔.

src/wallet/mod.rs
```diff
@@ -3251,30 +3253,31 @@ impl Wallet {
...
+                    .select_with_algorithm(selection_algorithm_lowest_fee_bnb(
+                        longterm_feerate,
+                        max_rounds,
+                    ))
+                    .map_err(CreatePsbtError::Bnb)?;
+            }
+            SelectionStrategy::SingleRandomDraw => {
+                // We should have shuffled candidates earlier, so just select
+                // until the target is met.
+                // TODO: Can we shuffle candidates here instead using select_with_algo?
```
oleonardolima: question: is this something you'd like to address in this PR or in a follow-up ?

oleonardolima: Got it in the next commit :)

justfile
```diff
@@ -17,8 +20,9 @@ build:
 # Check code: formatting, compilation, linting, and commit signature
 check:
    cargo +nightly fmt --all -- --check
-   cargo check --all-features --all-targets
-   cargo clippy --all-features --all-targets -- -D warnings
+   RUSTFLAGS="" cargo check --no-default-features --features miniscript/no-std,bdk_chain/hashbrown --all-targets
+   RUSTFLAGS="" cargo check --features {{FEATURES}} --all-targets
+   RUSTFLAGS="--cfg bdk_wallet_unstable" cargo clippy --all-features --all-targets -- -D warnings
```
oleonardolima: question: why not add add a check-unstable instead ?

oleonardolima (commented):
I left a few comments, and questions. I'd like to do some manual testing too, not only with the examples in regtest.

The commit history could use some work, by rewriting and restructuring it, it'll help other reviewers.

src/psbt/params.rs
notmandatory: I agree, `utxo_filter` is a clearer name and matches the internal field.

src/psbt/params.rs
```diff
@@ -0,0 +1,871 @@
...
+    ///
+    /// Note: this will only select coins that are eligible to be spent in the transaction,
+    /// and not those that are already excluded or ineligible e.g. if a spending plan can't be
+    /// created.
+    All,
+    /// Custom
+    Custom {
+        /// Selection algorithm
+        #[allow(clippy::type_complexity)]
+        algorithm: Arc<dyn Fn(&mut Selector) -> Result<(), SelectorError>>,
```
notmandatory: I think this custom algo function also needs to be `Send + Sync` to be consistent with the param type used above in the `coin_selection_algorithm` function. 

```suggestion
        algorithm: Arc<dyn Fn(&mut Selector) -> Result<(), SelectorError> + Send + Sync>,
```

src/psbt/params.rs
```diff
@@ -0,0 +1,871 @@
...
+/// Parameters to create a PSBT.
+// TODO: Can we derive `Clone` for this?
+#[derive(Debug)]
+pub struct PsbtParams<C> {
+    /// Set of selected UTXO outpoints, `HashSet` ensures uniqueness
+    pub(crate) set: HashSet<OutPoint>,
+    /// List of UTXO outpoints to spend.
+    pub(crate) utxos: Vec<OutPoint>,
+    /// List of planned transaction [`Input`]s.
+    pub(crate) inputs: Vec<Input>,
```
evanlinjin: Instead of having both `utxos` (resolved by wallet) and `inputs` (external UTXOs), I think there should be an unified enum:

```rust
enum InputSpec {
    Owned(OutPoint),   // to be resolved by wallet
    Planned(Input),    // taken as-is
}
```

This will fix an ordering bug with `Untouched` - where the order does not respect the order the UTXOs were introduced if we have both wallet-resolved and external inputs.

evanlinjin (commented):
Outstanding review comments from #297.

src/psbt/params.rs
```diff
@@ -0,0 +1,871 @@
...
+/// Marker type representing the PSBT creation state.
+#[derive(Debug)]
+pub struct CreateTx;
+
+/// Marker type representing the Replace-By-Fee (RBF) state.
+#[derive(Debug)]
+pub struct ReplaceTx;
+
+/// Alias for [`ReplaceTx`] context marker.
+pub type Rbf = ReplaceTx;
```
notmandatory: nit: do we really need this alias type? In the code `Rbf` is used everywhere instead of the `ReplaceTx` marker type. But the docs for `ReplaceTx` already describe it as used for doing a RBF. My preference would be to remove `Rbf` and just use `ReplaceTx` everywhere.

src/wallet/mod.rs
ValuedMammal: for self-sends (including change) we need the descriptor to update the PSBT outputs.

src/wallet/mod.rs
```diff
@@ -2895,6 +2934,774 @@ impl Wallet {
...
+
+        let target_outputs = self.target_outputs(&params);
+
+        let input_candidates = InputCandidates::new(must_spend, may_spend);
+        if input_candidates.inputs().next().is_none() {
+            let target_amount: Amount = target_outputs.iter().map(|output| output.value).sum();
+            let err = bdk_coin_select::InsufficientFunds {
+                missing: target_amount.to_sat(),
+            };
+            return Err(CreatePsbtError::InsufficientFunds(err))?;
```
notmandatory, line 3546: nit: it looks strange to throw an error while returning an error. The below should be more idiomatic.
```suggestion
            return Err(CreatePsbtError::InsufficientFunds(err).into());
```

notmandatory (commented):
I did an initial review (with LLM help) and new changes look good. 

One other suggestion, but doesn't need to be done in this PR, is figure out how we want to document *experimental* and possibly unstable APIs in the rust docs.

The LLM suggested adding:
```rust
/// **Unstable**: requires the `bdk-tx` Cargo feature and the
/// `--cfg bdk_wallet_unstable` rustc flag; may change in any minor release.
///
```

sdmg15 (commented):
tAck https://github.com/bitcoindevkit/bdk_wallet/pull/516/changes/319df5e3e634ffbac98a1060c6e1240e5998d165
Haven't spotted anything major. Would probably do a follow up and create examples on the book-of-bdk as suggested during the call which would include using planning, rbf and create_psbt.

ValuedMammal:
### How is the nLockTime set?

First, bdk_tx takes the min_locktime (defaulting to `LockTime::ZERO` if not set) and the absolute timelock extracted from each input's spending plan, and returns the maximum of those values — as long as the units agree. After that, if anti-fee-sniping is used, the locktime may be raised further to the given chain tip, or instead sets a relative sequence on a randomly chosen Taproot input (BIP326 nSequence path). AFS hard-errors if the locktime at that point is time-based, so enabling both time-based CLTVs and AFS is a `CreatePsbtError::Psbt(AntiFeeSnipingError::UnsupportedLockTime)`.

The four scenarios then fall out naturally:

**Neither min_locktime nor anti_fee_sniping set**: min_locktime = LockTime::ZERO. Final locktime is the maximum CLTV requirement across all inputs, or zero if none of the inputs carry a CLTV.

**Only min_locktime set**: The floor is raised from zero to the caller-specified value. Final locktime = max(min_locktime, max(input CLTVs)), unit-compatible. Useful for, e.g., enforcing a block-height floor independently of any script requirements.

**Only anti_fee_sniping set**: min_locktime = zero, so the pre-AFS locktime is just the input CLTVs (or zero). AFS then applies BIP326: with 50/50 probability it sets tx.lock_time to tip_height, otherwise it sets a relative sequence on a Taproot input. If the locktime after CLTV accumulation is already above tip_height (e.g., a CLTV that pins to a future height), AFS leaves it untouched. Note that AFS is not automatically derived from the wallet's chain tip; you must supply the current block height explicitly.

**Both min_locktime and anti_fee_sniping set**: min_locktime acts as the floor going into AFS. The final value is effectively max(min_locktime, input CLTVs, tip_height) — though the exact outcome depends on BIP326's nLockTime vs nSequence coin flip. If min_locktime is already above tip_height, AFS's locktime branch won't change it, so the min_locktime effectively wins.

Note that `maturity_height` has no bearing on nLockTime. It is used to decide whether coinbase UTXOs have reached the 100-confirmation threshold and are eligible for coin selection at all. It's a gate on which outputs may enter the candidate set, not a contributor to the transaction's locktime.

`add_assets` can be used to tell the planner "an OP_CLTV of up to this height will be satisfied." The planner uses that to choose a spending path. If it selects a branch that requires after(N), the resulting Plan carries absolute_timelock = Some(LockTime::Blocks(N)), and that is what `Input::absolute_timelock()` returns.

---

**Some important nuances**:

`Assets::after()` is a planner hint, not a direct locktime assertion. The height that ends up in the transaction's nLockTime is the descriptor's required CLTV value (from the plan), not the value you passed to `Assets::after()`. If you say `Assets::after(500)` but the descriptor has after(1000), the planner can't satisfy that branch anyway — it will either fail to plan or pick a different branch.

The fallback in `plan_input` is interesting. If you don't call `add_assets` at all, the wallet falls back to `LockTime::from_consensus(current_height)`, i.e. the current chain tip height, as the inferred absolute timelock for planning. This means the wallet will automatically select CLTV branches that are currently satisfiable at the last known tip height, which is a sensible default. It also means that for a simple mandatory-CLTV policy like `and_v(v:pk(A), after(1000))`, the plan will carry absolute_timelock = Some(1000) regardless of what you pass to add_assets, as long as the tip is past height 1000.

So the full picture of locktime inputs is actually three sources, one of which is indirect:

- `min_locktime`, a direct floor on tx.lock_time
- Input CLTV requirements, which flow from the descriptor's spending plan. That plan may be shaped by `add_assets`, making CLTV branches reachable, and through that, indirectly controls whether and what absolute timelock an input contributes to the transaction.
- `anti_fee_sniping_height(tip_height)`, raises locktime toward the tip, or deflects to nSequence

ValuedMammal:
Tested the PSBT and RBF happy path locally on a fork of [`satchl`](https://github.com/ValuedMammal/satchl) using a simple taproot descriptor (on signet). https://mempool.space/signet/tx/d45ebd30e2e2448f86ed8a32d3fcd859c3ab1bd5d1ccda6581246690d752509c

oleonardolima:
@ValuedMammal Are you looking forward to address the other comments in a follow-up ?

src/psbt/params.rs
ValuedMammal: To derive Clone for PsbtParams, we need `ChangeScript: Clone` which in turn depends on https://github.com/rust-bitcoin/rust-miniscript/pull/886.

src/psbt/params.rs
ValuedMammal: The marker type is used as the generic type parameter and this affords type safety: you can't call `replace_by_fee` with a `PsbtParams<CreateTx>` for example. Indeed there may be other variations of the typestate pattern in rust, https://cliffle.com/blog/rust-typestate/.

src/wallet/error.rs
ValuedMammal: > Does `bdk_coin_select` has a single `Error` enum ?

Not that I know of

> there's no BnB in the SelectionStrategy enum, it's LowestFee instead

LowestFee is what is known as a Branch n Bound "metric". It implements [`BnbMetric`](https://github.com/bitcoindevkit/coin-select/blob/a66cf8244fe4d221d818b2b3929c9107ed0e28d9/src/metrics/lowest_fee.rs#L25).

src/wallet/mod.rs
ValuedMammal: This is a style preference. It's equally lazy as far as I can tell.

src/wallet/error.rs
ValuedMammal: I agree with @noahjoeris on this one https://github.com/bitcoindevkit/bdk_wallet/pull/297#discussion_r3195484984

src/wallet/mod.rs
ValuedMammal: Currently calculating the descendant fee is done on a best-effort basis. IMO failure to calculate a descendant's fee shouldn't prevent you from creating the transaction.

ValuedMammal:
@oleonardolima No follow-up plans yet, except maybe the docs suggestion https://github.com/bitcoindevkit/bdk_wallet/pull/516#pullrequestreview-4740199325.

PR #520, opened 2026-07-28
fix(wallet): validate network and genesis hash consistency

### LINKED ISSUE #46: Verify network consistency between chain genesis and descriptors
"Consistency between descriptor network and chain genesis hash isn't checked in load_with_params nor in create_with_params."

busayo-OD:
Fixes #46

### Description

Previously, `Wallet::create_with_params` and `Wallet::load_with_params` allowed `Network::Bitcoin` to be paired with a non-mainnet genesis hash, creating an internally inconsistent wallet configuration.

This PR adds validation to both entry points:

- `Wallet::create_with_params` returns `DescriptorError::GenesisHashMismatch` when a mainnet wallet is configured with a non-mainnet genesis hash.
- `Wallet::load_with_params` returns  `LoadError::GenesisNetworkMismatch` when persisted wallet metadata contains the same inconsistency.

Regression tests have been added for both code paths.

### Notes to the reviewers

This PR revisits the earlier implementation attempt ([#1777](https://github.com/bitcoindevkit/bdk/pull/1777)) and implements the approach that emerged from the review discussion:
- reject only custom genesis hashes for `Network::Bitcoin`;
- preserve support for custom genesis hashes on Testnet, Signet, and Regtest;
- apply the same validation during both wallet creation and wallet loading.

To report these validation failures without changing the existing API signatures, this PR introduces `DescriptorError::GenesisHashMismatch` and `LoadError::GenesisNetworkMismatch`.

Neither `DescriptorError` nor `LoadError` is `#[non_exhaustive]`, so adding these variants is a semver-breaking change for any downstream code with an exhaustive `match` on either enum.

### Changelog notice

```text
Validate that `Network::Bitcoin` is paired with the Bitcoin genesis hash during wallet creation and loading.
```

#### All Submissions

- [x] I've signed all my commits
- [x] I ran `just p` before pushing

#### Bugfixes

- [ ] This pull request breaks the existing API
- [x] I've added tests to reproduce the issue which are now passing
- [x] I'm linking the issue being fixed by this PR

### DIFF
--- a/src/descriptor/error.rs
+++ b/src/descriptor/error.rs
@@ -10,6 +10,7 @@
 // licenses.
 
 //! Descriptor errors
+use bitcoin::{BlockHash, Network};
 use core::fmt;
 
 /// Errors related to the parsing and usage of descriptors
@@ -44,6 +45,13 @@ pub enum Error {
     Hex(bitcoin::hex::HexToBytesError),
     /// The provided wallet descriptors are identical
     ExternalAndInternalAreTheSame,
+    /// The provided genesis hash does not match the expected mainnet genesis hash
+    GenesisHashMismatch {
+        /// The configured network
+        network: Network,
+        /// The genesis hash that was provided
+        genesis_hash: BlockHash,
+    },
 }
 
 impl From<crate::keys::KeyError> for Error {
@@ -84,6 +92,13 @@ impl fmt::Display for Error {
             Self::ExternalAndInternalAreTheSame => {
                 write!(f, "External and internal descriptors are the same")
             }
+            Self::GenesisHashMismatch {
+                network,
+                genesis_hash,
+            } => write!(
+                f,
+                "Genesis hash {genesis_hash} does not match expected genesis hash for network {network}"
+            ),
         }
     }
 }
--- a/src/wallet/error.rs
+++ b/src/wallet/error.rs
@@ -38,6 +38,13 @@ pub enum LoadError {
     MissingGenesis,
     /// Data loaded from persistence is missing descriptor.
     MissingDescriptor(KeychainKind),
+    /// The network's mainnet genesis hash does not match the loaded genesis hash.
+    GenesisNetworkMismatch {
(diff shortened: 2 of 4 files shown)


ValuedMammal:
It's curious why `create_with_params` doesn't have its own CreateError (with DescriptorError as a variant). That would make sense, as `GenesisHashMismatch` isn't exactly related to the wallet descriptors.

This might expand the scope somewhat but I'd suggest having a `pub struct GenesisNetworkMismatch { ... }` with a single Display implementation, and reuse it in both `CreateError` and `LoadError`. In any case, adding an enum variant will be a breaking change, and should be noted in the commit/PR description. If updating an error type we might as well make it non_exhaustive in case we want to add variants in the future.

Also just noting it might help to wait until #524 is in since it also affects the wallet constructor methods.

busayo-OD:
> It's curious why `create_with_params` doesn't have its own CreateError (with DescriptorError as a variant). That would make sense, as `GenesisHashMismatch` isn't exactly related to the wallet descriptors.
> 
> This might expand the scope somewhat but I'd suggest having a `pub struct GenesisNetworkMismatch { ... }` with a single Display implementation, and reuse it in both `CreateError` and `LoadError`. In any case, adding an enum variant will be a breaking change, and should be noted in the commit/PR description. If updating an error type we might as well make it non_exhaustive in case we want to add variants in the future.
> 
> Also just noting it might help to wait until #524 is in since it also affects the wallet constructor methods.

Thanks for the review! Using a shared error struct makes more sense. `GenesisHashMismatch` in `DescriptorError` was just the minimal-change choice. I’ll revisit the error handling, including `#[non_exhaustive]`, after #524 lands and note the breaking changes accordingly.

SIDHARTH20K4:
Should we also test that mainnet with the correct genesis hash succeeds? And that Testnet/Signet still allow custom genesis hashes? I can try writing these if helpful

@ValuedMammal

PR #524, opened 2026-08-04
Feature/multi keychain take 3

thunderbiscuit:
Take 3 on the multi-keychain PR feature. I'm opening this as a new PR in order to not lose all the work done in #318, since it's a fairly new/reimagined way to work it (2 PRs together, etc.). See #523 for part 1 of this.

Note that this is made against the [pre-multi-keychain](https://github.com/bitcoindevkit/bdk_wallet/tree/pre-multi-keychain) branch.

Quite experimental at this point, but I wanted to share where I'm at so others can build upon it.

### TODO

- [ ] I have removed the APIs related to multipath descriptors as they were focused on 2-path descriptors, but we should now be able to handle all multipath. I think this part needed a bit more thinking and I wanted to punt on it for now to focus on the core of the feature. Multipath is still supported, you simply have to "unpack" them before giving them to the `KeyRing`.
- [ ] Figure out how to deal with change in the Wallet::create_psbt API. To keep things simple here I require users always set the keychain explicitly, but it's the sort of thing that users would probably love to be able to set once and forget. The example of wallets that use 2 keychains and one of those is the change keychain is the most obvious one.
- [ ] One of the issues left is the `Wallet::balance` method not being able to tell what's a change keychain and what's not, and so the concept of "trusted" is a bit lost in the current API (all defaults to untrusted until the UTXO is mined). But this is I believe fixed by #431 and would land hopefully before this, and so I choose to let this one sit as is for now. 

### Changelog notice

TODO

### Before submitting

- [x] This PR breaks the existing API

### DIFF
--- a/src/keyring/error.rs
+++ b/src/keyring/error.rs
@@ -1,0 +1,49 @@
+//! Errors produced while building a [`KeyRing`](crate::keyring::KeyRing).
+
+use crate::descriptor::DescriptorError;
+use alloc::boxed::Box;
+use core::fmt;
+use miniscript::{Descriptor, DescriptorPublicKey};
+
+/// Error returned when a descriptor cannot be added to a [`KeyRing`](crate::keyring::KeyRing).
+#[derive(Debug)]
+#[non_exhaustive]
+pub enum KeyRingError<K> {
+    /// The descriptor is invalid, does not match the keyring's network, or fails the checks
+    /// applied to every wallet descriptor.
+    Descriptor(DescriptorError),
+    /// The keychain is already assigned to a different descriptor.
+    ///
+    /// A keychain identifies exactly one descriptor for the life of the wallet.
+    KeychainAlreadyAssigned(K),
+    /// The descriptor is already assigned to a different keychain.
+    ///
+    /// Two keychains sharing a descriptor cannot be told apart when attributing discovered
+    /// outputs, so the indexer rejects it.
+    DescriptorAlreadyAssigned(Box<Descriptor<DescriptorPublicKey>>),
+}
+
+impl<K> From<DescriptorError> for KeyRingError<K> {
+    fn from(err: DescriptorError) -> Self {
+        Self::Descriptor(err)
+    }
+}
+
+impl<K: fmt::Display> fmt::Display for KeyRingError<K> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            Self::Descriptor(e) => e.fmt(f),
+            Self::KeychainAlreadyAssigned(keychain) => {
+                write!(f, "keychain {keychain} is already assigned to a descriptor")
+            }
+            Self::DescriptorAlreadyAssigned(descriptor) => {
+                write!(
+                    f,
+                    "descriptor {descriptor} is already assigned to a keychain"
+                )
+            }
+        }
+    }
+}
+
+impl<K: fmt::Display + fmt::Debug> core::error::Error for KeyRingError<K> {}
--- a/src/keyring/error.rs
+++ b/src/keyring/error.rs
@@ -29,12 +29,15 @@ impl<K> From<DescriptorError> for KeyRingError<K> {
     }
 }
 
-impl<K: fmt::Display> fmt::Display for KeyRingError<K> {
+impl<K: fmt::Debug> fmt::Display for KeyRingError<K> {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         match self {
             Self::Descriptor(e) => e.fmt(f),
             Self::KeychainAlreadyAssigned(keychain) => {
-                write!(f, "keychain {keychain} is already assigned to a descriptor")
+                write!(
+                    f,
+                    "keychain {keychain:?} is already assigned to a descriptor"
+                )
             }
             Self::DescriptorAlreadyAssigned(descriptor) => {
                 write!(
@@ -46,4 +49,4 @@ impl<K: fmt::Display> fmt::Display for KeyRingError<K> {
     }
 }
 
-impl<K: fmt::Display + fmt::Debug> core::error::Error for KeyRingError<K> {}
+impl<K: fmt::Debug> core::error::Error for KeyRingError<K> {}
--- a/src/wallet/changeset.rs
+++ b/src/wallet/changeset.rs
@@ -401,30 +401,27 @@ where
         self.indexer.persist_to_sqlite(db_tx)?;
         Ok(())
     }
-}
 
-#[cfg(feature = "rusqlite")]
-impl ChangeSet<crate::KeychainKind> {
     /// Recover descriptors written by schema v0 or v1.
     ///
     /// Those versions stored the wallet's two descriptors as `descriptor` and `change_descriptor`
-    /// columns on a single row, rather than one row per keychain. This reads them and maps them
-    /// onto [`External`](crate::KeychainKind::External) and
-    /// [`Internal`](crate::KeychainKind::Internal).
+    /// columns on a single row, rather than one row per keychain in
+    /// [`WALLET_KEYCHAIN_TABLE_NAME`](Self::WALLET_KEYCHAIN_TABLE_NAME).
     ///
-    /// Only meaningful for wallets keyed by [`KeychainKind`](crate::KeychainKind): interpreting
-    /// the two legacy columns *requires* knowing they mean external and change. A wallet using a
-    /// custom keychain type has no v0/v1 database to recover, since those schemas predate custom
-    /// keychains entirely.
+    /// The legacy columns carry no keychain identifier of their own — the schema encoded it
+    /// positionally. They are recovered here by asking `K` to parse the same strings
+    /// [`KeychainKind`](crate::KeychainKind) serialises to, `"external"` and `"internal"`. A
+    /// keychain type that does not recognise them cannot have written a v0/v1 database in the
+    /// first place, so it is left untouched.
     ///
-    /// Descriptors already present are left alone, so this never overwrites what schema v2 holds.
+    /// Existing keychains are never overwritten: anything already read from the v2 table wins.
     pub fn read_legacy_descriptors(
         db_tx: &chain::rusqlite::Transaction,
         changeset: &mut Self,
     ) -> chain::rusqlite::Result<()> {
-        use crate::KeychainKind;
         use chain::Impl;
         use chain::rusqlite::OptionalExtension;
+        use chain::rusqlite::types::ValueRef;
 
         let mut statement = db_tx.prepare(&format!(
             "SELECT descriptor, change_descriptor FROM {}",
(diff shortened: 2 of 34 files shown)


thunderbiscuit:
A few ideas for interesting helper features to add to the `KeyRing`:

1. A `KeyRing::standard` method that takes two descriptors and slots them right into `KeychainKind::External` and `KeychainKind::Internal`. This is basically allowing anyone who builds very run-of-the-mill wallets to do it quickly and simply, using the current approach of KeychainKind as their `K` keychain identifier.
2. A more elegant API for dealing with multipath descriptors. This has been a request for a while and we currently only have so-so methods to work with them, mostly because we only could work with 2 keychains at a time. With the multi-keychain paradigm, really any number of keychains that the multpath descriptors expand to can be added to the wallet. We just need a way to deal with this elegantly (and minimize footguns).

thunderbiscuit:
TODO: Figure out how to deal with change in the `Wallet::create_psbt` API. To keep things simple here I require users always set the keychain explicitly, but it's the sort of thing that users would probably love to be able to set once and forget. The example of wallets that use 2 keychains and one of those is the change keychain is the most obvious of those, where clearly the user intends for one of those keychains to be used for change for all transactions. 

My current naive idea to deal with this would be to add a field on the wallet `change: K` where users can define a keychain that the Wallet::create_psbt could always use in case no change keychain is set in the psbt creation.

@ValuedMammal happy to hear your thoughts on this.

One thought on [this comment](https://github.com/bitcoindevkit/bdk_wallet/pull/318#issuecomment-5006792184): by default the wallet would currently use UTXOs from all keychains to build transactions (maybe that needs work too; kind of like locking UTXOs, you might want to lock keychains or at least prevent the wallet mixing up all your keychains for privacy reasons), but in any case the base workflow would use all keychains. If each of those had defined a sister change keychain, then the wallet is back at square 1, not really knowing which one to use for change. Here again, explicitly setting the keychain to use for change is preferable IMO.

src/keyring/error.rs
```diff
@@ -0,0 +1,52 @@
...
+    Descriptor(DescriptorError),
+    /// The keychain is already assigned to a different descriptor.
+    ///
+    /// A keychain identifies exactly one descriptor for the life of the wallet.
+    KeychainAlreadyAssigned(K),
+    /// The descriptor is already assigned to a different keychain.
+    ///
+    /// Two keychains sharing a descriptor cannot be told apart when attributing discovered
+    /// outputs, so the indexer rejects it.
+    DescriptorAlreadyAssigned(Box<Descriptor<DescriptorPublicKey>>),
```
thunderbiscuit, line 23: Need to confirm where exactly this check will succeed/fail given corner cases, because I know we've discussed this in the past and I need a refresh on it. Posting here for self-review.

thunderbiscuit (commented):
Leaving here for review for tomorrow.

src/keyring/error.rs
110CodingP: Won't we encounter this, if we try to add a descriptor to the keyring but it already exists under a different keychain?

src/wallet/mod.rs
```diff
@@ -1105,2638 +927,2714 @@ impl Wallet {
...
+        mut confirmation_height_for_input: F,
+        clear_output_derivations: bool,
+    ) -> Result<FinalizePsbtOutcome, IndexOutOfBoundsError>
+    where
+        F: FnMut(usize, &bitcoin::TxIn) -> Option<u32>,
+    {
+        let tx = &psbt.unsigned_tx;
+        if psbt.inputs.len() < tx.input.len() {
+            return Err(IndexOutOfBoundsError::new(
+                psbt.inputs.len(),
```
110CodingP, line 1029: `tx.input.len()` ?

src/wallet/mod.rs
```diff
@@ -627,13 +448,13 @@ impl Wallet {
...
-        let keychain = self.map_keychain(keychain);
+    pub fn reveal_next_address(&mut self, keychain: K) -> AddressInfo<K> {
+        let keychain = keychain;
         let index = &mut self.tx_graph.index;
         let stage = &mut self.stage;
 
         let ((index, spk), index_changeset) = index
-            .reveal_next_spk(keychain)
+            .reveal_next_spk(keychain.clone())
             .expect("keychain must exist");
```
110CodingP, line 458: We should return an error instead?

src/wallet/mod.rs
```diff
@@ -1105,2638 +927,2714 @@ impl Wallet {
...
-        let current_height = match params.current_height {
-            // If they didn't tell us the current height, we assume it's the latest sync height.
-            None => {
-                let tip_height = self.chain.tip().height();
-                absolute::LockTime::from_height(tip_height).expect("invalid height")
+        for (n, input) in tx.input.iter().enumerate() {
+            let psbt_input = &psbt
+                .inputs
+                .get(n)
+                .ok_or(IndexOutOfBoundsError::new(n, psbt.inputs.len()))?;
```
110CodingP, line 1040: We have already ensured this won't happen?

src/wallet/mod.rs
```diff
@@ -1105,2638 +927,2714 @@ impl Wallet {
...
 
-            outgoing += value;
-        }
+    /// Return the secp256k1 context used for all signing operations.
+    pub fn secp_ctx(&self) -> &SecpCtx {
+        &self.secp
+    }
 
-        fee_amount += fee_rate * tx.weight();
+    /// The derivation index of this wallet. It will return `None` if it has not derived any
```
110CodingP, line 1131: We need to handle the case when `keychain` doesn't exist here too.

110CodingP:
> My current naive idea to deal with this would be to add a field on the wallet `change: K` where users can define a keychain that the Wallet::create_psbt could always use in case no change keychain is set in the psbt creation.

If convenience is the only reason for this change can we avoid persisting this? Can we have `change_keychains: Vec<K>`? The user could supply the value whenever we create/load the wallet. We could the choose the first one if we cannot decide which one to use? If the no change-keychain is specified and also no change_keychain is provided during tx-building we could throw up an error.

CHEF-SAVY:
Run cargo audit locally to see the full details of the 4 securities issues.

src/keyring/error.rs
thunderbiscuit: > Won't we encounter this, if we try to add a descriptor to the keyring but it already exists under a different keychain?

Yes and no. The equality check is not perfect if I remember correctly, and we ended up deciding that a hash of the first script derived by the descriptor was a more robust (if not perfect) way to check that two descriptors were basically "the same".

The wallet creation does that check using the `descriptor_id()` method (what I'm referring to above), so I'll use that instead of the equality used previously.

thunderbiscuit:
I'll follow up on the [comments here](https://github.com/bitcoindevkit/bdk_wallet/pull/318#issuecomment-5243820008) in this thread! @ValuedMammal 

> The privacy focused approach is to choose the change keychain based on the output type of the outputs of the transaction to be created, not the keychain of the utxos pulled in as inputs.

You're correct I got this backwards there for a second.

> the cleanest way to define the change keychain is at wallet creation time, because it's tied directly to a wallet descriptor. Supplying the change keychains another way means you'll have to handle a possibly non-existing keychain.

Yeah I see your point there. It conflicts with @110CodingP's idea above ("If convenience is the only reason for this change can we avoid persisting this?"), but if we really value this idea of "set it and forget it" then I think adding it to persistence could be worth the trouble.

<br>

### On having multiple change keychains

I don't personally like the idea of defining more than 1 change keychain wallet-wide _at the `Wallet` level_ (and even then! But I don't like change keychains at all to be honest so I understand I come at this a bit biased). Overall, I think it implies the wallet can do more than what I think it can reasonably do (picking correctly between multiple change is just a hard problem that's better handled outside of the wallet IMO, see next section). Defining them app-layer within your keychain identifiers `K` is of course perfect for this, and can be used powerfully.

I think a reasonable feature would be to be able to define a single global change keychain, which would come into play when going down the order of preference for change keychain selection in the psbt creation process as you defined in the linked post. This would also facilitate and keep simple the "basic case" of 2-keychain wallets as we have them now in `Wallet`. 

<br>

### Order of preference for change

> So for PSBT creation the order of precedence is
> 1. User provided change-script/descriptor via the params
> 2. Auto derive from a change keychain the Wallet/KeyRing knows about
> 3. A third option could be a single globally configurable K value, but the real benefit of this comes from persisting it, so you can truly set it and forget it

Thanks for laying this out. I would do exactly this, but remove step 2. Or maybe I can be convinced, or it could come as a new feature at some point?

<br>

### On choosing the change based on payment scripts

For me, this lives better at the application layer than the bdk_wallet layer. The basic privacy workflow is this:
1. Figure out what type of script I'm paying to
4. Pick a change script of the same type to ensure alll outputs are of the same script type

Attempting to do this on the bdk_wallet side is probably not impossible, but feels messy to me and very hard to do well, because the wallet would need to know so much about the user/app layer, i.e. carry a lot of metadata on all keychains and when to use which, which is exactly what the `K` approach is powerful for: letting the user hold keychain metadata with the descriptors, and leverage this for transaction building.

Even for the base case, the wallet would need to be able to parse the script type being paid to (what if it's a few different ones?), then see if it has keychains that are marked as `change` and have the same script type (again what if you pay to a tapscript address and all your 7 keychains are tapscript keychains?), and then pick one of those (which one? random? me don't like random). Again not impossible by any means, but this feels perfect for business-logic added on the application layer (at least to me at the moment).

thunderbiscuit:
This was a long post but here is what I suggest I focus on for the next round of this:

1. Make sure the current 2-keychains `External/Internal` flow is clean and simple to use, with similar features to what we have now.
2. Keep complete flexibility for what I expect is more advanced + custom users of the library (those who define many keychains).

110CodingP:
Do we want to allow the `change_keychain` or `internal` marker to be modified after wallet creation? Then this would also require a new field on the `ChangeSet` !

src/wallet/changeset.rs
```diff
@@ -53,10 +55,9 @@ type IndexedTxGraphChangeSet =
 /// * [`tx_graph`](Self::tx_graph)
 /// * [`indexer`](Self::indexer)
 ///
-/// The [`change_descriptor`] is special in that its presence is optional, however the value of the
-/// change descriptor should be defined at wallet creation time and respected for the life of the
-/// wallet, meaning that if a change descriptor is originally defined, it must also be present in
-/// the aggregate change set.
+/// A keychain may be introduced by a later change set — a wallet can start tracking a new one at
```
110CodingP, line 58: Are we planning to allow addition of descriptors in bdk_wallet 4.0 itself?

src/wallet/changeset.rs
```diff
@@ -133,12 +133,13 @@ type IndexedTxGraphChangeSet =
...
 /// [Semantic Versioning]: <https://doc.rust-lang.org/cargo/reference/semver.html>
-#[derive(Default, Debug, Clone, PartialEq, Deserialize, Serialize)]
-pub struct ChangeSet {
-    /// Descriptor for recipient addresses.
-    pub descriptor: Option<Descriptor<DescriptorPublicKey>>,
-    /// Descriptor for change addresses.
-    pub change_descriptor: Option<Descriptor<DescriptorPublicKey>>,
+#[derive(Debug, Clone, PartialEq, Deserialize, Serialize)]
+pub struct ChangeSet<K: Ord> {
+    /// The descriptor tracked by each keychain.
```
110CodingP, line 138: Mammal brought this up on #318 : Since we need to follow the `ChangeSet`'s field-deprecation policy we would need to keep the `change_descriptor` and `descriptor` fields around for v4.

110CodingP:
Also ig we could just let the user specify the `change_keychain` in this PR and have another PR (fixing the way user specifies the change keychain) when we have an agreement over the final API?

Edit: For `KeychainKind` type wallets, users can also create a custom wrapper around our APIs which pass `KeychainKind::External` , this can also solve the convenience problem.

PR #528, opened 2026-08-10
fix(wallet): make sign()'s sighash guard context-aware

### LINKED ISSUE #519: `Wallet::sign()`'s sighash pre-flight guard doesn't reject non-standard sighashes on non-Taproot inputs
While working on #476 (context-aware `allow_all_sighashes` check in `SignerWrapper::sign_input`), it came up that `Wallet::sign()` has a pre-flight guard with the same underlying issue: a dead comparison and no awareness of which inputs are actually Taproot.

The guard in question (src/wallet/mod.rs):

```
if !sign_options.allow_all_sighashes
    && !psbt.inputs.iter().all(|i| {
        i.sighash_type.is_none()
            || i.sighash_type == Some(EcdsaSighashType::All.into())
            || i.sighash_type == Some(TapSighashType::All.into())
            || i.sighash_type == Some(TapSighashType::Default.into())
    })
{
    return Err(SignerError::NonStandardSighash);
}
```

### Problems:

`EcdsaSighashType::All.into()` and `TapSighashType::All.into()` resolve to the same `PsbtSighashType` value, so one of these branches is redundant, same dead comparison as before #476.
More importantly, this check runs as a flat loop over every PSBT input with no idea which input is Taproot and which isn't. `TapSighashType::Default` (byte 0x00) is accepted unconditionally for all inputs, including Legacy/Segwitv0 ones, where `Default` isn't a meaningful ECDSA sighash at all.

### Impact:

I wrote a test that builds a normal Segwitv0, manually sets input 0's sighash_type to `TapSighashType::Default`'s byte value, then calls wallet.sign(...) with `allow_all_sighashes: false`. The guard lets it through as expected from the code above.

The `sign_input()` fix from #476 catches it downstream and returns NonStandardSighash anyway. But Wallet::sign() dispatches to any TransactionSigner, and the whole point of this is that individual signers can reasonably trust it's already filtered non-standard sighashes and skip re-validating themselves. For a custom signer that does trust the guard, this is a real bypass.

### Suggested fix:

Since psbt::Input carries tap_internal_key / tap_merkle_root, Taproot can be inferred per-input without needing a SignerContext lookup

Happy to open a PR for this, flagging as an issue first per usual since I want to confirm the framing is right before writing the fix.

muhahahmad68:
Fixes #519.

### Description

The pre-flight sighash guard in Wallet::sign() checked every PSBT input against a flat set of allowed sighash values without regard to whether that input was actually Taproot. As a result, TapSighashType::Default (0x00) was accepted unconditionally for every input, including Legacy/Segwitv0 ones, where it has no valid ECDSA meaning.

This is currently masked for the built-in software signer, since sign_input()'s guard (introduced in #476) independently rejects it downstream. However, Wallet::sign() dispatches to any TransactionSigner, and the whole point of this pre-flight guard is that individual signers can trust it has already filtered non-standard sighashes.

### Notes to the reviewers

Infer Taproot-ness per input from witness_utxo.script_pubkey, consistent with how complete_transaction() determines Taproot-ness for foreign UTXOs, and apply the correct sighash check for that input's context.

### Tests added:

* sign_rejects_ecdsa_input_with_taproot_default_sighash — confirms a non-Taproot input carrying TapSighashType::Default is now rejected
* sign_allows_taproot_input_with_default_sighash — confirms a genuine Taproot spend still signs and finalizes normally

### Before submitting

### DIFF
--- a/src/wallet/mod.rs
+++ b/src/wallet/mod.rs
@@ -1913,11 +1913,23 @@ impl Wallet {
         // If the user hasn't explicitly opted-in, refuse to sign the transaction unless every input
         // is using `SIGHASH_ALL` or `SIGHASH_DEFAULT` for Taproot.
         if !sign_options.allow_all_sighashes
-            && !psbt.inputs.iter().all(|i| {
-                i.sighash_type.is_none()
-                    || i.sighash_type == Some(EcdsaSighashType::All.into())
-                    || i.sighash_type == Some(TapSighashType::All.into())
-                    || i.sighash_type == Some(TapSighashType::Default.into())
+            && !psbt.inputs.iter().all(|i| match i.sighash_type {
+                None => true,
+                Some(sighash_type) => {
+                    let is_taproot = i
+                        .witness_utxo
+                        .as_ref()
+                        .map(|utxo| utxo.script_pubkey.is_p2tr())
+                        .unwrap_or(false);
+                    if is_taproot {
+                        matches!(
+                            sighash_type.taproot_hash_ty(),
+                            Ok(TapSighashType::All) | Ok(TapSighashType::Default)
+                        )
+                    } else {
+                        matches!(sighash_type.ecdsa_hash_ty(), Ok(EcdsaSighashType::All))
+                    }
+                }
             })
         {
             return Err(SignerError::NonStandardSighash);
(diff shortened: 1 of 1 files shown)


ValuedMammal:
One suggestion I have for determining whether an input is a p2tr spend is it may be safer and more reliabe to test `witness_utxo.script_pubkey.is_p2tr()` rather than the `tap_*` metadata fields (at least that's consistent with how the check is handled elsewhere e.g. `complete_transaction`).

muhahahmad68:
> One suggestion I have for determining whether an input is a p2tr spend is it may be safer and more reliabe to test `witness_utxo.script_pubkey.is_p2tr()` rather than the `tap_*` metadata fields (at least that's consistent with how the check is handled elsewhere e.g. `complete_transaction`).

Good catch

ValuedMammal:
I opened a draft PR to this branch muhahahmad68/bdk_wallet#1, still open for discussion.
The motivation is to handle the shared logic when it comes to prevout resolution
and input validation. For instance in the `allow_all_sighashes` guard, we were
determining P2TR-ness by looking at the `witness_utxo` alone with no additional
validation.

PR #533, opened 2026-08-14
fix(descriptor): propagate threshold errors instead of panicking

### LINKED ISSUE #532: [dsl] Handle `ThresholdError` in `fragment!` macro
**Describe the bug**  

The thresh_vec, multi_vec, and multi_a_vec fragments of the `descriptor!`/`fragment!` macros accept `$thresh:expr` and `$keys:expr` runtime values, with thresh_vec explicitly documented for building "dynamic descriptors" from a runtime Vec. All three call `Threshold::new` and `.expect("valid threshold and pks collection")` on the result. `Threshold::new` returns Err for k == 0, k > n, or n > MAX, so invalid runtime parameters crash the consumer process even though the macro's normal error channel is `Result<_, DescriptorError>`.

**To Reproduce**  

```rust
// `k` (3) is greater than the number of items (2), which is an invalid threshold.
let result = descriptor!(wsh(thresh(3, pk(key1), pk(key2))));

// Same invalid-threshold scenario, but going through the `multi()` fragment
// which builds a `Terminal::Multi` threshold via `make_multi`.
let result = descriptor!(wsh(multi(3, key1, key2)));
```

**Build environment**  
- BDK tag/commit: 49f7955c34d1f558ca3566656faa8b17acffba1c
- OS+version:   
- Rust/Cargo version:  
- Rust/Cargo target:   

- [ ] Yes
- [x] No

**Additional context**  

Suggested fix

Propagate the `ThresholdError` as a `DescriptorError` variant instead of `.expect()` in all three arms, and add a test asserting Err (not panic) for k > n via multi_vec.

busayo-OD:
Fixes #532

### Description

`thresh_vec`, `multi_vec`, and `multi_a_vec` used `.expect(...)` on `Threshold::new`, causing invalid runtime-supplied thresholds to panic instead of returning through the macros' usual `Result<_, DescriptorError>`.

For `thresh_vec`, the error is propagated through its existing `Result` chain. For `multi_vec` and `multi_a_vec`, `keys::make_multi` now accepts a closure returning `Result<Terminal<...>, DescriptorError>`, matching the existing `make_sortedmulti` pattern and removing the need for `.expect(...)`.

### Notes to the reviewers

`miniscript` already provides `Error::Threshold` for invalid thresholds, and `DescriptorError` already wraps `miniscript::Error`, so I reused the existing error path rather than introducing a new variant.

Added `From<miniscript::ThresholdError> for DescriptorError` and simplified all three fragments to propagate the error via `?` directly.

### Changelog notice
```
`keys::make_multi`'s closure now returns `Result<Terminal<DescriptorPublicKey, Ctx>, DescriptorError>` instead of `Terminal<...>` directly.
```

#### All Submissions

#### Bugfixes

### DIFF
--- a/src/descriptor/dsl.rs
+++ b/src/descriptor/dsl.rs
@@ -765,9 +765,11 @@ macro_rules! fragment {
             (keys_acc, net_acc)
         });
 
-        let thresh = $crate::miniscript::Threshold::new($thresh, items).expect("valid threshold and pks collection");
-        $crate::impl_leaf_opcode_value!(Thresh, thresh)
-            .map(|(minisc, _, _)| (minisc, key_maps, valid_network_kinds))
+        (|| -> Result<_, $crate::descriptor::DescriptorError> {
+            let thresh = $crate::miniscript::Threshold::new($thresh, items)?;
+            let (minisc, _, _) = $crate::impl_leaf_opcode_value!(Thresh, thresh)?;
+            Ok((minisc, key_maps, valid_network_kinds))
+        })()
     });
     ( thresh ( $thresh:expr, $( $inner:tt )* ) ) => ({
         let items = $crate::fragment_internal!( @v $( $inner )* );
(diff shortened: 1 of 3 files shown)


Arowolokehinde (commented):
I pulled this branch and ran all three tests locally. The fix correctly stops the fragment! macro from panicking on invalid thresholds and returns `Err(DescriptorError::Miniscript(miniscript::Error::Threshold(_)))` instead, routing errors through the existing Result channel rather than bypassing it with .expect().

hash-ty (approved):
This looks great. The implementation could be cleaned up slightly by implementing `From<ThresholdError> for DescriptorError`, wrapping it in the `Miniscript` variant, and propagating the result of `Threshold::new` directly using `?`.

busayo-OD:
> This looks great. The implementation could be cleaned up slightly by implementing `From<ThresholdError> for DescriptorError`, wrapping it in the `Miniscript` variant, and propagating the result of `Threshold::new` directly using `?`.

Thanks for the review. Done!

PR #534, opened 2026-08-15
feat: Add lock_utxos option to prevent double-spending unbroadcasted txs

dorinengabdoh:
### Description

Title: feat: Add lock_utxos option to prevent double-spending created transactions

Description: This PR addresses #40 by introducing an option to automatically lock the selected UTXOs (inputs) of a transaction when it is created.

Added lock_utxos option (disabled by default) to TxParams and PsbtParams.
Implemented lock_utxos(bool) builder methods on TxBuilder and PsbtParams.
Updated Wallet::create_tx and Wallet::create_psbt_with_rng to call lock_outpoint on the transaction inputs if lock_utxos is enabled.
Verification: Added the following integration tests to verify that locked UTXOs are excluded from subsequent transaction creations:

test_tx_builder_lock_utxos in tests/wallet.rs
test_create_psbt_lock_utxos in tests/create_psbt.rs
Run tests with: RUSTFLAGS="--cfg bdk_wallet_unstable" cargo test

### DIFF
--- a/src/wallet/mod.rs
+++ b/src/wallet/mod.rs
@@ -1572,6 +1572,7 @@ impl Wallet {
         // Sort inputs/outputs according to the chosen algorithm.
         params.ordering.sort_tx_with_aux_rand(&mut tx, rng);
 
+        let lock_utxos = params.lock_utxos;
         let psbt = self.complete_transaction(tx, coin_selection.selected, params)?;
 
         // Recording changes to the change keychain.
@@ -1581,6 +1582,12 @@ impl Wallet {
             {
                 self.stage.merge(index_changeset.into());
                 self.mark_used(keychain, index);
+            }
+        }
+
+        if lock_utxos {
+            for input in &psbt.unsigned_tx.input {
+                self.lock_outpoint(input.previous_output);
             }
         }
 
@@ -3267,6 +3274,12 @@ impl Wallet {
                 {
                     self.stage.merge(index_changeset.into());
                 }
+            }
+        }
+
+        if params.lock_utxos {
+            for input in &psbt.unsigned_tx.input {
+                self.lock_outpoint(input.previous_output);
             }
         }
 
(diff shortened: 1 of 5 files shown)


dorinengabdoh:
Hi everyone, I've worked on a fix for this issue. Since BDK 3.0 already supports persistent UTXO locking via lock_outpoint, I have opened a Pull Request that introduces a lock_utxos(bool) option on both TxBuilder and PsbtParams (disabled by default to maintain backward compatibility).

When enabled, it automatically locks the selected inputs of the generated transaction so that subsequent transaction creations do not double-spend them. I've also added integration tests to verify this behavior. I would love to get your feedback on the PR!

busayo-OD (commented):
This could serve as a convenient wrapper around lock_outpoint, applying the lock automatically when creating the transaction rather than leaving it to the caller. I also noticed there was an earlier design discussion around this in #257.

Since locks don't expire automatically, what is the expected way to handle the case where the created transaction is abandoned or never broadcast? Is the caller expected to explicitly call unlock_outpoint in that case? If so, it might be worth documenting.

src/wallet/mod.rs
```diff
@@ -1584,6 +1585,12 @@ impl Wallet {
             }
         }
 
+        if lock_utxos {
+            for input in &psbt.unsigned_tx.input {
+                self.lock_outpoint(input.previous_output);
+            }
+        }
```
aagbotemi, line 1592: This locks every input, including foreign UTXOs. `lock_outpoint` is documented as locking a wallet output, so worth filtering to outputs the wallet actually tracks first.

src/wallet/mod.rs
```diff
@@ -3270,6 +3277,12 @@ impl Wallet {
             }
         }
 
+        if params.lock_utxos {
+            for input in &psbt.unsigned_tx.input {
+                self.lock_outpoint(input.previous_output);
+            }
+        }
```
aagbotemi, line 3284: Same here for filtering wallet owned UTXOs.

aagbotemi (commented):
Thanks for working on this. I left some comments.

This PR doesn't build under `RUSTFLAGS="--cfg bdk_wallet_unstable" cargo build --all-features`: `into_replace_params` constructs `PsbtParams<ReplaceTx>` field by field, so the new field (`lock_utxos`) has to be listed there too.

PR #537, opened 2026-08-21
fix(ci): pin nightly toolchain for rustfmt

### LINKED ISSUE #535: [cont_integration] `fmt` job should pin the nightly compiler
### Problem

`rustfmt.toml` sets options that are only supported on the nightly channel:

```toml
comment_width = 100
format_code_in_doc_comments = true
wrap_comments = true
```

The `justfile`'s `fmt` and `check` recipes correctly invoke `cargo +nightly fmt`, but the CI `fmt` job in `.github/workflows/cont_integration.yml` does not:

```yaml
fmt:
  name: Rust fmt
  runs-on: ubuntu-latest
  steps:
    ...
    - name: Check fmt
      run: cargo fmt --all --check
```

This uses whatever toolchain `rust-toolchain.toml` resolves to (currently stable `1.96.0`), not nightly. Stable `rustfmt` ignores unknown config keys, so the `fmt` job passes even when the file doesn't actually conform to `wrap_comments`/`format_code_in_doc_comments`. See a recent run https://github.com/bitcoindevkit/bdk_wallet/actions/runs/32433625759/job/96630281726.

CI's `fmt` does not enforce the same formatting rules that `just p` enforces locally, so PRs can pass CI with formatting that a contributor running `just p` would have to reformat.

### Second, related problem

Even after pointing CI at `+nightly`, `cargo +nightly fmt` resolves to whatever nightly happens to be installed at the time, with no version pin. Two contributors on different nightly snapshots can get different "correctly formatted" output for the same source, causing formatting churn and back-and-forth in PRs.

### Suggested fix

- Pin the nightly compiler in both `cont_integration.yml` and `justfile`, and periodically bump the pinned version alongside other toolchain bumps

Arowolokehinde:
Fixes #535

### Description

The CI fmt job ran cargo fmt on the stable toolchain resolved from rust-toolchain.toml, so the nightly-only options in rustfmt.toml (wrap_comments, comment_width, format_code_in_doc_comments) were silently ignored and never enforced. PRs could pass CI with formatting that `just p` would reject locally.

Point the fmt job at a pinned nightly (nightly-2026-08-01) with the rustfmt component, and pin the same version in the justfile recipes so CI and local just runs produce identical formatting output.

### Before submitting

### DIFF
--- a/.github/workflows/cont_integration.yml
+++ b/.github/workflows/cont_integration.yml
@@ -124,13 +124,15 @@ jobs:
         uses: actions/checkout@v7
         with:
           persist-credentials: false
-      # This action automatically reads and applies rust-toolchain.toml
+      # Pin nightly for rustfmt — rustfmt.toml uses nightly-only options
       - name: Install Rust toolchain
         uses: actions-rust-lang/setup-rust-toolchain@v1
         with:
+          toolchain: nightly-2026-08-01
+          components: rustfmt
           cache: true
       - name: Check fmt
-        run: cargo fmt --all --check
+        run: cargo +nightly-2026-08-01 fmt --all --check
 
   clippy_check:
     name: Rust clippy
--- a/justfile
+++ b/justfile
@@ -19,14 +19,14 @@ build:
 
 # Check formatting, compilation, linting
 _check:
-   cargo +nightly fmt --all -- --check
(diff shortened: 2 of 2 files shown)


ValuedMammal (approved):
ACK d43f63949d0453303db46414218e3b6811ce6655

PR #541, opened 2026-09-02
feat(wallet): add TxBuilder::avoid_reuse to skip reused-address UTXOs

tvpeter:
### Description

__Problem:__

An adversary can attack a wallet's privacy through forced address reuse: after observing one of the addresses the user have already spent from, they can send small outputs to it. If the users' wallet later select those coins into a transaction, the adversary learns which inputs the user controls and the destinations they pay to, linking their UTXOs.

The TxBuilder has no way to automatically prevent spending outputs in an address that has already been spent from. This is one of the open privacy items tracked in #28, and mirrors Bitcoin Core's `avoid_reuse` wallet flag ([bitcoin/bitcoin#13756](https://github.com/bitcoin/bitcoin/pull/13756)).

__Approach:__

Add an opt-in transaction builder method:

```rust
builder
    .add_recipient(addr.script_pubkey(), Amount::from_sat(20_000))
    .avoid_reuse();
let psbt = builder.finish()?;
```

__Tradeoff:__

- An address that received change and was later used to receive outputs again will be flagged, and its outputs excluded from automatic selection (they remain spendable via add_utxo()). This can lead to InsufficientFunds when the only funds available are on reused addresses.

### Notes to the reviewers

- An address (scriptPubKey) is treated as reused when the wallet has indexed more than one output to it. Since each (keychain, derivation_index) maps 1:1 to a scriptPubKey, outputs are counted per index from spk_index().outpoints() (which includes already-spent outputs, so the spend then dust attack is caught).
- Every UTXO on a reused address is added to the existing unspendable set. They can still be spent when selected explicitly via `add_utxo()`.

### Changelog notice

- Add `TxBuilder::avoid_reuse` to prevent TxBuilder from selecting reused address UTXOs

### Before submitting

### DIFF
--- a/src/wallet/tx_builder.rs
+++ b/src/wallet/tx_builder.rs
@@ -619,6 +619,50 @@ impl<'a, Cs> TxBuilder<'a, Cs> {
         self.exclude_below_confirmations(1)
     }
 
+    /// Avoid spending UTXOs that sit on an address the wallet has used more than once.
+    ///
+    /// When called, every currently-unspent UTXO whose address (script pubkey) appears in more
+    /// than one of the wallet's indexed outputs is added to the "unspendable" list (see
+    /// [`TxBuilder::unspendable`]), so automatic coin selection will not pick it. This improves
+    /// privacy by thwarting *forced address reuse* — e.g. an adversary peppering an already-used
+    /// address of yours with dust, hoping your wallet later merges it into a payment and thereby
+    /// links your UTXOs together.
+    ///
+    /// Reused-address UTXOs can still be spent by selecting them explicitly with
+    /// [`TxBuilder::add_utxo`], since manual selection overrides the unspendable list.
+    ///
+    /// This inspects the wallet's current state, so call it after syncing. Chaining it with other
+    /// filtering methods yields the union of all excluded outpoints.
+    ///
+    /// This mirrors Bitcoin Core's `avoid_reuse` wallet flag (`bitcoin/bitcoin#13756`).
+    pub fn avoid_reuse(&mut self) -> &mut Self {
+        // Count the wallet's indexed outputs (spent or unspent) per address. Each
+        // `(keychain, derivation index)` maps 1:1 to a script pubkey, so a count > 1 means the
+        // address received funds more than once, i.e. it was reused.
+        let mut output_counts: HashMap<(KeychainKind, u32), usize> = HashMap::new();
+        for &((keychain, index), _) in self.wallet.spk_index().outpoints() {
+            *output_counts.entry((keychain, index)).or_default() += 1;
+        }
+
+        let to_exclude = self
+            .wallet
+            .list_unspent()
+            .filter(|utxo| {
+                output_counts
+                    .get(&(utxo.keychain, utxo.derivation_index))
+                    .copied()
+                    .unwrap_or(0)
+                    > 1
+            })
+            .map(|utxo| utxo.outpoint)
+            .collect::<Vec<_>>();
+
+        for outpoint in to_exclude {
+            self.params.unspendable.insert(outpoint);
+        }
+        self
+    }
+
     /// Sign with a specific sig hash
     ///
     /// **Use this option very carefully**
--- a/src/wallet/tx_builder.rs
+++ b/src/wallet/tx_builder.rs
@@ -1071,9 +1071,11 @@ mod test {
     }
 
     use crate::test_utils::*;
-    use bitcoin::TxOut;
+    use alloc::string::ToString;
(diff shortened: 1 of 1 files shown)


src/wallet/tx_builder.rs
```diff
@@ -619,6 +619,50 @@ impl<'a, Cs> TxBuilder<'a, Cs> {
...
+    /// This inspects the wallet's current state, so call it after syncing. Chaining it with other
+    /// filtering methods yields the union of all excluded outpoints.
+    ///
+    /// This mirrors Bitcoin Core's `avoid_reuse` wallet flag (`bitcoin/bitcoin#13756`).
+    pub fn avoid_reuse(&mut self) -> &mut Self {
+        // Count the wallet's indexed outputs (spent or unspent) per address. Each
+        // `(keychain, derivation index)` maps 1:1 to a script pubkey, so a count > 1 means the
+        // address received funds more than once, i.e. it was reused.
+        let mut output_counts: HashMap<(KeychainKind, u32), usize> = HashMap::new();
+        for &((keychain, index), _) in self.wallet.spk_index().outpoints() {
```
Dmenec, line 643: `wallet.list_output()` may fit better here. It is canonical and includes both spent and unspent outputs. 

I think in any case it should be canonicalized, for cases like someone paying you and bumping fees, which right now would be seen as a double payment to this address, excluding your coin.

Maybe we could add some tests trying replaced payments.

src/wallet/tx_builder.rs
```diff
@@ -1614,4 +1616,81 @@ mod test {
...
+        // reused coins remain spendable when selected explicitly
+        let mut builder = wallet.build_tx();
+        builder.add_recipient(recipient.script_pubkey(), Amount::from_sat(80_000));
+        builder.avoid_reuse();
+        builder
+            .add_utxo(reused_outpoint)
+            .unwrap()
+            .add_utxo(reused_outpoint_2)
+            .unwrap();
+        assert!(builder.finish().is_ok());
```
Dmenec, line 1694: nit: could assert the reused coins are actually selected

```suggestion
        let psbt = builder.finish().unwrap();
        let selected: Vec<OutPoint> = psbt
            .unsigned_tx
            .input
            .iter()
            .map(|i| i.previous_output)
            .collect();
        assert!(selected.contains(&reused_outpoint));
        assert!(selected.contains(&reused_outpoint_2));
```

Dmenec (commented):
cACK, nice feature :)

AFAIK Core only flags an address as used once the wallet has spent from it from then on, any coin sent to it is avoided. The PR description describes it the same way, but the implementation flags any address with more than one received output, even if it was never spent from.

Small test showing it:

```rust
let addr = wallet.reveal_next_address(KeychainKind::External).address;

// sent 2 outputs to the same address
receive_output_to_address(&mut wallet, addr.clone(), Amount::from_sat(100_000), ReceiveTo::Mempool(0));
receive_output_to_address(&mut wallet, addr, Amount::from_sat(546), ReceiveTo::Mempool(0));

builder.add_recipient(recipient.script_pubkey(), Amount::from_sat(10_000)).avoid_reuse();

// the 100k sats are excluded
assert!(matches!(
    builder.finish(),
    Err(CreateTxError::CoinSelection(e)) if e.available == Amount::ZERO
));
```

Not sure if this was intentional. If it was, I'd document that it differs from Core. I would follow Core's approach as anyone could grief UTXOs knowing this feature.

Chibey-max:
Should “reused address” mean “received more than one output,” or should it mean “the wallet has already spent from this address, and then later received more coins there”?

tvpeter:
> AFAIK Core only flags an address as used once the wallet has spent from it from then on, any coin sent to it is avoided. The PR description describes it the same way, but the implementation flags any address with more than one received output, even if it was never spent from.

Thank you so much for this review. I have updated the approach to use the wallet's canonical history (`Wallet::list_output`) to know an address that has been spent from. I have also updated the test cases to show that a never-spent from address is not excluded even if it receives outputs more than once; a spent-from address is excluded (spendable via explicit selection), and a fee-bumped transaction is not counted as a reuse.

tvpeter:
The CI is failing because of a transitive dependency `yoke-derive`. The recent updated version not been compatible with rustc 1.85. This was fixed on `release/3.x` branch by #568. I will duplicate the fix in another PR.

tvpeter:
> The CI is failing because of a transitive dependency `yoke-derive`. The recent updated version not been compatible with rustc 1.85. This was fixed on `release/3.x` branch by #568. I will duplicate the fix in another PR.

Minor PR to pin yoke-derive #572

PR #544, opened 2026-09-06
fix(wallet): reject txs exceeding MAX_STANDARD_TX_WEIGHT

### LINKED ISSUE #543: Missing MAX_STANDARD_TX_WEIGHT check produces unrelayable txs
**Describe the bug**

Neither `create_tx` nor `create_psbt` path checks the final transaction weight against the 400,000 wu standardness limit. Output-side standardness is enforced (dust check, `OutputBelowDustLimit`).

A wallet with enough UTXOs can drain into a fully signed, finalized PSBT over 400,000 wu, which every standardness-enforcing mempool rejects. 

**To Reproduce**

1. Fund a wallet with ~1,500 small P2WPKH UTXOs.
2. `build_tx().drain_wallet().drain_to(addr).finish()` → `Ok(psbt)`; `wallet.sign(...)` → `Ok(true)`. Extracted tx ≈406,000 wu; broadcast rejected.

**Expected behavior**

After assembling the transaction (`complete_transaction` and `create_psbt` path), reject when final weight exceeds `MAX_STANDARD_TX_WEIGHT` with a dedicated error, symmetric to the existing `OutputBelowDustLimit` check.

**Build environment**
- BDK tag/commit: `49f7955c` (master, ≤ 3.1.0)

- [x] None / not backend-related

- [x] No

**Additional context**

Availability/UX footgun: An oversized tx cannot enter a mempool, but returning `Ok(true)` from signing is a misleading success signal.

cestercian:
### Description

Neither `create_tx` nor `create_psbt` checked the assembled transaction against Bitcoin's standardness weight limit (`bitcoin::policy::MAX_STANDARD_TX_WEIGHT`, 400_000 WU). Dust was already rejected (`OutputBelowDustLimit`); weight was not.

A drain of many small UTXOs could therefore produce a fully signed PSBT over 400k WU. Every standardness-enforcing mempool rejects that transaction, but `sign` still returned `Ok(true)` — a misleading success.

**Root cause:** After coin selection the unsigned transaction is assembled with empty witnesses. `Transaction::weight()` on that value undercounts the final signed size by each input's satisfaction (witness / scriptSig) weight. No later check compared the estimated signed weight to the standardness limit.

**Fix:** After the unsigned tx is assembled, estimate the signed weight (`tx.weight()` plus each input's satisfaction weight, plus the 2-WU segwit marker when witnesses will be present) and reject when it exceeds `MAX_STANDARD_TX_WEIGHT`.

- New `CreateTxError::TxWeightLimitExceeded { weight, limit }` (stable `create_tx` path)
- New `CreatePsbtError::TxWeightLimitExceeded { weight, limit }` (unstable `create_psbt` / `create_psbt_from_selector` path, so RBF is covered too)

**Tests:**
- Drain of 1,500 small P2WPKH UTXOs (the reported case) via a single in-memory funding transaction — no Electrum/Esplora/RPC
- Unit-level foreign UTXO whose satisfaction weight alone exceeds the limit
- Matching `create_psbt` drain regression

Fixes #543

### Notes to the reviewers

- This is a **breaking change**: `CreateTxError` is exhaustive, so downstream `match`es must handle the new variant. `CreatePsbtError` is already `#[non_exhaustive]`.
- The check uses estimated signed weight, not raw unsigned `tx.weight()`. Checking only the unsigned weight would miss the 1,500-P2WPKH-input case (~246k WU unsigned vs ~408k WU signed).
- Satisfaction weights for `create_tx` come from the `WeightedUtxo`s used in coin selection (including foreign UTXOs). The `create_psbt` path looks up local descriptors via `max_weight_to_satisfy`.

### Changelog notice

- Fixed `create_tx` / `create_psbt` producing unrelayable transactions over `MAX_STANDARD_TX_WEIGHT` by returning `TxWeightLimitExceeded`.

### Before submitting

- [x] This PR breaks the existing API

Assistance: implementation drafted with Cursor (cloud agent); author is Cestercian. Commits are SSH-signed; GitHub may show Unverified if the cloud-agent SSH signing key is not registered on the account as a signing key.

### DIFF
--- a/src/wallet/mod.rs
+++ b/src/wallet/mod.rs
@@ -1505,6 +1505,12 @@ impl Wallet {
             }
         };
 
+        let satisfaction_by_outpoint: HashMap<OutPoint, Weight> = required_utxos
+            .iter()
+            .chain(optional_utxos.iter())
+            .map(|weighted| (weighted.utxo.outpoint(), weighted.satisfaction_weight))
+            .collect();
+
         // Get drain script.
         let mut drain_index = Option::<(KeychainKind, u32)>::None;
         let drain_script = match params.drain_to {
@@ -1598,6 +1604,18 @@ impl Wallet {
 
         // Sort inputs/outputs according to the chosen algorithm.
         params.ordering.sort_tx_with_aux_rand(&mut tx, rng);
+
+        // Reject transactions that would exceed the standardness weight limit once signed.
+        // `tx.weight()` is the unsigned weight (empty witnesses); add each input's satisfaction
+        // weight so we compare against the same limit mempools enforce on the final tx.
+        let satisfaction_weights = tx.input.iter().map(|txin| {
+            satisfaction_by_outpoint
+                .get(&txin.previous_output)
+                .copied()
+                .unwrap_or_else(|| self.satisfaction_weight_for_outpoint(txin.previous_output))
+        });
+        check_max_standard_tx_weight(&tx, satisfaction_weights)
+            .map_err(|(weight, limit)| CreateTxError::TxWeightLimitExceeded { weight, limit })?;
 
         let psbt = self.complete_transaction(tx, coin_selection.selected, params)?;
 
@@ -2944,6 +2962,48 @@ impl Wallet {
             })
             .collect()
     }
+
+    /// Satisfaction weight used to estimate the signed size of spending `outpoint`.
+    ///
+    /// Local wallet UTXOs use the descriptor's [`max_weight_to_satisfy`]. Foreign or unknown
+    /// outpoints return [`Weight::ZERO`], which underestimates; callers that still have the
+    /// original [`WeightedUtxo`] should prefer that value.
+    ///
+    /// [`max_weight_to_satisfy`]: miniscript::Descriptor::max_weight_to_satisfy
+    fn satisfaction_weight_for_outpoint(&self, outpoint: OutPoint) -> Weight {
+        self.get_utxo(outpoint)
+            .and_then(|utxo| {
+                self.public_descriptor(utxo.keychain)
+                    .max_weight_to_satisfy()
+                    .ok()
+            })
+            .unwrap_or(Weight::ZERO)
+    }
+}
+
+/// Estimate the signed weight of an assembled unsigned transaction and reject it when it
+/// exceeds [`bitcoin::policy::MAX_STANDARD_TX_WEIGHT`].
+///
+/// `tx` is unsigned (empty witnesses). `satisfaction_weights` are the additional scriptSig /
+/// witness weights for each input. Empty-witness transactions serialize without the 2-WU
+/// segwit marker/flag; that is added once any satisfaction weight is present.
+fn check_max_standard_tx_weight(
+    tx: &Transaction,
+    satisfaction_weights: impl IntoIterator<Item = Weight>,
+) -> Result<(), (Weight, Weight)> {
+    let satisfaction: Weight = satisfaction_weights.into_iter().sum();
+    let segwit_marker = if satisfaction > Weight::ZERO {
+        Weight::from_wu(2)
+    } else {
+        Weight::ZERO
+    };
+    let weight = tx.weight() + satisfaction + segwit_marker;
+    let limit = Weight::from_wu(u64::from(bitcoin::policy::MAX_STANDARD_TX_WEIGHT));
+    if weight > limit {
+        Err((weight, limit))
+    } else {
+        Ok(())
+    }
 }
 
 /// Methods to construct sync/full-scan requests for spk-based chain sources.
@@ -3220,6 +3280,7 @@ impl Wallet {
     /// - A manually selected input is missing from the wallet, or could not be planned
     /// - The input value is insufficient to fund the outputs
     /// - Failure to complete coin selection
+    /// - The assembled transaction exceeds [`bitcoin::policy::MAX_STANDARD_TX_WEIGHT`]
     /// - Failure to create or update the PSBT.
     ///
     /// # Change address
@@ -3423,6 +3484,14 @@ impl Wallet {
                 rng,
             )
             .map_err(CreatePsbtError::Psbt)?;
+
+        let satisfaction_weights = psbt
+            .unsigned_tx
+            .input
+            .iter()
+            .map(|txin| self.satisfaction_weight_for_outpoint(txin.previous_output));
+        check_max_standard_tx_weight(&psbt.unsigned_tx, satisfaction_weights)
+            .map_err(|(weight, limit)| CreatePsbtError::TxWeightLimitExceeded { weight, limit })?;
 
         // Add global xpubs.
         if params.add_global_xpubs {
--- a/src/wallet/error.rs
+++ b/src/wallet/error.rs
@@ -21,7 +21,7 @@ use alloc::{
 };
 #[cfg(all(bdk_wallet_unstable, feature = "bdk-tx"))]
 use bdk_tx::bdk_coin_select;
-use bitcoin::{Amount, BlockHash, Network, OutPoint, Sequence, Txid, absolute, psbt};
+use bitcoin::{Amount, BlockHash, Network, OutPoint, Sequence, Txid, Weight, absolute, psbt};
 use core::fmt;
 
 /// The error type when loading a [`Wallet`] from a [`ChangeSet`].
@@ -199,6 +199,17 @@ pub enum CreateTxError {
     NoUtxosSelected,
     /// Output created is under the dust limit, 546 satoshis
     OutputBelowDustLimit(usize),
+    /// Assembled transaction exceeds [`bitcoin::policy::MAX_STANDARD_TX_WEIGHT`].
+    ///
+    /// The reported `weight` is the estimated signed weight (unsigned transaction plus each
+    /// input's satisfaction weight). Standard mempools reject transactions above this limit,
+    /// so building them is treated as an error.
+    TxWeightLimitExceeded {
+        /// Estimated signed transaction weight.
+        weight: Weight,
+        /// Standardness weight limit that was exceeded.
+        limit: Weight,
+    },
     /// There was an error with coin selection
     CoinSelection(coin_selection::InsufficientFunds),
     /// Cannot build a tx without recipients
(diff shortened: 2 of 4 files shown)


src/wallet/mod.rs
```diff
@@ -3424,6 +3485,14 @@ impl Wallet {
             )
             .map_err(CreatePsbtError::Psbt)?;
 
+        let satisfaction_weights = psbt
+            .unsigned_tx
+            .input
+            .iter()
+            .map(|txin| self.satisfaction_weight_for_outpoint(txin.previous_output));
+        check_max_standard_tx_weight(&psbt.unsigned_tx, satisfaction_weights)
```
j-kon: In `create_psbt`, satisfaction weights are resolved solely via `self.satisfaction_weight_for_outpoint(txin.previous_output)`:

```rust
fn satisfaction_weight_for_outpoint(&self, outpoint: OutPoint) -> Weight {
    self.get_utxo(outpoint)
        .and_then(|utxo| {
            self.public_descriptor(utxo.keychain)
                .max_weight_to_satisfy()
                .ok()
        })
        .unwrap_or(Weight::ZERO)
}
```

For foreign or planned inputs (e.g. added via `PsbtParams::add_planned_input` or coin-selected from external sources), `self.get_utxo` returns `None`, so their satisfaction weight evaluates to `Weight::ZERO`.

As a result, `create_psbt` completely discounts satisfaction weights for non-local inputs. A PSBT spending foreign/planned inputs whose signed size would exceed `MAX_STANDARD_TX_WEIGHT` passes with `Ok`, whereas the equivalent transaction in `create_tx` correctly fails with `TxWeightLimitExceeded`.

**Reproduction:**
Adding a planned/foreign input with satisfaction weight exceeding 400,000 WU to `PsbtParams` results in `wallet.create_psbt(params)` returning `Ok(psbt)` instead of `Err(CreatePsbtError::TxWeightLimitExceeded)`.

**Suggestion:**
Consider retrieving the planned satisfaction weights from the selection candidates (e.g., `selection.inputs()`, where `bdk_tx::Input::satisfaction_weight()` is already tracked) rather than querying `self.satisfaction_weight_for_outpoint`.

src/wallet/mod.rs
```diff
@@ -2944,6 +2962,48 @@ impl Wallet {
...
+/// `tx` is unsigned (empty witnesses). `satisfaction_weights` are the additional scriptSig /
+/// witness weights for each input. Empty-witness transactions serialize without the 2-WU
+/// segwit marker/flag; that is added once any satisfaction weight is present.
+fn check_max_standard_tx_weight(
+    tx: &Transaction,
+    satisfaction_weights: impl IntoIterator<Item = Weight>,
+) -> Result<(), (Weight, Weight)> {
+    let satisfaction: Weight = satisfaction_weights.into_iter().sum();
+    let segwit_marker = if satisfaction > Weight::ZERO {
+        Weight::from_wu(2)
```
j-kon: The witness overhead calculation has two accounting discrepancies with BIP 141 serialization:

1. **Pure legacy overestimation (+2 WU):**
   `satisfaction > Weight::ZERO` is true for legacy transactions because `scriptSig` satisfaction weights are positive (e.g., 428 WU for P2PKH). This adds 2 WU for a SegWit marker/flag (`0x0001`) that is never serialized in legacy transactions.

2. **SegWit & Mixed underestimation (-1 WU per input):**
   When `tx` is unsigned, all input witnesses are empty, so `tx.weight()` evaluates as a legacy transaction with zero witness bytes.
   Under BIP 141 extended serialization, every input requires a 1-byte witness stack length varint (`0x00`). Miniscript's `max_weight_to_satisfy()` returns `txin.segwit_weight() - TxIn::default().segwit_weight()`, where `TxIn::default().segwit_weight()` already assumed that 1-byte empty witness was present. Adding only `+ 2 WU` misses 1 WU for every input ($K$ WU total).

On transactions with many inputs (such as UTXO consolidation), this systematically underestimates the worst-case signed weight by $K$ WU:

| Case | Inputs ($K$) | Unsigned | Satisfaction | PR Estimate | Actual Worst-Case Signed | Discrepancy |
| :--- | :--- | :--- | :--- | :--- | :--- | :--- |
| Pure P2PKH | 1 | 340 WU | 428 WU | 770 WU | 768 WU | **+2 WU** (overcounted) |
| Mixed (1 WPKH + 2 PKH) | 3 | 656 WU | 963 WU | 1,621 WU | 1,624 WU | **-3 WU** (undercounted) |
| Pure P2WPKH | 100 | 16,564 WU | 10,700 WU | 27,266 WU | 27,366 WU | **-100 WU** (undercounted) |

**Suggestion:**
Consider differentiating transactions where witnesses will be present:
- If no inputs have witness satisfaction (pure legacy): witness overhead is `0 WU`.
- If any input has witness satisfaction (SegWit): witness overhead is `2 WU + (1 WU * tx.input.len())`.

src/wallet/mod.rs
```diff
@@ -2944,6 +2962,48 @@ impl Wallet {
...
+/// exceeds [`bitcoin::policy::MAX_STANDARD_TX_WEIGHT`].
+///
+/// `tx` is unsigned (empty witnesses). `satisfaction_weights` are the additional scriptSig /
+/// witness weights for each input. Empty-witness transactions serialize without the 2-WU
+/// segwit marker/flag; that is added once any satisfaction weight is present.
+fn check_max_standard_tx_weight(
+    tx: &Transaction,
+    satisfaction_weights: impl IntoIterator<Item = Weight>,
+) -> Result<(), (Weight, Weight)> {
+    let satisfaction: Weight = satisfaction_weights.into_iter().sum();
```
j-kon: `bitcoin::Weight` wraps `u64` and implements `Add` and `Sum` via primitive integer arithmetic without overflow protection. If foreign inputs have large caller-supplied satisfaction weights (via `TxBuilder::add_foreign_utxo`), this arithmetic overflows:

- In debug builds (`cargo test`), `tx.weight() + satisfaction` panics with `attempt to add with overflow` instead of returning `CreateTxError::TxWeightLimitExceeded`.
- In release builds (where `overflow-checks = false`), the addition silently wraps around modulo $2^{64}$. A transaction with an astronomical satisfaction weight wraps to a small number, passing `weight <= MAX_STANDARD_TX_WEIGHT` and completely bypassing the check.

**Reproduction:**
```rust
builder.add_foreign_utxo(outpoint, psbt_in, Weight::from_wu(u64::MAX - 200)).unwrap();
builder.manually_selected_only();
builder.fee_absolute(Amount::from_sat(1000));
builder.drain_wallet().drain_to(addr.script_pubkey());

let res = builder.finish();
```
- Under `cargo test`, this panics with `attempt to add with overflow` at line 3000.
- Under `cargo test --release`, unsigned weight 328 WU + $(2^{64} - 201)\text{ WU} + 2\text{ WU}$ wraps to 129 WU, returning `Ok(psbt)` and bypassing the standardness limit.

**Suggestion:**
Use checked arithmetic (`satisfaction_weights.into_iter().try_fold(Weight::ZERO, Weight::checked_add)` and `checked_add`), and return `Err((Weight::MAX, limit))` if an overflow occurs.

j-kon (changes requested):
Thanks for tackling issue #543! Preventing wallets from building transactions that standard mempools will inevitably reject is an important safety improvement.

While the high-level intent and the tests for local P2WPKH drains are solid, our line-by-line review and test verification identified three concrete issues in the implementation that need to be addressed before this is ready to merge:

1. **Foreign / planned input satisfaction weights are dropped in `create_psbt`:**
   In `src/wallet/mod.rs:3488-3494`, `create_psbt` resolves satisfaction weights solely via `self.satisfaction_weight_for_outpoint`. For foreign or planned inputs, this returns `Weight::ZERO`, meaning the standardness check underestimates transactions containing external or planned inputs.
2. **Witness serialization discrepancies in `check_max_standard_tx_weight`:**
   In `src/wallet/mod.rs:2995-3000`:
   - **Pure legacy transactions (+2 WU):** Any non-zero satisfaction weight (including legacy `scriptSig` satisfaction like 428 WU for P2PKH) triggers `+ 2 WU` for SegWit marker/flag, which are never serialized.
   - **SegWit & Mixed transactions (-1 WU per input):** When an unsigned transaction has empty witnesses, `tx.weight()` evaluates as a legacy transaction with zero witness bytes. Under BIP 141 extended serialization, every input requires a 1-byte witness stack length varint (`0x00`). Miniscript's `max_weight_to_satisfy()` returns the delta from `TxIn::default().segwit_weight()` (which already assumed that 1 byte was present). Adding only `+ 2 WU` fails to account for this 1 byte per input ($K$ WU total), underestimating multi-input SegWit and mixed transactions.
3. **Unchecked arithmetic leads to debug panic and release-mode validation bypass:**
   In `src/wallet/mod.rs:2994-3000`, `satisfaction_weights.into_iter().sum()` and `tx.weight() + satisfaction + segwit_marker` use unchecked `Weight` addition. When foreign UTXOs have large satisfaction weights:
   - In debug builds, it panics with `attempt to add with overflow` instead of returning `TxWeightLimitExceeded`.
   - In release builds (`overflow-checks = false`), the addition silently wraps around modulo $2^{64}$. A transaction with an astronomical satisfaction weight wraps to a small number (e.g. 129 WU), passing `weight <= MAX_STANDARD_TX_WEIGHT` and completely bypassing the policy check.

Detailed suggestions and minimal reproductions are provided in the inline comments below.

src/wallet/mod.rs
cestercian: create_psbt now takes satisfaction_weight() from the selected inputs, so planned/foreign ones are not zeroed out anymore.

src/wallet/mod.rs
cestercian: Yeah — pure legacy adds no marker now; if any input has a witness, it's 2 WU plus 1 per input. Checked against 768 / 1,624 / 27,366 from your table.

src/wallet/mod.rs
cestercian: Switched to checked_add. Overflow comes back as TxWeightLimitExceeded with Weight::MAX instead of panicking or wrapping past the limit.

cestercian:
Thanks for the review — addressed all three.

Planned/foreign inputs keep their satisfaction weight from the selection, witness overhead now splits pure legacy (0) vs segwit/mixed (2 + 1 per input), and a huge foreign satisfaction returns TxWeightLimitExceeded instead of panicking or wrapping. The 1,500-input drains still fail the check; added coverage for the planned-input case, the overhead table, and the overflow path.

Issue #542, opened 2026-09-03
Wallet signing may finalize invalid pre-seeded signatures and return Ok(true)


noahjoeris:
**Describe the bug**

`Wallet::sign_with_signers` skips signing when the PSBT already has a signature in the wallet key's slot, without verifying it. With default `SignOptions` (`try_finalize: true`), `finalize_psbt` consumes that signature, builds the final witness, and returns `Ok(true)`, although the transaction fails script validation.

Skip sites in `src/wallet/signer.rs`:

- ECDSA: `partial_sigs.contains_key(&pubkey)` 
- Taproot key path: `tap_key_sig.is_none()` 
- Taproot script path: `tap_script_sigs.contains_key(...)`

**To Reproduce**

1. Create a single-input `wpkh` PSBT.
2. Insert an invalid parseable signature for the wallet pubkey into `partial_sigs`.
3. `wallet.sign_with_signers(&mut psbt, &signers, SignOptions::default())` → `Ok(true)`, invalid sig lands in `final_script_witness`.

Same for Taproot: pre-seed `tap_key_sig` or `tap_script_sigs`.

**Expected behavior**

Overwrite wallet-controlled slots (or verify and overwrite invalid ones) on all three paths. Never `Ok(true)` with an invalid final witness.

**Build environment**
- BDK tag/commit: `6fc68462` (master)
- Rust/Cargo version: 1.96.0

**Which backend(s) are relevant (if any)?**
- [ ] Electrum
- [ ] Esplora
- [ ] Bitcoin Core RPC
- [x] None / not backend-related (e.g. `bdk_chain`, `bdk_core`)
- [ ] Other (please specify): `____`

**Is this blocking production use?**
- [ ] Yes
- [x] No

**Additional context**

`finalize_psbt` checks signature availability, not validity. The `allow_all_sighashes` guard is also bypassed: it checks the PSBT-level `sighash_type` field, not the sighash byte in an existing `partial_sigs` entry.

Jennycj:
Hello, I was able to reproduce this bug.

I would like to take this on.

PR #546, opened 2026-09-12
Fix: Wallet signing may finalize invalid pre-seeded signatures

Naomi-Gift:
Resloves #542 

## Problem

`sign_input` skipped signing when the PSBT already had a signature, without checking if it was valid. This could make `finalize_psbt` return `Ok(true)` with an invalid signature.

This affected ECDSA, Taproot key path, and Taproot script path signing.

## Fix

Removed the checks that skipped existing wallet-controlled signature slots. The wallet now always creates and writes a fresh, verified signature.

The checks for already finalized inputs remain unchanged.

## Testing

Existing test suite passes with the guards removed. No new tests added in this PR.


Naomi-Gift:
@oleonardolima Please can you review?

busayo-OD (changes requested):
The fix looks good. It might be worth adding regression tests covering invalid pre-seeded signatures in the affected paths.

src/wallet/signer.rs
```diff
@@ -1205,4 +1197,348 @@ mod signers_container_tests {
...
+            sig.signature.as_ref(),
+            &[0u8; 64],
+            "dummy tap_key_sig was not replaced"
+        );
+    }
+
+    /// An input that already has `final_script_sig` set must be skipped entirely —
+    /// the signer must not touch it, even if `partial_sigs` contains a dummy entry.
+    #[test]
+    fn sign_input_skips_already_finalized_legacy_input() {
```
busayo-OD, line 1473: This test and the sign_input_skips_already_finalized_segwit_input test cover existing behavior rather than the invalid pre-seeded signature case, so I think they and their helpers can be left out of this PR. The second commit can also be squashed into the first since it only fixes changes introduced there.

src/wallet/signer.rs
```diff
@@ -1205,4 +1197,348 @@ mod signers_container_tests {
...
+            ..Default::default()
+        };
+
+        (psbt, priv_key, pubkey)
+    }
+
+    /// A pre-seeded invalid ECDSA signature in `partial_sigs` must be replaced
+    /// by a valid signature after `sign_input` runs on a Legacy (P2PKH) input.
+    #[test]
+    fn sign_input_legacy_overwrites_invalid_preseeded_sig() {
```
busayo-OD, line 1354: This unit test, the sign_input_segwitv0_overwrites_invalid_preseeded_sig test, and the sign_input_tap_key_spend_overwrites_invalid_preseeded_sig test duplicate the regression coverage provided by the sign_with_signers tests, so I think they can be left out of this PR.

tests/wallet.rs
```diff
@@ -1547,42 +1547,171 @@ fn test_sign_with_signers() {
 }
 
 #[test]
-fn test_sign_single_xprv_with_master_fingerprint_and_path() {
```
busayo-OD, line 1550: Could we keep the existing test and add the regression test separately? Same for test_sign_single_xprv_bip44_path, since both are existing coverage unrelated to this fix.

busayo-OD (changes requested):
Thanks for the test coverage. I have a few comments on the test changes.

PR #564, opened 2026-09-16
fix(psbt): return None from fee_amount on overflow or negative fee

### LINKED ISSUE #558: `PsbtUtils::fee_amount` panics when PSBT outputs exceed inputs
**Describe the bug**

`PsbtUtils::fee_amount` (`src/psbt/mod.rs`) computes the fee as `input_amount.checked_sub(output_amount).expect("input amount must be greater than output amount")`. When the PSBT's outputs sum to more than its inputs, `checked_sub` returns `None` and the `expect` panics. `fee_rate` delegates to `fee_amount`, so it panics too.

The trait docs only say `None` is returned when a `TxOut` is missing for an input; there is no documented precondition that outputs are bounded by inputs. Since PSBTs are commonly received from other parties (coinjoin, payjoin, externally produced PSBTs), callers displaying a fee before signing can hit this on well-formed but inconsistent input.

This issue was found by AI.

**To Reproduce**

```rust
use bdk_wallet::psbt::PsbtUtils;
use bitcoin::{Amount, OutPoint, Psbt, ScriptBuf, Transaction, TxIn, TxOut, absolute, transaction};

#[test]
fn fee_amount_when_outputs_exceed_inputs() {
    let prev_tx = Transaction {
        version: transaction::Version::TWO,
        input: vec![TxIn::default()],
        output: vec![TxOut {
            value: Amount::from_sat(50_000),
            script_pubkey: ScriptBuf::default(),
        }],
    };
    let unsigned_tx = Transaction {
        version: transaction::Version::TWO,
        input: vec![TxIn {
            previous_output: OutPoint::new(prev_tx.compute_txid(), 0),
            ..Default::default()
        }],
        output: vec![TxOut {
            value: Amount::from_sat(90_000),
            script_pubkey: ScriptBuf::default(),
        }],
    };
    psbt.inputs[0].non_witness_utxo = Some(prev_tx);

    // panics at src/psbt/mod.rs:79: "input amount must be greater than output amount"
    let _ = psbt.fee_amount();
}
```

**Expected behavior**

`fee_amount` / `fee_rate` should not panic on a PSBT whose outputs exceed its inputs; the condition should be signalled to the caller (e.g. via the return value) or documented as a precondition.

**Build environment**

- [x] No

### LINKED ISSUE #559: `PsbtUtils::fee_amount` overflows when summing input or output amounts
**Describe the bug**

`PsbtUtils::fee_amount` (`src/psbt/mod.rs`) sums the input `TxOut` values and the output values with `Iterator::sum::<Amount>()`, which uses plain `u64` addition. If the values sum past `u64::MAX`, the addition overflows: with overflow checks enabled (debug/test builds) the call panics with `attempt to add with overflow`; in release builds it wraps silently and `fee_amount` returns a wrong fee (e.g. `Some(18446744073709550614 SAT)` for the reproducer below). `fee_rate` delegates to `fee_amount`, so it is affected too.

The input values come from the PSBT's `witness_utxo` / `non_witness_utxo`, which are plain `u64`s on decode (no `MAX_MONEY` check), so an externally produced PSBT can trigger this.

This is separate from #558: that one is about the `checked_sub(..).expect(..)` on the difference, whereas here the overflow happens while summing.

This issue was found by AI.

**To Reproduce**

```rust
use bdk_wallet::psbt::PsbtUtils;
use bitcoin::{Amount, Psbt, ScriptBuf, Transaction, TxIn, TxOut, absolute, transaction};

#[test]
fn fee_amount_when_input_amounts_overflow() {
    let unsigned_tx = Transaction {
        version: transaction::Version::TWO,
        input: vec![TxIn::default(), TxIn::default()],
        output: vec![TxOut {
            value: Amount::from_sat(1_000),
            script_pubkey: ScriptBuf::default(),
        }],
    };
    for input in &mut psbt.inputs {
        input.witness_utxo = Some(TxOut {
            value: Amount::from_sat(u64::MAX),
            script_pubkey: ScriptBuf::default(),
        });
    }

    // debug: panics with "attempt to add with overflow"
    // release: returns Some(18446744073709550614 SAT)
    let _ = psbt.fee_amount();
}
```

**Expected behavior**

`fee_amount` / `fee_rate` should not panic or return a wrapped value when the summed amounts overflow; the condition should be signalled to the caller (e.g. via the return value).

**Build environment**

- [x] No

shubham5080:
### Description

`PsbtUtils::fee_amount` panics or wraps on hostile/inconsistent PSBTs:

- outputs larger than inputs hit `checked_sub(..).expect(..)` (#558)
- summing input or output amounts past `u64::MAX` overflows `Iterator::sum` (#559)

`fee_rate` delegates to `fee_amount`, so it fails the same way. PSBTs often come from another party (coinjoin, payjoin, an external builder), so this is reachable before signing.

This change uses `checked_add` / `checked_sub` and returns `None` when the fee cannot be computed. Callers already treat `None` as "unknown fee".

Fixes #558
Fixes #559

### Notes to the reviewers

Kept using `get_utxo_for` rather than delegating to rust-bitcoin's `Psbt::fee()`, so the existing txid / vout checks on `non_witness_utxo` still apply.

Return type is unchanged (`Option<Amount>`). Invalid PSBTs that previously panicked now return `None`.

#### Fixed

- fix(psbt): `fee_amount` / `fee_rate` return `None` instead of panicking when outputs exceed inputs or when summing amounts overflows (#558, #559)

### Before submitting


src/psbt/mod.rs
```diff
@@ -167,4 +175,102 @@ mod tests {
...
+            }],
+            output: vec![
+                TxOut {
+                    value: Amount::from_sat(u64::MAX),
+                    script_pubkey: ScriptBuf::default(),
+                },
+                TxOut {
+                    value: Amount::from_sat(u64::MAX),
+                    script_pubkey: ScriptBuf::default(),
+                },
```
j-kon, line 264: In this test, the outputs (`[u64::MAX, u64::MAX]`) wrap modulo $2^{64}$ to $2^{64} - 2$. Because the input is $1,000$, the wrapped output sum still exceeds the input amount ($2^{64} - 2 > 1,000$).

Consequently, in release builds (or under wrapping arithmetic), this test passes even if the output `try_fold` checked addition is omitted, because `input_amount.checked_sub(output_amount)` catches it as `outputs > inputs`.

To ensure this test specifically isolates and protects the output summation overflow branch, consider choosing output values whose wrapped sum is less than or equal to the input amount (for example, an input of `1_000` sats with outputs `u64::MAX - 100` and `200` sats, which wrap to $99 \le 1,000$).

j-kon (commented):
Thanks for fixing this! The core arithmetic and negative fee fixes using checked operations cleanly resolve #558 and #559 across both debug and release profiles.

Left one suggestion on `fee_amount_returns_none_when_output_amounts_overflow` regarding test isolation against `outputs > inputs`.

PR #565, opened 2026-09-16
fix(wallet): treat unreachable relative locktimes as unsatisfied

### LINKED ISSUE #557: `finalize_psbt` panics with "Overflowing addition" when spending an unconfirmed UTXO of a descriptor with an `older()` branch
**Describe the bug**

`Wallet::finalize_psbt` maps unconfirmed previous transactions to a confirmation height of `u32::MAX` (`src/wallet/mod.rs`, `ChainPosition::Unconfirmed { .. } => (txid, u32::MAX)`). That value is later used as `create_height` in `Older::check_older` (`src/wallet/utils.rs`), which does `create_height.checked_add(n).expect("Overflowing addition")`. For any `older(n)` with `n > 0`, the addition overflows and the finalizer panics.

The `Older` satisfier is only consulted when the PSBT input's `nSequence` does not already satisfy the CSV branch (the tuple satisfier short-circuits on `PsbtInputSatisfier::check_older`). So this triggers when spending an unconfirmed UTXO of a descriptor that contains an `older()` branch which is *not* the one being used, e.g. spending via `pk(A)` of `or_d(pk(A),and_v(v:pk(B),older(144)))`. This is reachable through ordinary use (chaining a spend off an unconfirmed output), since `Wallet::sign` / `sign_with_signers` call `finalize_psbt` by default.

This issue was found by AI.

**To Reproduce**

```rust
use bdk_wallet::descriptor::{ExtractPolicy, policy::BuildSatisfaction};
use bdk_wallet::signer::{SignOptions, SignersContainer};
use bdk_wallet::test_utils::*;
use bdk_wallet::{KeychainKind, Wallet};
use bitcoin::{Amount, Network, Transaction, TxOut};

#[test]
fn finalize_psbt_spending_unconfirmed_utxo_with_csv_branch() {
    // wsh(or_d(pk(A),and_v(v:pk(B),older(144))))
    let mut wallet = Wallet::create_single(get_test_a_or_b_plus_csv())
        .network(Network::Regtest)
        .create_wallet_no_persist()
        .unwrap();
    let addr = wallet.next_unused_address(KeychainKind::External);

    // Fund the wallet with an unconfirmed tx.
    let funding_tx = Transaction {
        output: vec![TxOut {
            value: Amount::from_sat(50_000),
            script_pubkey: addr.script_pubkey(),
        }],
        ..new_tx(0)
    };
    insert_tx(&mut wallet, funding_tx);

    // Spend via the pk(A) branch (child #0).
    let policy = wallet
        .public_descriptor(KeychainKind::External)
        .extract_policy(&SignersContainer::default(), BuildSatisfaction::None, wallet.secp_ctx())
        .unwrap()
        .unwrap();
    let path = [(policy.id.clone(), vec![0])].into_iter().collect();
    let condition = policy.get_condition(&path).unwrap();
    builder
        .add_recipient(addr.script_pubkey(), Amount::from_sat(10_000))
        .set_condition(condition);
    let mut psbt = builder.finish().unwrap();

    // panics at src/wallet/utils.rs:117: "Overflowing addition"
    let _ = wallet.finalize_psbt(&mut psbt, SignOptions::default());
}
```

**Expected behavior**

`finalize_psbt` should not panic when a spent output is unconfirmed; the `older()` branch should simply be treated as not satisfied (or the input reported as not finalizable).

**Build environment**

- [x] No

trakshan-mishra:
Fixes #557.

### Problem

`Older::check_older` (`src/wallet/utils.rs`) computes the height at which a relative locktime is satisfied as `create_height + n` and unwraps the sum with `expect("Overflowing addition")`. `Wallet::finalize_psbt` maps an unconfirmed previous transaction to a `create_height` of `u32::MAX`, so that addition overflows and panics for any `older(n)` with `n` greater than zero.

The `Older` satisfier is only consulted when the input's `nSequence` does not already satisfy the CSV branch, so this is reachable through ordinary use: spending an unconfirmed UTXO of a descriptor that contains an `older()` branch which is *not* the one being used, e.g. spending via `pk(A)` of `or_d(pk(A),and_v(v:pk(B),older(144)))`. Since `Wallet::sign` finalizes by default, callers reach it without opting in.

### Fix

Return `false` when the satisfaction height does not fit in a `u32`, rather than panicking. Such a height can never be reached, so the branch is simply not satisfied — the same answer the comparison would give with wider integers. This matches the behaviour the issue asks for ("the `older()` branch should simply be treated as not satisfied").

The change is in the `Satisfier` impl, so it covers both call sites: finalization in `src/wallet/mod.rs` and policy extraction in `src/descriptor/policy.rs`.

`After::check_after` performs no addition and is unaffected.

### Tests

- `test_finalize_psbt_with_unconfirmed_input_and_unused_csv_branch` (`tests/wallet.rs`) reproduces the reported panic end to end. Verified to fail on master with `panicked at src/wallet/utils.rs:117: Overflowing addition`, and to pass with this change.
- `test_check_older_unreachable_satisfaction_height_is_not_satisfied` covers the overflow directly at the unit level.
- `test_check_older_compares_against_the_satisfaction_height` covers the ordinary satisfied and unsatisfied heights, which had no unit coverage before.

### Checklist

- [x] All commits are GPG signed.
- [x] The PR description links to the issue it solves.
- [x] Ran the `just pre-push` steps locally: `cargo +nightly fmt --all -- --check`, both `cargo check` variants, `clippy` with `-D warnings` on the stable and unstable surfaces, `cargo test --workspace` on both surfaces, and `cargo doc` with `-D warnings`. All pass.
- [x] Tests reproducing the bug (and now passing) have been added.
- [x] No API change, so nothing breaking.

One note: `cargo +nightly fmt` on current master also wants to reformat `src/descriptor/policy.rs`, which is unrelated to this fix (and looks like the nightly drift tracked in #535), so I left that file untouched.


j-kon (approved):
ACK 7461a4341dfa53dc0d292573154758f16d28f376

Verified:
- Confirmed panic reproduction on master at `src/wallet/utils.rs:117` in both debug and release builds when finalizing an unconfirmed input in a descriptor with an unused `older(n)` branch.
- Confirmed PR #565 cleanly resolves the panic and passes all unit/integration tests in both profiles.
- Verified arithmetic boundary transitions around `u32::MAX` (`create_height = u32::MAX - 1`, `u32::MAX - 100`, etc.) and confirmed exact `>=` comparison matches BIP68 relative locktime semantics.
- Verified that returning `false` on unreachable height correctly signals an unsatisfied condition to Miniscript satisfiers.

Issue #549, opened 2026-09-14
Fee bump with `fee_absolute` accepts equal fee and does not check fee rate


noahjoeris:
**Describe the bug**

When bumping a fee with `TxBuilder::fee_absolute`, `Wallet::create_tx` only checks `fee < previous_fee.absolute` (`src/wallet/mod.rs`, `FeePolicy::FeeAmount` arm). Two things are missing compared to the `FeePolicy::FeeRate` arm:

- A replacement paying exactly the same absolute fee as the original passes the check.
- The resulting fee rate is never compared to the original's. A replacement that is larger than the original (e.g. more inputs after coin selection) can end up with a lower fee rate than the original, and even a larger fee may not cover the incremental relay fee.

Such transactions are built without error but are rejected by BIP125 replacement policy (rule 4: must pay for its own bandwidth at the incremental relay fee; Bitcoin Core additionally requires a higher fee rate than the original).

This issue was found by AI.

**To Reproduce**

```rust
let (mut wallet, _) = get_funded_wallet_wpkh();
let addr = wallet.next_unused_address(KeychainKind::External);
let mut builder = wallet.build_tx();
builder.add_recipient(addr.script_pubkey(), Amount::from_sat(25_000));
let psbt = builder.finish().unwrap();
let tx = psbt.extract_tx().unwrap();
let original_fee = wallet.calculate_fee(&tx).unwrap();
let txid = tx.compute_txid();
insert_tx(&mut wallet, tx);

let mut builder = wallet.build_fee_bump(txid).unwrap();
builder.fee_absolute(original_fee);
// Succeeds although the replacement pays no more than the original.
builder.finish().unwrap();
```

**Expected behavior**

`fee_absolute` fee bumps should be validated consistently with the `fee_rate` path so that the builder rejects replacements that would not be relayed.

**Build environment**
- BDK tag/commit: `6fc68462` (master)

**Which backend(s) are relevant (if any)?**
- [x] None / not backend-related (e.g. `bdk_chain`, `bdk_core`)

**Is this blocking production use?**
- [x] No

PR #566, opened 2026-09-17
Fix(wallet): validate absolute fee bump replacements

Jennycj:
Fixes issue #549 

### Description

Previously, `Wallet::create_tx` only checks `fee < previous_fee.absolute` when bumping a fee with `TxBuilder::fee_absolute`. This misses out on 2 arms of the `FeePolicy::FeeRate` arm, causing transactions to be built without error, but then be rejected by BIP125 replacement policy (rule 4: must pay for its own bandwidth at the incremental relay fee; Bitcoin Core additionally requires a higher fee rate than the original).

`Wallet::create_tx` now:
- Rejects an absolute replacement fee that is less than or equal to the original transaction fee.
- Calculates the replacement’s effective fee rate after coin selection and final transaction weight are known.
- Rejects absolute fee replacements whose fee rate is below the original fee rate plus `FeeRate::BROADCAST_MIN`.

This ensures that the appropriate fee error is returned when  `TxBuilder::fee_absolute` attempts to build replacement transactions that do not satisfy the required absolute fee and incremental relay fee rules

### Notes to the reviewers

The selected `fee_policy` is retained so absolute-fee replacements can be validated after coin selection. The final transaction weight is used to ensure the replacement fee rate exceeds the original fee rate by the required incremental relay fee.

Appropriate regression tests were added, and the absolute-fee values were updated in tests:
- `test_bump_fee_reduce_change`
- `test_bump_fee_absolute_force_add_input`

These fixtures previously passed the absolute-fee-only check but no longer satisfy the final replacement fee-rate requirement.

### Changelog notice

`Wallet::create_tx` enforces a strictly higher replacement fee and validates the final replacement fee rate after coin selection.

### Before submitting


src/wallet/mod.rs
```diff
@@ -1599,6 +1600,32 @@ impl Wallet {
...
+                );
+                if fee_amount / tx.weight() < required_feerate {
+                    return Err(CreateTxError::FeeRateTooLow {
+                        required: required_feerate,
+                    });
+                }
+            }
+        }
+
+        if matches!(fee_policy, FeePolicy::FeeAmount(_)) {
```
busayo-OD, line 1616: This FeeAmount check appears twice consecutively. The duplicate block can be removed.

src/wallet/mod.rs
```diff
@@ -1599,6 +1600,32 @@ impl Wallet {
         // Sort inputs/outputs according to the chosen algorithm.
         params.ordering.sort_tx_with_aux_rand(&mut tx, rng);
 
+        if matches!(fee_policy, FeePolicy::FeeAmount(_)) {
+            if let Some(previous_fee) = params.bumping_fee {
+                let required_feerate = FeeRate::from_sat_per_kwu(
```
busayo-OD, line 1605: This enforces a fee-rate floor (previous_rate + 1 sat/vB) rather than the BIP125 fee requirements. These differ when the replacement's weight changes, so the resulting minimum fee can be either higher or lower than the BIP125 requirement.

busayo-OD, line 1605: I’m wondering if this should use the final realized fee instead of the requested fee_amount, since dust handling can add excess to the actual fee.

busayo-OD (commented):
Thanks for working on this! There are a few things around the fee validation logic that I wanted to mention.

src/wallet/mod.rs
Jennycj: The caller passes the requested fee_amount which will serve as the final realized fee. The check is placed after coin selection, just before the psbt is finalized, so dust handling should already be factored in.

Jennycj: The current implementation does a fee rate increment. I would go back to work on this. 
Thank you!

src/wallet/mod.rs
Jennycj: I missed removing this after testing. Good catch!.

src/wallet/mod.rs
busayo-OD: I think there may still be a case where fee_amount differs from the realized fee. I was able to reproduce this with a 720 sat requested fee resulting in a 1,000 sat realized fee due to dust handling. Since fee_amount isn't recomputed after coin selection, the check would still use 720 sat rather than the realized fee.

Jennycj: Yeah, this was covered in the updates made in [94f3941](https://github.com/bitcoindevkit/bdk_wallet/pull/566/commits/94f3941fb02a0ddcbc445ef0a2ceb65c69fe5897)

PR #569, opened 2026-09-17
fix(signer): validate wildcard derivation paths

### LINKED ISSUE #563: Xprv signer signs for hardened child under an unhardened wildcard descriptor
**Describe the bug**

`SignerWrapper<DescriptorXKey<Xpriv>>::sign_input` (`src/wallet/signer.rs`) decides whether a PSBT `bip32_derivation` / `tap_key_origins` entry belongs to the signer via `DescriptorXKey::matches`, then derives the child for the full path in the PSBT and signs. `matches` is documented to ignore the last path step entirely when the key has a wildcard, so it does not check whether that step is hardened or not. As a result, a signer built from `wpkh(tprv.../0/*)` will happily derive and sign for `m/0/1h` (hardened last step), even though the public descriptor `.../0/*` cannot represent that key. The mirror case (`/*h` descriptor, unhardened last step) is accepted the same way.

`Descriptor::derive_from_psbt_input` already filters hardened steps in this situation (see `test_derive_from_psbt_input_with_hardened_key_origin_does_not_panic` in `src/descriptor/mod.rs`); the signer path has no equivalent check, so a PSBT can obtain a signature for a key outside the descriptor's key space.

This issue was found by AI.

**To Reproduce**

```rust
// tests/
use std::str::FromStr;

use bdk_wallet::descriptor::IntoWalletDescriptor;
use bdk_wallet::signer::{SignOptions, SignersContainer, TransactionSigner};
use bitcoin::secp256k1::Secp256k1;
use bitcoin::{
    Amount, NetworkKind, OutPoint, PrivateKey, Psbt, ScriptBuf, Transaction, TxIn, TxOut, absolute,
    bip32, transaction,
};
use miniscript::Descriptor;

const TPRV: &str = "tprv8ZgxMBicQKsPdZXrcHNLf5JAJWFAoJ2TrstMRdSKtEggz6PddbuSkvHKM9oKJyFgZV1B7rw8oChspxyYbtmEXYyg1AjfWbL3ho3XHDpHRZf";

#[test]
fn xprv_signer_signs_hardened_child_of_unhardened_wildcard() {
    let secp = Secp256k1::new();
    let (wallet_desc, keymap) = format!("wpkh({TPRV}/0/*)")
        .as_str()
        .into_wallet_descriptor(&secp, NetworkKind::Test)
        .unwrap();
    let signers = SignersContainer::build(keymap, &wallet_desc, &secp);

    // Key at m/0/1h: not representable by the public descriptor `.../0/*`.
    let tprv = bip32::Xpriv::from_str(TPRV).unwrap();
    let full_path = bip32::DerivationPath::from_str("m/0/1h").unwrap();
    let child = tprv.derive_priv(&secp, &full_path).unwrap();
    let child_pub = PrivateKey::new(child.private_key, NetworkKind::Test).public_key(&secp);
    let child_spk = Descriptor::<bitcoin::PublicKey>::from_str(&format!("wpkh({child_pub})"))
        .unwrap()
        .script_pubkey();

    let prev_tx = Transaction {
        version: transaction::Version::TWO,
        input: vec![TxIn::default()],
        output: vec![TxOut {
            value: Amount::from_sat(100_000),
            script_pubkey: child_spk,
        }],
    };
    let unsigned_tx = Transaction {
        version: transaction::Version::TWO,
        input: vec![TxIn {
            previous_output: OutPoint::new(prev_tx.compute_txid(), 0),
            ..Default::default()
        }],
        output: vec![TxOut {
            value: Amount::from_sat(90_000),
            script_pubkey: ScriptBuf::default(),
        }],
    };
    psbt.inputs[0].witness_utxo = Some(prev_tx.output[0].clone());
    psbt.inputs[0].non_witness_utxo = Some(prev_tx);
    psbt.inputs[0]
        .bip32_derivation
        .insert(child_pub.inner, (tprv.fingerprint(&secp), full_path));

    signers.signers()[0]
        .sign_transaction(&mut psbt, &SignOptions::default(), &secp)
        .unwrap();

    // Currently fails: one partial signature for m/0/1h is added.
    assert!(psbt.inputs[0].partial_sigs.is_empty());
}
```

**Expected behavior**

The xprv signer should only sign for key origins whose final step is consistent with the descriptor's wildcard (`*` → unhardened, `*h` → hardened, none → exact path).

**Build environment**

- [x] No

busayo-OD:
Fixes #563

### Description

`SignerWrapper<DescriptorXKey<Xpriv>>::sign_input` uses `DescriptorXKey::matches()` to determine whether a PSBT key origin
matches the signer. Since `matches()` ignores the final derivation step for wildcard descriptors, a hardened child could be accepted for an unhardened wildcard, and vice versa.

### Notes to the reviewers

`DescriptorXKey::matches()` intentionally ignores the wildcard's final path step, so the additional check is required to validate the wildcard-specific derivation constraint.

Added regression tests covering both hardened and unhardened wildcard descriptors and their valid and invalid derivation paths.

### Changelog notice

- Validate wildcard derivation hardness when signing with an xprv signer

### Before submitting


j-kon (approved):
ACK 6b1518515c8ce45d7fa5c6c2b6e902c9dc59f2a6

Reviewed and verified:
- Confirmed reproduction of #563 on master: `0/*` signed for `m/0/1h` and `0/*h` signed for `m/0/1`.
- Confirmed PR #569 rejects both invalid combinations and accepts legitimate ones across both debug and release builds.
- Verified that `DescriptorXKey::matches()` guarantees prefix equality and length (`compare_path.len() + 1`), so `keysource.1.into_iter().last()` is guaranteed to be the wildcard child step.
- Verified path length edge cases (missing child, extra steps, empty paths) and `Wildcard::None` exact path matching.
- Verified that `tap_key_origins` (Taproot key-path and script-path) enforces the same wildcard hardness checks.
- Verified that multiple candidate origins in `find_map` skip invalid entries without error and sign valid ones.

